PR #1, opened 2016-09-08
Jogl upgrade

Sailsman63:
Changes to update our version of the JOGL and Gluegen libraries to Jogamp 2.3.2. This includes upgrading the required Java libraries, and code changes to match their API changes. Mostly, everything was still there, but they've done some major refactoring.

One, potentially significant change, is in commit 9dd6a7c. The buffers used to transfer data to the native JOGL drivers are now explicitly created as direct buffers. (That's what the API requires) I'm not sure if the original buffers were direct or not, but in theory, a large scene could run you out of direct memory, even when the heap still has plenty of space.

I'm going to leave this up for comment for a bit before merging, but it does build and runs.

This PR does __NOT__ include the various Native libraries that are needed for everything to actually work. The old ones were not in the repository, and I'm not sure that we want the new ones there, either. They're available on bintray, and on www.jogamp.org. You'll need the gluegen-rt-natives-yourplatform.jar and jogl-all-natives-yourplatform.jar, though it does no harm to have all platform jars placed in the lib/ directory. Gluegen choses the correct one at runtime.

Longer term, I'd like to consider switching the build scripts around to automatically download/update such dependencies rather than shipping them around in the source repo.


Sailsman63:
EDIT: Never mind the bit about not having the natives. They're supposed to be in the installer tree. I'll add in a bit.

Sailsman63:
And, I realized that adding to the installer tree is going to mean editing some installer configuration files, as well. This I've done, but I can't test because I'm having issues using the compatible 4.3.2 izpack. I should update everything to izpack 5.0.9, along with fixing the old windows install issue, but that's a little more involved. I'd rather save that for a more general cleanup of the build system. @peastman, if you're still set up to do the older installation builds, can you confirm that the izpack at least processes?

peastman:
Everything looks good from a quick scan.  I'll see if I can build an installer.

peastman:
Building installers seems to work fine, though I didn't test that all of the installers actually work.

Sailsman63:
Sounding good. This last push should fix the "Step 7" issue on Windows w/64bit JVM. Could you run up a Windows installer, and email or message it to me? I'll test on my system. I'm still having issues with the old version of izpack - you can't get it fully installed, and the standalone jar has a slightly different set of options... As I said, a proper upgrade is a little out of scope for this PR.

I didn't realize, the last time that I looked at this, how weird a fix it is. We essentially have part of the izpack installation internals in the AOI source tree. Whomever wrote the installers, I think, had a somewhat messed up install of izpack to test with.

peastman:
I don't have Windows, so I can't test that myself.  I'll email you a link to the installer.

Sailsman63:
No joy. The installer hangs at step 7, as it has for years. I see a couple of options at this point: 

 1. Continue tweaking here. This could be frustrating to test, and no guarantees.
 2. Go for a full update to the latest, supported izpack. This will greatly expand the scope of this particular branch. Not my favorite.
 3. Leave as is, accepting that 64 bit JVM on windows will have the installation bug. Leave the update for another day.
 4. Remove desktop integration/file assocation support *temporarily* from the windows build. Add it back when updating the build process, but installers created in the meantime will finish cleanly.

peastman:
The Windows installer problem isn't really related to updating JOGL, which is what this PR is for.  So let's go ahead and finish this up, get it merged, and then deal with that in a separate PR.

Sailsman63:
That's what I was thinking. Do we have a preference for history style? Some projects would want the entire history merged, with all the interim commits. Some would do a squash merge (--no-ff) and others would request that the original author rebase into a couple of logical commits - in this case, 3:

 - Changing the library files, and updating GLCanvasDrawer to use them, up to 	3d1e0d1
 - Runtime GL availability check (not found until testing a compiled build)
 - Updates to the packagers to use the correct files.

I'm considering this PR a bit of a test case, getting my feet wet on what goes into both sides of the process.

peastman:
I don't care very much.  If you want to rebase to get a cleaner version history, that's fine.  But if you don't feel like doing that, it won't bother me much either!

Sailsman63:
I did choose to clean this up. Mostly, I wanted to make sure that there were no commits that absolutely broke the build, such as failing to compile! As I gain some practice, many of those checkpoint commits will never show where you can see them in the first place!

peastman:
Yay!  Our first merged pull request!

PR #4, opened 2016-09-21
Normalize line terminators to LF

Sailsman63:
Various editors have differing default settings regarding changing
line termination characters on save. Unless every contributor sets
their editor up the same way, this will lead to diffs that appear
to replace an entire file for a relatively small change.

This commit normalizes all existing text files to LF (Unix style)
line endings, and adds a '.gitattributes' file that will ensure that
all future commits use the same style, no matter what the contributor
might set for their own working directory and toolset.


Sailsman63:
This commit should prevent any further readability issues such as experienced with PR #3.

PR #6, opened 2016-09-22
Makiam performance vs clarity

Sailsman63:
Theses changes, originally proposed in PR #3, tend to improve code clarity at the possible expense of performance. Please discuss, or share benchmarks if you have them.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/SplineMeshEditorWindow.java
+++ b/ArtOfIllusion/src/artofillusion/SplineMeshEditorWindow.java
@@ -772,8 +772,7 @@ public class SplineMeshEditorWindow extends MeshEditorWindow implements EditingW
     param = new double [usize+usplitcount][vsize][numParam];
     for (i = 0; i < param.length; i++)
       for (j = 0; j < param[i].length; j++)
-        for (int k = 0; k < param[i][j].length; k++)
-          param[i][j][k] = newparam[j][i][k];
+        System.arraycopy(newparam[j][i], 0, param[i][j], 0, param[i][j].length);
     newparam = new double [usize+usplitcount][vsize+vsplitcount][numParam];
     splitOneAxis(v, newv, vs, newvs, splitv, param, newparam, theMesh.isVClosed());
 
@@ -843,15 +842,13 @@ public class SplineMeshEditorWindow extends MeshEditorWindow implements EditingW
         for (k = 0; k < v.length; k++)
         {
           newv[k][j] = SplineMesh.calcApproxPoint(v[k], s, param[k], paramTemp, p1, i, p3);
-          for (int m = 0; m < numParam; m++)
-            newparam[k][j][m] = paramTemp[m];
+          System.arraycopy(paramTemp, 0, newparam[k][j], 0, numParam);
         }
       else
         for (k = 0; k < v.length; k++)
         {
           newv[k][j] = v[k][i];
-          for (int m = 0; m < numParam; m++)
-            newparam[k][j][m] = param[k][i][m];
+          System.arraycopy(param[k][i], 0, newparam[k][j], 0, numParam);
         }
       if (split[i] || split[p1])
         news[j] = Math.min(s[i]*2.0f, 1.0f);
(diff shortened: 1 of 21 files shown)


peastman:
For short arrays, manually copying it tends to be faster.  For long arrays, `arraycopy()` tends to be faster.  Of course, most place it comes up aren't performance sensitive, so it doesn't matter there.  It's mainly important in rendering and mesh manipulation.

The boxing and unboxing shouldn't affect performance, since the behavior isn't really changing.  It's just letting the compiler insert the conversions rather than writing them out explicitly.

Sailsman63:
The question is: in a mesh manipulation case, would the difference be great enough to be noticeable? I'm not sure how to benchmark this.

The only array copy that I'm seeing in any of the rendering code is when a `ResourcePool`  has decided it needs to grow. Hopefully, that is a rare code path.

peastman:
> The question is: in a mesh manipulation case, would the difference be great enough to be noticeable? I'm not sure how to benchmark this.

One way to test is to add timing code to the function.  Call `System.nanoTime()` at the start and end of the mesh manipulation function, then print out the difference.  See whether it changes significantly when you replace the manual array copies with `arraycopy()`.

> The only array copy that I'm seeing in any of the rendering code is when a ResourcePool has decided it needs to grow. Hopefully, that is a rare code path.

Yes, that should be quite rare.

ArtOfIllusion/src/artofillusion/SplineMeshEditorWindow.java
```diff
@@ -758,8 +758,7 @@ void subdivideCommand()
     param = new double [usize+usplitcount][vsize][numParam];
     for (i = 0; i < param.length; i++)
       for (j = 0; j < param[i].length; j++)
-        for (int k = 0; k < param[i][j].length; k++)
```
Sailsman63: As a bit of a smoke test, I chose to benchmark this particular method. I used `System.nanoTime()` to time the method `subdivideCommand()` from method entrance to end of the block. Then I loaded up a file that I have with a nice, 5X9 vertice spline mesh, and started subdividing. Timing is pretty comparable. In the early runs, 1st and 2nd subdivision, the times are within about 2%, with a slight edge to the `System.arraycopy()` version. As expected, arraycopy strengthens its lead in later subdivisions with larger arrays to copy. After about 5 subdivisions, method time was increasing radically in both cases, but the bottleneck seemed to be with rendering the subdivided mesh, not in the subdivision. The version with the `System.arrayCopy()` was able to recover a little better. (At least I was able to do a proper shutdown, instead of force-killing the process)

I'm sold - I don't see any reason not to merge this one.

Sailsman63: I just realized that I did something wrong with this benchmark. This code is copying the texture parameter lists, and the object I tested on has no parameters. Basically measuring copying an empty array reference. I'll redo with real parameters. in this snippet, `k` is the number of different names parameters, correct? This means that it will tend to be a small array, and in many cases, be length '0'

Sailsman63: Just repeated the test with two texture parameters mapped. The results are reversed. For a fairly small subdivide, (subdividing an 8x16 spline mesh) the arrayCopy version is approximately 14% slower. (mean of 5 runs) Larger subdivides are worse, but there seem to be other bottlenecks. The times for this run are all over the place.

Of note: I added a counter to see how many times the inner loop in question gets called: 81 loops through this bit of code to subdivide a 4X8 mesh.

Sailsman63: To be clear, for the small mesh, we are talking differences of a couple tenths of a millisecond for the entire method to return. When we move up to a 32X64 mesh being subdivided, it's more like a couple hundredths of a second, but at that scale other things are causing bigger issues. The System.arrayCopy() seems to create more junk, as well. Other factors, possibly GC, inflated the Standard Deviation on that run up to 95% of the mean. At that scale of subdivision, my UI becomes unresponsive.

Sailsman63:
I cherry-picked the boxing and un-boxing simplifications, as they do make the code more readable. For the time being, System.arraycopy does not seem to be the way to go, especially in some of the mesh manipulation code.

PR #7, opened 2016-09-22
Refactoring: Replace printStackTrace() calls using logger output

Sailsman63:
This change, originally proposed in PR #3, requires some more thought. While using a logging framework vs. `System.out.println()` is arguably a more flexible option, we need to consider what we want logging in AOI to look like. As is, this PR will require additional configuration to log to anywhere useful - it currently will not even print to stdout.

If we want to go this route, some further conversion is in order. Some of the standard translators, at least, use `System.out.Printerr()` and were missed in the original conversion. @peastman, your input would be appreciated.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
+++ b/ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
@@ -16,11 +16,15 @@ import artofillusion.ui.*;
 import java.io.*;
 import java.util.*;
 import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 
 /** This class keeps track of program-wide user preferences. */
 
 public class ApplicationPreferences
 {
+    private static final Logger logger = Logger.getLogger(ApplicationPreferences.class.getName());
+
   private Properties properties;
   private int defaultDisplayMode, undoLevels;
   private double interactiveTol;
@@ -58,7 +62,7 @@ public class ApplicationPreferences
       }
     catch (IOException ex)
       {
-        ex.printStackTrace();
+        logger.log(Level.INFO, "Exception", ex);
       }
   }
 
@@ -76,7 +80,7 @@ public class ApplicationPreferences
     }
     catch (IOException ex)
     {
-      ex.printStackTrace();
+      logger.log(Level.INFO, "Exception", ex);
     }
   }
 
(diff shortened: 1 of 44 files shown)


peastman:
I think you summarized the situation well.  The advantage of using a logging framework, rather than just printing to stdout, is that it's more flexible.  But that's only an advantage if you actually use that flexibility.

Is there demand for improved logging features?  If so, we should discuss what those are, then make this change as part of implementing them.  If not, then let's not fix what isn't broken.

makiam:
Main idea to see and differentiate logging data importance.
Easy to configure more advanced logging options ex. log to file...

makiam:
I create commit to fix formatting: [Format](https://github.com/makiam/ArtOfIllusion/commit/e8f18b8fcbd5b094dafcf8084f37ddc7e583885d)

Sailsman63:
@makiam, There's been enough confusion with the revision history that I'm finding it difficult to apply your latest. Your fork history is now messed up enough that you will find it difficult to do pull requests going forward. I'm sorry, some of that was my fault. I'm going to suggest a set of steps to clean it up, starting in your master branch:

 * ` git reset --hard <upstream>` 'upstream' referring to whichever branch you are pulling down the official AOI repo to. This will clear out all of the now outdated commits, etc.
 * `git push -f` to make those cleanups appear on your github account.

So far, so good? The next few are for any time you would like to set up some new contributions to _any_ github project:
 * `git branch <name-of-new-contribution>` You don't actually want to commit your contributions directly in master - if you do, you'll end up with more cleanup work
 * `git checkout <name-of-new-contribution>` Move over to your new branch. This is where you want to do your work.
* git push origin <name-of-new-contribution> puts the changes up under your account in a separate branch. Start your pull request from this branch. Use a different branch for changes related to a different feature or bug.

Sailsman63:
We also need to figure out what use, if any, we would make of a more flexible logging system. When trying to debug user issues, we usually want just about everything possible. I'm not seeing a reason to have multiple logging levels at this point.

How would a user configure logging to file? Many of our users are not very technical, so we end up posting a command-line start option for them that redirect stdout and stderr to a text file.

makiam:
I think we may drop this PR at this moment,,,

Sailsman63:
This branch can't really be merged anymore. The question of improving the logging story is continued in #50

PR #8, opened 2016-09-22
Makiam's changes - need clarification

Sailsman63:
These changes, originally proposed in PR #3, need a little discussion and clarification.

In particular:

 - Why `StringBuffer` over `StringBuilder`? The buffer version is syncronized, but none of the affected code needs to be thread safe.
 - Due to the somewhat deep inheritance structure of AOI code, it might be desirable for a subclass to declare that it implements an interface for clarity purposes, even if the Parent class already does.
 - removing unnecessary `return`s makes sense, but in the case of `ArtOfIllusion/src/artofillusion/math/PerlinNoise.java` I need to think about what the class is doing. If the returns are simply to prevent the switch statement from continuing through and testing further cases, perhaps they should be changed to `break`?

### DIFF
--- a/ArtOfIllusion/src/artofillusion/CurveEditorWindow.java
+++ b/ArtOfIllusion/src/artofillusion/CurveEditorWindow.java
@@ -20,7 +20,7 @@ import java.awt.*;
 
 /** The CurveEditorWindow class represents the window for editing Curve objects. */
 
-public class CurveEditorWindow extends MeshEditorWindow implements EditingWindow
+public class CurveEditorWindow extends MeshEditorWindow
 {
   protected BMenu editMenu, meshMenu, smoothMenu;
   protected BMenuItem editMenuItem[], meshMenuItem[];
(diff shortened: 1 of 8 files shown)


peastman:
> Why StringBuffer over StringBuilder?

It looks to me like the commit comment has it backwards: the change actually replaces `StringBuffer` with `StringBuilder`, not the other way around!

> Due to the somewhat deep inheritance structure of AOI code, it might be desirable for a subclass to declare that it implements an interface for clarity purposes, even if the Parent class already does.

Agreed.

> If the returns are simply to prevent the switch statement from continuing through and testing further cases, perhaps they should be changed to break?

The returns are fine as they are.  I think the warning was probably just about the very last one, in the `default` clause.  That one isn't strictly necessary, since it's the end of the method, although the symmetry of having a return there might make the code a little clearer.  Of course, given the comment on that line that it should never be reached, possibly it should really be throwing an exception instead.

makiam:
>> Replace StringBuilder with StringBuffer
This was just typo error. Actual one is replace StringBuilder with StringBuffer
>> Refactoring: Suppress or remove unnecessary return statement
For PerlinNoise class I just suppress IDE warning.

Sailsman63:
I Cherry-picked the StringBuilder conversion over to master. The other two, I think we will leave un-done for now. We can revisit if something changes.

PR #10, opened 2016-09-26
Keep reference to currently intersected object resolves #2

### LINKED ISSUE #2: Render Bug involving meshes and refractive materials
When a mesh object with a transparent and partially shiny surface and refractive material are shaded from at least one of the primitive lights in the scene by another mesh object, the refractive properties of the material are ignored.

Since 3.0.

Sailsman63:
The SurfaceIntersection object provided to spawnRay()
may be recycled during call to getDirectLight()

Moved assigenment of hitObject to as soon as there is a
valid object to assign.


peastman:
Wow, that must have been really painful to track down.  Well done!

Sailsman63:
It did have its frustrating moments. At least, I have a bit better understanding of how parts of the raytracer work.... and I can see why it happened in the first place. The old, pre-refactor version of spawnRay() took an RTObject directly, so recycling the Intersection object caused no issues.

PR #12, opened 2016-10-28
Viewmanipulation

peteihis:
Hi.

The basic functionality is there with the center-click to center function.

The scroll wheel is now handled in the MoveView tool because the function seemed more logical in the environment where the rest of the movements are happening. I wonder if that is acceptable to Mac users? I just realized that it probably requires the Meta key to be pressed to work ?? If so, that has to be fixed somehow.

BR

-Pete-


ArtOfIllusion/src/artofillusion/math/TriangleMath.java
```diff
@@ -0,0 +1,134 @@
...
+import artofillusion.ui.*;
+import artofillusion.math.*;
+
+/**
+ *  TriangleProps contains frequently used calculations about triangles
+ */
+public class TriangleMath
+{
+	/** Constructor for TriangleProps */
+	public void TriangleMath(){}
```
Sailsman63: This is not actually a constructor at all, it is a method named `TriangleMath()` with a `void` return type. Unless you are adding state (a data structure to hold a generic triangle, etc) you probably want a *private* constructor, so that no one can create a TriangleMath object. Like so:

    private TriangleMath() {}

ArtOfIllusion/src/artofillusion/math/TriangleMath.java
```diff
@@ -0,0 +1,134 @@
...
+	
+	//********************************//
+	//*   METHODS FOR 3D TRIANGLES   *//
+	//********************************//
+
+	/** Cross product of two side vectors in 3D*/ 
+	public Vec3 cross(Vec3 A, Vec3 B, Vec3 C)
```
Sailsman63: And then, you would mark all of the methods as static, like this:

    public static Vec3 cross(Vec3 A, Vec3 B, Vec3 C)
    {
        return B.minus(A).cross(C.minus(A));
    }

It might also be worth considering potential method name collisions with, eg. Vec3.

Sailsman63: Following on from the bit about a private constructor, you would then mark all of the methods as static, like this:

public static Vec3 cross(Vec3 A, Vec3 B, Vec3 C)
{
    return B.minus(A).cross(C.minus(A));
}
It might also be worth considering potential method name collisions with, eg. Vec3.

ArtOfIllusion/src/artofillusion/view/ClickedPointFinder.java
```diff
@@ -0,0 +1,186 @@
...
+
+import artofillusion.math.*;
+import artofillusion.ui.*;
+import artofillusion.object.*;
+import artofillusion.*;
+import java.awt.*;
+
+public class ClickedPointFinder
+{
+	private TriangleMath T = new TriangleMath();
```
Sailsman63: You should not have to do this. You're creating a dummy object to get access to methods that do not operate on an internal data structure. This follows along with the comments regarding making TriangleMath methods static.

ArtOfIllusion/src/artofillusion/view/ClickedPointFinder.java
```diff
@@ -0,0 +1,186 @@
...
+			return true;
+		else
+			return false;
+	}
+
+	private boolean onTriangle(Vec2[] corner2D, Vec2 point2D)
+	{
+		// The sum of barycentric coordinates is always 1.0. 
+		// If the point is outside, some of those will be negative. 
+		bary = T.bary(corner2D, point2D);
```
Sailsman63: And here, instead of using the object reference `T.` you would call `bary = TriangleMath.bary(corner2d, point2d);`

Or, if you are using a lot of TriangleMath, you could add `TriangleMath.*` to the imports list. Then you would call: `bary = bary(corner2d, point2d);`

peteihis:
Thanks! good comments.

I'll take your advice on those. Some of those I already had a hunch on
somewhere in the back of my head, but learning takes repetition as always.
:)

More methods can and will be added as needed or requested in future. One
obvious one would be something like public boolean inside() to replace the
onTriangle() in the calling class...

Thanks again.

-P-

2016-10-31 17:24 GMT+02:00 Sailsman63 <notifications@github.com>:

peteihis:
Oops,,,, I don't exactly master this thing yet. No big harm done I hope?

Sailsman63:
Not at all. Just a matter of style, etc. Every time master is merged in, you get another merge commit, and they will stay in the history.

ArtOfIllusion/src/artofillusion/ScrollViewTool.java
```diff
@@ -0,0 +1,301 @@
+/* Copyright (C) 2017 by Petri Ihalainen
```
Sailsman63: This file concerns me a bit. Its not really a tool at all, its a scrollEvent handler. 

The existing MoveViewTool and RotateViewTool are real, clickable tools that are also set up to be activated by <meta>(<ctrl>) and <alt> keys. The way that this experiment is developing, I'm starting to think that we should consider reverting the two to their 'classic' style of operation, and leaving them as click-only tools. 

The more inclusive movement controls that you have been working on should be handled as MouseEvent handlers in the canvas itself, or as an inclusive specialized mouse-handler class. In either case, it should be separated from the 'ViewXYZTools' as such. I don't think we want another 'tool' that does not really fit in the tool paradigm.

Sailsman63:
@peteihis I'd like to get this branch updated to be compatible with all of the changes that went into 3.0.3, and clean its history a bit while we're at it.

I want to keep your latest changes to the UI, but I'd prefer dropping the creation of the ScrollTool file.

Would you be interested in learning how to do a moderately complex re-base? (There are parts that will require manual merge resolution - going into a file and figuring out how to get the two versions to combine - and some less-used options involved)

If so, I can walk you through step by step, with an explanation.
If not, would you be okay with me forcing an update to this branch with a cleaned up history? (You may not be aware of this, but project admins have push access to branches for which you created a pull request, unless you change the settings on the branch)

ArtOfIllusion/src/artofillusion/ScrollViewTool.java
peteihis: One reason to put scroll wheel handling into it's own file/class was, that the ViewerCanvas is not aware of the parent window that it is attached to. For a while I had the scroll events handled in the VC and MoveViewTool handle them a bit more to create the view cone graphics to the other views but obviously that was not a very logical to do it.

And it is not a "tool" in the sense either, that I did not see any reason to make it an extension to EditingTool. _(IMHO the Move and Rotate tools are not really "editing tools" either as don't edit anything. Though that arrangement makes sense in the way, that they are assigned to Icons.)_

By click-only tools you mean that they have icons in the tool palette? And they would only produce the very basic functions? -- Anyway the idea of a mose-handler class sounds about right to me. Somehow I'm thinking that the ViewerCanvas is large enough a file already and adding more weight to it .. hmmm...

On the other hand most places where the ViewerCanvas is used would still be using all the mouse-event handling -- Maybe the only exception being the UVMappingViewer, that would do just fine with less options.

peteihis:
> I'd like to get this branch updated to be compatible with all of the changes that went into 3.0.3, and clean its history a bit while we're at it.

Likewise!

> I want to keep your latest changes to the UI, but I'd prefer dropping the creation of the ScrollTool file.

Should the direction then be one or several mouse event handler classes? Eg. ScrollEventHandler and MouseEventHandler? The scroll wheel information (and move and rotate) needs some way it can be projected to the neighboring views in the case that there are any.

> Would you be interested in learning how to do a moderately complex re-base?

Why not? :)  The fastest way forward would of course be, that you'd force the update but thinking, that I might be hanging here still in the unforeseeable future, I think I'd better learn everything I can. And as it happens, currently time is not an issue to me.

> project admins have push access to branches for which you created a pull request

So I had understood. _Admins usually have access to a lot of things for pretty good reasons._ :)

ArtOfIllusion/src/artofillusion/ScrollViewTool.java
Sailsman63: >For a while I had the scroll events handled in the VC and MoveViewTool handle them a bit more to create the view cone graphics to the other views but obviously that was not a very logical way to do it.

That seems wrong -- A ViewerCanvas should not be forcibly drawn on if you are interacting with *different* ViewerCanvas. I'd almost wonder if a better option would be to temporarily add an object - an actual 3d wireframe - into the scene. It could be removed when the mouse action ends. This end of things requires more thought.

By click only, I mean that they should _only_ have Icons. They should be stripped of the "MetaTool" and "AltTool" arrangement. They would have the functions that they currently have in the 3.0.3 master branch. (In fact, they might even be removed entirely in the future.) The improved navigation tools that you are developing here would be an entirely separate implementation.

I would not worry about the size of a source file over-much. While it is true that a large class *can* be a sign of something wrong with a design, the real litmus test is whether the code makes sense in isolation. Would the proposed mouseHandler *ever* make sense outside the context of a ViewerCanvas? If not, it should remain in the ViewerCanvas, perhaps in a nested class to keep the class API cleaner. What to do with UVMeshViewer would take a little thought, but I don't expect it to be hard to block certain parts of the functionality.

Sailsman63:
Lets leave the Scroll-wheel stuff as is for the moment. We'll keep discussing where to go with that in the specific thread.

So: a rebase, part 1.

**Make sure that you are set up to work with git on the command line:**
You need to have a text editor set up for git to launch for certain advanced features. If you are already using an automaticaly launched editor to add commit messages, you're ready to go. If not, run: `git config core.editor notepad` on the command line, to set up notepad for this use.

**Safety First**
Advanced rebases can go wrong if you miss a detail in the commands. Create a duplicate branch to work in, so that you don't get into a hole that you can't get out of: `git checkout -b viewmanipulationcleanup` or similar. Make sure that you have checked out the viewmanipulation branch before doing this step.

**A first interactive rebase**
To get a quick, fairly safe look at the Interactive rebasing environment, we're going to do a very simple fix: 

`git rebase -i HEAD~2`
The `-i` tells git that we want to run a rebase with some manually specified changes. The `HEAD~2` tells it that we are only interested in the two most recent commits.

- Git should launch notepad with a filled in document. At the top will be the hash codes for commits 12eaf6d and  b8a4648, oldest commit first. Like many things in git, some abreviated instructions are written directly in the file. Take a few moments to consider the options.
- In this case, we want to change the instruction in front of commit b8a4648 to 'Squash,' to remove any trace of those two binary files from the repository history.
- Save the changes, and exit the editor.
- Git will do some processing, and then launch the editor again, asking you to pick a commit message for the combined commit. In this case, you can just pick one of the existing ones, and comment the other out by adding a "#" to the beginning of the line.
- Save changes, and exit again. Git will now finish up. I you run `git log` you will now see a new commit where once there were two.

**Now for the slightly dangerous bit:**
Once you are satisfied with the changes to your cleanup branch, you can `git push -f origin viewmanipulationcleanup:viewmanipulation`
the `-f` stands for `force` which tells git to go ahead, even though you have changed your history. The later bit tells git to take your cleanup branch, and push it up to your viewmaniplulation branch in your fork (And thus to this PR)

Once that is done, we can look at some of the more complex stuff.

peteihis:
Done. So far so good, I hope. There are a couple of places in the history, where I could use this, but maybe later then....

For reasons I don't fully understand there was yet another Thumbs.db trying to slip into the git system though again it was not existing in the repository folder. It was read from somewhere else. Mysterious.

ArtOfIllusion/src/artofillusion/ScrollViewTool.java
peteihis: > A ViewerCanvas should not be forcibly drawn on if you are interacting with different ViewerCanvas. I'd almost wonder if a better option would be to temporarily add an object - an actual 3d wireframe - into the scene.

Well, actually ViewerCanvases are not drawing on eachothers. Each viewer canvas has a drawable public viewcone "object" built-in and that is then just given values/activated/deactivated by the tool, that has acces to the parent window. Then at drawOverlay each ViewerCanvas calls a method to draw the "object", if the object is existing. (Using renderLine of the CanvasDrawers.)

That seemed to make better sense, than drawing anything temporary on the scene, deciding which views to draw in on and then deleting it each time a Camera is moving as the same information is needed "all the time".

At he moment the "object" is just a group of Vec3Ds with specific names and the method that draws the view cone knows, what to make of them. That of course is a very prototype kind of contraption at the moment. I'm thinking, that the "externalGraphicObject" could be more like a list of things to draw, that is then used to display any temporary information about things that are not happening on that particilar ViewerCanvas.

To me it seemd better to put the function that uses the graphics object into a tool/handler/something that is created along with the parent window, rather than code it into every subclass of the ViewerCanvas sepatrately. At the moment all the "tools" (Scroll, Move, Rotate) contain the same piece of code that uses the graphics "object" of the ViewerCanvases of "theWindow".

> By click only, I mean that they should only have Icons. They should be stripped of the "MetaTool" and "AltTool" arrangement. They would have the functions that they currently have in the 3.0.3 master branch. (In fact, they might even be removed entirely in the future.) The improved navigation tools that you are developing here would be an entirely separate implementation.

Yes, this is very much the situation with all the 3D-software I have ever used. The move and rotate tools for the view are usually existing and found as icons in the toolbars/palettes but they are hardly ever used as the same things and much more are found under mouse buttons. Of course I have no idea how the software behind there is written. In some ways the icon activated tools seem a bit like a remnant from the times, when mouses and mouse handling in the software were not that developed but so do many other things, so let's not go removing those just yet. :) -- And AFAIK any travel modes in view handling were not existing 17 years ago. The first ones, I saw, came along a few years later and BOY, were they hard to handle!

> Would the proposed mouseHandler ever make sense outside the context of a ViewerCanvas? If not, it should remain in the ViewerCanvas.

Hmm... Isn't that, what the situation is with the MoveViewTool and RotateViewTool at the moment? They are only used with ViewerCanvas though they are just added to every sub class at the creation of the parent window. I guess that should provide the option to use the the ViewerCanvas in contexts, where the user is not supposed to move or rotate the view.

Any way, so far, of course, I have been mostly focusing on creting the features, that I think should be included. -- To use mechanical terms, it could be called a "functional prototype" and the final compositon that can be "released for manufacturing" has to be figured out after we know what the machine needs to work and what the excact supply chain is....   -- From that world, I have picked the idea, that generally speaking a relatively high level of modularity is a good thing, when it comes to keeping highly complex things under control and to some extent flexible :)

Sailsman63:
Looks good.

> There are a couple of more places in the history, where I could use this,

Well, that's part of the idea. The tricky part is that some of the commits from the master branch are interspersed. I'll get some more advanced steps posted later.

ArtOfIllusion/src/artofillusion/ScrollViewTool.java
Sailsman63: Hmm... I think that `MoveViewTool` and `RotateViewTool` are separate because they are being registered as `EditingTools` with the toolpalette.

For the moment, we'll leave it. Let's get the rebase done first.

Sailsman63:
**Rebasing, Part 2**
We're going to want to do this in a copy of the viewmanipulation branch as it currently stands. This one is easier to break...

In the copy, our command is:
`git rebase -i -Xignore-all-space -Xrenormalize master`

Again, this starts an interactive rebase. The two `-X<options>` are to make sure that the windows line endings that have crept in get cleaned up and standardized automatically. (You branched before the `.gitattributes` file that controls this was in place) the `master` tells git that we want everything that is on master used as a base for your own changes. This is how we get rid of that merge history.

Git will launch an editor with all of the changes added in this branch. You will notice that the changes from the merged master branch are not listed - That's because git knows that we want to use them as a parent chain.

- First, mark commit  8ac694f to be `drop` This is a duplicate commit, due to branching before that commit was added to master, and we want it cleaned up.
- Squash any remaining commits that are all part of a previous change - In my test run, I found several, such as  4fa999c, that were files missed from the commit before
- Save and exit the file

The rebase will now mostly progress automatically. Git will occasionally launch the editor, and you will need to confirm or tweak commit messages, as in the shorter, practice sequence.

**"Cannot Rebase" merge errors**
At a couple of points, git will complain that it cannot rebase a certain commit. This is normal, its just trying to get you to manually confirm a more complex merge. When this happens, run `git status` and git will tell you exactly which files need attention.

Opening the file, you will find standard git merge conflict markers. Go ahead and clean up the file for a clean merge. If you have not handled a merge conflict, post back here, and I'll help you out.

When done with the conflicting files, run `git add -u` to stage them, and then `git rebase --continue` to get the rebase back on track.Repeat untill all commits have been rebased, and then make an AOI test build to confirm that nothing broke.

When you're happy with the results, force push as you did last time to update the pull request.

Questions? Let me know.

peteihis:
The .gitattributes looksi like this:

```
*.java text=auto
*.properties text=auto
*.xml text=auto
*.html text=auto
*.txt text=auto
*.md text=auto
```
Is that, what it's supposed to? I don't remember doing anything to it in the past.

Sailsman63:
Can you tell me **exactly** what commands you used? There should be no merge conflicts left at all, and the merge commit 6e291f0 should be completely gone.

- Somehow,  8ac694f was not dropped.
- was the `master` that you rebased off of exactly the same as the current mainline master?

peteihis:
> Can you tell me exactly what commands you used?

Not really any more. Sorry. I was in the viewmanipualtioncleanup branch and tried to follow your instructions. Aborted the rebase process a couple of times, before I got hold of it. 

>Somehow, 8ac694f was not dropped.

It did not show up in the rebase command with the Xoptions, so I tried the other way with 'HEAD~25'. It seemed to work, but it still shows in graph (though somehow disconnected). As a matter of fact at one point the graph showed my branch to the brached off a branch of makiam's... Now it seems to be back the way it was beore. 

> was the master that you rebased off of exactly the same as the current mainline master?

I have no idea, but probably not. How do I make sure it is?

> What's up with this commit? ( 9d4ea8a) Can you tell a little more about the bug its trying to fix?

This is where I got lost with the conflicts of LayoutWindow, so the 'bug(s)' were all my own doing just today.  

At first I had missed one large bit of the conflicts .... Actually the entire file was there twice in <<<===>>>'s. Then, after that, I had an array with too small size (left behind from the older version).  Pushed fixes to just those and squashed... Or so I thought. -- Probably a wrong way of doing it, but I I guess, I did not understand how to handle that by rebase. Somehow, in that commit,  there are modifications to other files showing up, that were actually pushed many steps earlier.

> It also looks like you are editing the way <ctrl> and <shift> are handled in a mouse drag. 

Actually no. The old behavior has not been changed, but some new stuff has been added. Now there are just much more of those options with the different navigation modes & temporary switching between model and travel, so I had to map them out in greater detail. _(And I'll be the first to admit that the code can take a lot of simplifying and compressing. Now there are practically similar bit copied into the files several times over.)_

Why you see the `<ctrl> + <shift>` combinations edited there is, that when I was first writing the code, I was not very consistent in keeping the order in the almost identical looking `if (?ctrl && ?shift)`'s , so what you see in that commit, is just me reorganizing the order in which the values are checked. First `ctrl` or `!ctrl` and the shift or no shift. Much easier to follow the logic. Because the modifiers are independent of each others, this does not change the way it works.

Sailsman63:
>>    was the master that you rebased off of exactly the same as the current mainline master?

>I have no idea, but probably not. How do I make sure it is? Go to local master branch and pull?

Well, given:
>It did not show up in the rebase command with the Xoptions, so I tried the other way with 'HEAD~25'. It seemed to work, but it still was shown on the graph (though somehow disconnected).

`HEAD~25` syntax does not try to rebase onto a different branch, it just allows you to rearrange the existing branch through the rebase mechanism, so you were definitely not rebasing on master. In your local git you should have one branch that is only used to track `ArtOfIllusion/ArtOfIllusion master.` This may be your `master` branch, or depending on how you set up the repository originally, may also be `upstream.` The only things that you normally do with this branch is `git pull` to get the latest core updates, and `git checkout -b` to start new branches. Its also a handy rebase target...

The layout window conflicts are the result of not having the `-X<options>` in the rebase command line. - that file is the main reason that I worried about them in the first place.

Now that you say it, the mouse drag stuff does make sense that way... sorry, I didn't read closely enough.

Would you like me to do the rebase this time, and push it up? Don't worry, you'll get there eventually! Its just that the longer this goes, and the more changes get added, the more complex the rebase will actually become.

If we go that route, Once I've pushed, you will need to:
- pull AOI master updates into your tracking branch (master or upstream, as the case may be)
- create a new *local* (on your computer) branch for further development of viewmanipulation
- pull peteihis/ArtOfIllusion viewmanipulation branch into the new development branch. This is where you work from now on
- delete the old viewmanipulation branch from your local repository.

Let me know what you think.

peteihis:
> Would you like me to do the rebase this time, and push it up?

Go ahead. :) 

I still tried something and now the graph looks different but probably not what it is supposed to. Trying to rebase on my master I see 
```
noop

# Rebase 5da5c96..5da5c96 onto 5da5c96 (1 command)
#
```

and on the viewmanipulation I see things from ViewAnimation engine forward. Git keeps telling me that everythig is up to date and rebase is not doing anything at all. (Yesterday, I did get into the process where files were I was able to edit the conflicts, but not today.) ...And I have deleted and fetched branches and nothing is changing....

Sailsman63:
Should be good to go.

Please remember that you need to delet your local viewmanipulation branch and run `git branch viewmanipulation origin/manipulation` to get a clean copy to work forward from before doing any more work on the viewmanipulation system.

If, in the future, you forget to add a few files to a commit, you can use `git commit --amend` to add them on the top of the same commit before pushing.

peteihis:
Thanks -- looks like it's working now. :)

Sailsman63:
Well, it has no merge conflicts now. However, something still went wrong... I think that you are working from a slightly different `master` than the main project is. 63302f7 should not exist.

Never mind for the moment. I can pick your upcoming changes into a clean history, and push that when its time to merge them all in. Once

Sailsman63:
What was this last commit supposed to change? In my, admittedly short, testing, the perspective change with OpenGL enabled went smoother at commit b74512b than at 520dba0.

peteihis:
The problem was that GLCavasDrawer was drawing elements in two different scales and that's why I had to decide which ones to show during animation (See the screen capture videos on the SF-thread). 
Unfortunately the https://github.com/ArtOfIllusion/ArtOfIllusion/commit/520dba0c9757a5eaa6121b4c60b6d35092e22105 still had a couple of mistakes in it. Now they are fixed and it seems to be working as expected (https://github.com/ArtOfIllusion/ArtOfIllusion/pull/12/commits/63434b1d0fc8136d2b434e8bdd78fcf029493f2d).

Only that, for some unexplained reason, there is a quick flicker in the beginning of the first perspective change. That only happens the first time the perspective is changing. The following times I don't see it.

Sailsman63:
@peteihis could discuss the scope of this PR? What are its specific goals? What defines when it is "ready?"

I'm noticing that the changes here seem to be growing roots into ever more sections of the codebase. If I'd realized how widespread the changes would be, I'd have suggested splitting up the changes into more than one branch, as we have at least two interleaved feature sets here:
- Rework of the view controls, including the new navigation modes
- Animated transitions
One could also argue that small things, like center-clicking in the object tree, while worthwhile themselves, should be considered extra small features in their own right.

This comment is _not_ intended to hurry you along. Time is not a concern right now. Nor is it intended to discourage you from experimenting. I'm very excited by the things that you have come up with, and think that the improvements to AOI user experience are completely worth pursuing.

peteihis:
> the scope of this PR? What are its specific goals? What defines when it is "ready?"

Exactly, what I have been thinking lately.

When I started, I thought that I had a few clear goals in my mind and I certainly did not expect the effects to be as wide as they are. Things have been escalating and partially that is due to some findings I have made along the way. Then I got a little bit to a side track in mid winter, but I guess that sort of a thing could be expected... Had to take a few steps back to get back on track.

Myself I can see about ten sub projects here.

1. The rotation center based view rotation in two modes, the fully 3D and the tray mode. This includes the perspective mode working in conjunction with the parallel and the click to center function.
2. Having the camera and light views follow that same logic. _(BTW, I have a few changes pending on this one, that make the saved .aoi incompatible with older AoI:s but I have commented those parts out for the time being.)_
3. The renewed _Fit To XXX_ functions, including zooming to a selection of vertices in mesh editors and bones in the bone editor. etc.
4. The animation engine.
5. User adjustable perspective effect. After the latest changes that is now possible.
6. The Fit To Selection by center clicking the tree. (Relies on 3)
7. The requested user defined view point presets.. (This one seems to mean a new type of a 3D-object. Not really given much thought on it yet).
8. Displaying the view cones of the active view on the other views. 
9. Making children of the camera follow the camera in real time, when moved by 2. Now happening with scroll wheel zoom.
10. The travel modes. -- With some cue graphics. 
11. The transparent overlay drawer -- related to 8 an 10 and in some cases 1.

When I started, I already knew, that the matrix maths behind the view system was fine, but the way it worked had not been clear, when the 'infrastructure' around it had been built. It took me some time to figure out, what was happening there. Now that part seems to be clear to me, though there still is one method in the ViewerCanvas that I don't understand .... multiplying a matrix by zero and then making something out of that. By some miracle that still works, so I have not touched it.

Now in many cases I have felt, that many of the things I have done so far are interconnected in ways, that make them impossible to be separated, but that might actually not be the case. The working order just needs to be figured out. (Before now, I probably did not have enough information for that.)

Now there are a few choices how to go forward:

- Start from the current AoI master, fix the internal oddities first (now that I know them) and then start to feed the project in as smaller portions. Say 1,2 and 7 as the first step and then 4 and 3. The 3 is built on the 4 and they both are built on 1 and 2. This would probably be the best option from master point of view and would make it possible to make releases with a smaller amount of changes.
- Prune down the current project a bit.... Comment out or remove travel modes and cue graphics and get the perspective related animations (=+ camera changes)  finished. -- AND clean the code quite a bit. There are parts, where I may have mixed styles in a way, that is not acceptable to real software people and parts, that are not very fluent otherwise.
- Keep going and see where it ends.... 

Right now (continuing on the current path) there are still a few things, that need to be finished: The already mentioned perspective related animations, a view preferences menu and dialog window and some improvements to how the displaying of view cone happens. Plus the code trimming.... _(comments will be welcome, when that day happens)_. -- And after that I'd call it a project. Possibly around Easter time thinking of the recent pace of progress :/ -- I admit, that I have some ideas, that would be better left for later.

How does that/those sound to you?

peteihis:
Now, thinking of the _"Start from the current AoI master"_-option I created a new local git repository from my current remote master. Somehow the ViewAnimation.java has ended up in there and causes error at compile. I '_git rm_'ed that and the build works and the files seem to be 3.0.3.

The file is an old version (4 months) and it definitely is not supposed to be there. If I now just push the master (without the file) will I get rid of it or should I just forget about it?

Sailsman63:
I think that 10 sub-projects is probably a little too fine-grained.

I see that you now have a master that matches the main repository! And that your latest rebase is built on top of it correctly. Great! What you need to do is to remember to keep one local branch that is _only_ used to `git pull` from the main repo. You create your branches from this one, and then work on them separately.

I may be adding some commits to the main repo shortly. (next few days) There should be no need to re-base on top of them - That's really an operation that we should need very rarely, and in other circumstances, would be confusing to others.

We really want to avoid committing files with commented-out code. Its fine to comment out lines when experimenting, but when you've got things working as expected, go back and completely remove the old stuff. We can always use git to get it back, if we need to.

I'm also not too keen on trying to split the project up now. I could probably do it, but some of the commits are actually working on several parts, even if they don't interconnect.

One principle that comes to mind is 'minimum viable product.' Sure, the cue graphics could probably be improved, or other things could be polished, but we won't know the best way to do these things until we have some real world time under our belts.

Couple of questions regarding your todo list:
- Perspective animation changes seem to work as expected. What exactly needs to be finished/fixed?
- What sort of preferences? This sounds like a separate feature, for future work - the main line does not have such preferences, and adding them to the new view controls may not be urgent.
- Improvements to displaying the viewcone: Are there any specific difficulties that the current method will cause _users?_ Internal refactors can be run as separate projects.
- Mouse Handlers: Need a refactor, but that might be able to wait. I need to ask Peter and the rest of the community one thing before deciding
- Code trimming: depends on what you mean. I have one big thing, here. More below.

The one thing that I would like to change is the manner in which `TriangleMath` got implemented. Do you recall the discussion above? I think that the conversion could be done in a single commit.

If none of the above are actually blocking, I think that we should consider moving ahead on merging what we have, and getting it out where we can get feedback from the wider community. I would probably do one final clean-up re-base to remove commented-out code, etc, (I'm going to do a `.gitignore` from the ground up. Some of the selectors that you used are a little too big of a sledgehammer.) We've both learned a lot about managing something like this, and can work more smoothly in the future.

peteihis:
> I see that you now have a master that matches the main repository!

I started to wonder about two files/commits that were not supposed to be there... For a while I got things looking really weird, but eventually managed to ask internet the right questions and combine the right answers in a way that worked. Pfhewhh... _(Studying the manual is not very helpful, when you have no idea what you are looking for.)_

- **Perspective animation:** At the moment there are two cases where the perspective part of the animation is skipped: Changing navigation mode and changing the view from/to a camera/light. (Also there is something funny about the light-camera views, but that's a different fix.) That is nothing big and ugly and it can be taken care of later.
- **What sort of preferences?** Something like this and maybe the animation settings too could be moved to the same place with the rest. Call it Settings, Preferences, Features, I don't know... It would seem to be abi all of those. Many of the things on the UI-prototype don't exist yet (I just had my imagination fly), so there probably is no urgent need for it now.
view_settings_prototype

- **View cone:** Currently it does not make distinction between perspective and parallel modes, which can be a bit confusing. Also the way the information is created and is stored is very prototypish... The plan is to have a method that creates a couple of 3D-points an a list of lines, that are rendered onto the screen. As a list (or array) rather than a fixed set of variables it'd be more flexible about the content.
- **Mouse Handlers:** OK.
- **Code trimming** ... Well... for example there are parts where the existing file is using two spaces to intend the text and I have added parts that use tab-character and so on. That's just about how it looks. -- And then in the Move and Rotate tools there are things, that could be done a bit differently, but that is not very high on the priority list to me as they work now. -- And as I came to think about it, the .properties files should be rearranged to match the current menu system. That can be taken care of as separate fix.
- **TriangelMath:** What did you have in mind?

On the last commit I forgot to remove some commented-out parts. Meant to delete them after testing but forgot, when I finally was happy with the way it worked... I'll probably have to fix something there anyway so I'll try to remember next time...

But it's fine to me to put it out for testing as it is (Call it 3.1.ea1 or something?) and keep working on the details while collecting feedback.

Sailsman63:
- The tweaks that you mentioned regarding perspective animation definitely should be left for later.
- Same for the options menu. Some of these I don't even know what they would do.
- View Cone: I see that you have made some updates here. Currently, I'm getting a rectangular prism with a pyramid at the end ending at the view location...
- Mouse handlers: we may need to preserve the original move and rotate, with new functionality moved into a single mousehandler that uses <ctrl> <alt> etc. Still waiting for a community consensus on this. If needed, I might be able to do that particular cleanup.
- TriangleMath: This should a utility class like `FastMath`. No constructor, all `static` methods, referenced by `TriangleMath.methodName()` rather than by a dummy object in the local scope.
- I think I can clean up the commented out bits before merging. Want to know for sure what we are doing with the mousehandlers first.
- Reorganizing the properties file does not belong here. If we're going to do that, we should reorganize _all_ of the properties files at the same time. I'll strip that commit before merging, I've been thinking about a small tool to make working with the translations easier.

peteihis:
Mostly I agree. A couple of points though:

> we may need to preserve the original move and rotate

The MoveView and RotateView tools are in the very core of all of this. The old ones work with a different logic, which a kind of is _lacking_ consistent logic. This is very much, why I wanted to start this project in the first place. If needed it'd be easy to simply the new ones to fit the purpose of using them as icon-activated tools.

I thought Mr. YardStick's comment on SF was actually a good one, so I did a bit of testing on my lap top without the mouse attached. In Space and Tray modes the tools work well. In travel modes not so much.... Many of the functions rely on modifier keys and the scroll wheel and if you'd be using a device without a mouse and a keyboard, then those modes are basically useless.

Then _(I haven't really paid attention to this before)_ I began to wonder why the view tools are placed where they are in the tool palette? Now some plugins push them forward and some don't and they end up somewhere in the middle and not even on the same row. To me the logical place for them would be on the top of the pile, where they could go undisturbed and even be hidden if the user wishes so. -- But then I don't know, what all the tools and plugins would do if the icons were on the top or hidden.

viewtools

At least this part of the thinking is very clearly out of the scope now. Worth of another _(hopefylly much smaller)_ project, I think.

> This should a utility class like FastMath. No constructor, all static methods,

That's what it is now. I pushed it a long time ago, along with some other stuff. Though, for reasons unexplained I wrote there a private constructor ... I have no idea why or if that kind of a thing really even exists. :D Some of the comments could be more informative, I'll see to that soon. 

> options menu

I was really just sketching something for the future and trying to figure out how to make it look logical. Though, I think, the cue graphics should have some choices and that would be be easiest to show on a dialog window rather than in menu items.  Then if there were a dialog, some of the things in the view menu could also be moved into a common dialog window, where you could see all the settings in one glance rather than go browsing through menus and sub menus... 

Another project head.

> View Cone

That rectangular pyramid presents parallel mode. There is triangular one for perspective. I'm not sure if that is intuitive enough as it is now but at least it is less confusing than the earlier one especially with very large or very small magnifications in parallel mode

The thing is, that now the cone is produced by an inner class in ViewerCanvas. That took a lot of fuss away from the tools, that use it and now it _(in principle)_ can be used to display other informative temporary 3D graphics as well, as soon as someone comes up with any. The down side is, that it does not recycle object _(Vec3s, Colors, and ArrayLists)_ so I guess it is a bit trashy.

Sailsman63:
- ToolPallette: In the past, @peastman has said that this is due to the load order of the various plugins, which is not deterministic. Certainly something to do a future project on, not really related to this one.
- TriangleMath: Sorry, I somehow missed that you had already changed that. Exactly what I meant, except we really need to get rid of that factory method and put in a dummy constructor: `private TriangleMath() {}`
- View Cone: Ahh... Makes much more sense. We can refine the reuse of objects later, if necessary.
- Old vs. New tools: I see what you mean. I've done some testing with my laptop, and as long as I have two buttons available on my trackpad, everything seems good for the modeling type views. I think that the issue with the other modes is that they _Are_ currently implemented in the separate view tools. The new navigation modes don't really fit with the separation of 'rotate' and 'move,' as most mouse actions in one of those modes typically involve moth moving and rotating the camera. The modifier keys are partly being passed based on `alt tool` and `meta tool,` and some of the combinations end up being non-intuitive.

If we move all of the new functionality into a single mousehandler and remove the 'alt tool' and 'meta tool' settings in ViewerCanvas we can:
- Build a single, unified hierarchy of modifiers and click events that will work even with a single mouse button: eg,  `<ctrl><alt><shift><drag>` would be equivalent to `<ctrl><alt><right-drag>` would be equivalent to `<scroll>
- The classic view tool behavior would be left in the existing tool classes, and would be available by clicking the tools in the palette.

Right now, I think that its safe to assume that a user _has_ a real keyboard. Upgrades that would allow the more advanced navigation modes to work well in a no-keyboard environment (Tablet) are way out of scope. There would need to be some major restructuring before any other part of the interface would even begin to be workable in such an environment.

peteihis:
> dummy constructor: `private TriangleMath() {}`

It is there already ... But FastMath does not have one, so I'm wondering if it is necessary after all...

I began to think one other thing, though. Now the cross product is a private method, that is used as the starting point to get the surface normal, surface area and barycentric coordinates. But there just might be cases when it could be needed as such... I'm not sure, but what if someone needs a vector field, that contains both the size and the direction of the triangle in one variable? Say calculating air-drag or something of that sort....

I think I'll do some renaming in there and make it public .. say add `tri` or `triangle` to some of the method names.

> These classic tools would completely ignore the navigation modes.

I was thinking that since the navigation mode selection is there, the tools should obey the modeling modes. At least as a user I would expect them to. ... Or then there would have to be a mode that somehow combines them both -- like a tiltable tray. _(I'm planning to make the tray mode tiltable anyway but not before some more urgent things. Probably not a very big task but the exact way it works needs to be figured out first.)_

It came to my mind, that just for the time being (and to avoid double coding)  it might be possible to make the icon activated tools and the mouse tools/handler use the same code but so, that the code recognizes who is calling it and provides only the appropriate functions.... Whether that is in the current kind of tools or a MouseEventHander, that would for one thing make sure that the logic stays consistent and another, that there would not be need to keep two copies of the same code.

> Tablet

Just to mention: I seem to remember that there was a discussion on SF about tablet use. As I recall they said that surprisingly much was working "out of box". _I have hard time trying imagine modeling on a touchscreen device but actually I remember somebody dreaming of a giant touchscreen -- I think he meant pen operated -- for modeling/drafting already in 1990. So, nothing new... :)_

Sailsman63:
How did I miss that? I would have sworn that I specifically checked the constructor. Haveing a private constructor prevents other users from accidentally creating an instance of the class. Fast math probably should have one as well, but the lack does no harm, until someone forgets...

I do remember the discussion by a tablet user, but you'd need a big tablet to use all of the editing interface as currently implemented. My point was that I don't see a good way to make the new modes work properly without a keyboard, so they don't need to have a 'click to activate' option.

peteihis:
I'm almost done with a tweak, that emulates the build-up, where the tool palette icons activate simpler view tools and the rest is handled by a more complex handler. (I'm not even sure how many files were edited). It works so that selecting the tool disables the navigation controls and using the tool switches the navigation to mode 0. This way only the view, that was handled is affected.

Originally I built that system so, that the meta- and alt-modifiers were ignored entirely, when either of the tools were selected, but that felt a kind of blunt, so I disabled those a few lines that did it. (In `mouseDragged()` methods of the tools.)

Actually I'm realizing that this thing is feature wise almost ready in regard to my original list. Just some minor thing to take  care of. Some things have come up along the way and there are some requests that I'd like to make happen too, but those should be a different set of features....

Expecting to push the current thing latest tomorrow.

peteihis:
I manged to fix the up side down flipping tray mode

peteihis:
Now, I thinks it is very close to how it would work with the Move and Rotate tools activated by the 1st mouse button and the MouseEventHandler-to-be activated by the other buttons or alt/meta modifiers.

If the view is in either of the travel modes using the 1st-mouse button switches it to modeling but the user has to switch back to travel. (Making the UI remember, what the user selected setting was seemed a bit of extra work, though, that's how it probably would work.)

Sailsman63:
From a user standpoint, this is looking pretty good. One thing that I noticed is that when using the "Move Tool" and "Rotate Tool" the projection is forced back to Parallel when dragging the view. Is this intentional?

peteihis:
> the projection is forced back to Parallel when dragging the view. Is this intentional?

Not exactly. Or let's say, that I did not think of that. 

What happens is, that while it is in Space or Tray modes it is using the perspective setting that the user had selected before switching to Fly or Drive. So if the view was already set to Perspective before going to a travel mode it would stay perspective while dragging.

You are right though, that it looks a bit of an unnecessary change. I'll have a look at that. It should not be too difficult to build another temporary switching there.

peteihis:
Go that covered and managed to simplify the logic a bit. And fixed a related a bug, that I had earlier forgotten.

Sailsman63:
Okay, I guess my question would be: why are 'fly' and 'drive' mode forced into perspective mode to begin with? I get the feeling that you said something about that earlier, but I can't find the note, and I'm not sure that I understood that even then....

Otherwise, I think that this is as ready as it's going to get without some real-world feedback. Is there anything else that you feel needs _immediate_ attention before merging?

I also notice that you seem to have copied the changes from #22 into your latest commit here... I'll probably clean that out before merging...

peteihis:
> ... why are 'fly' and 'drive' mode forced into perspective mode to begin with? 

The idea of those modes is, that you are traveling inside your model space -- not looking at is from the outside as you are doing, when you are modeling something. That does not work in parallel mode, where, theoretically, the viewing point is infinitely far from your point of focus.

What is "Zoom Closer" in modeling modes becomes "Travel Forward" in travel modes and rotating does not happen around the point, that you are working on, but instead makes _you_ turn. 

I guess it would be possible to create a kind of a "pseudo parallel" mode for traveling, but then you'd have to control the camera location and scaling separately and you'd lose the sense of distances which a kind of is the thing here.  -- Actually I'd like to do just the opposite and make the perspective wider in travel modes wider.  That would give better sense of the environment. To test that you can run this script:

`window.getView().getCamera().setDistToScreen(8.0);`

The travel modes begin to make sense in cases when you are, say moving inside building or roaming the streets of a city... Otherwise not so much.

> Is there anything else that you feel needs immediate attention before merging?

Not really. There are a few things, that should be done before a "full decimal release" and a few more that would be nice improvements, but all the essential things for feedback are now in. 

> I also notice that you seem to have copied the changes from #22 into your latest commit here... I'll probably clean that out before merging...

Well actually it happened more like the other way, but feel free to do as you see best. :) I might comment on the labeling from the _international point of view_.

peteihis:
What about that idea for an introduction video? Should I put something quick together? Just focusing on the essential things, not going _the Blender way_?

peteihis:
I discovered another bug. 

Sometimes scroll wheel zoom on a camera window stops working. When the problem appears, it seems to follow the scene but switching the view to another orientation (other than 'Other') and back to Camera may help. 

With a new scene things work as expected. Also the <Ctrl> + right mouse button zoom works when the scroll wheel is blocked.

Will investigate, now that I have a defected scene.

Sailsman63:
If the is a difference, it's out of scope for this project.

How do you feel about calling this 'done enough' for a nightly release? I want to take care of a couple of merge conflicts, but otherwise I think this is ready for a wider test audience.

peteihis:
> If there is a difference, it's out of scope for this project.

That's right -- just 'making notes'. 

It looks like there is some _historical weight_ here and there, but getting rid of that is slow work. Though as I have seen necessary once in a while.  A lot of questions start to raise, when you _(meaning myself)_ sit at this thing long enough. .)

> How do you feel about calling this 'done enough' for a nightly release?

Absolutely! :)

peteihis:
> What changed that it would matter?

It's hard to say. I guess there are now new performance requirements, that occasionally reveals issues, that have been suppressed before. 

The `Camera.perspective` can probably be eliminated entirely, as it looks like just a mirror of the `ViewerCanvas.perspective` to me. -- Though there are a couple of more things in the interactive rendering that seem to as for attention.  _(I has been working fine for a long time, but how it does that still seems to contain some strange wiring...)_ I may get back on those in one or more of the possible futures.. ;)

peastman:
> The Camera.perspective can probably be eliminated entirely, as it looks like just a mirror of the ViewerCanvas.perspective to me.

`Camera` gets used in a lot of places, not just in `ViewerCanvas`.

peteihis:
`Camera` gets used in a lot of places, not just in `ViewerCanvas`.

That may be true but the only thing I can find, that uses `Camera.isPerspective()` is the `SoftwareCanvasDrawer` and that only in the `renderLine`methods. The rest of the SWC checks the same information from `ViewerCanvas and the corresponding GL methods use the transformation matrices of `Camera`.

`Camera` uses it's `perspective` parameter internally though but I'm not sure why it does not seem to be a problem, that the two `perspective`s sometimes have opposite values.

peastman:
It gets used internally in methods like `findScreenBounds()` and `visibility()`, and those get called by classes other than `ViewerCanvas`.

When do the two fields have opposite values?

peteihis:
> When do the two fields have opposite values?

At least, when a `SceneCamera` is set to parallel viewing, `ViewerCanvas.perspective` goes false and `Camera.perspective` true (checked on 3.0.3). However that is the case, when everything works as it should and ViewerCanvas is following SceneCamera. Wireframes may get clipped in a strange way though by `SoftwareCanvasDrawer`.

I can't remember, when I first ran into the mismatch, but I was testing something using just the public methods of `ViewerCanvas`. I have no idea, what triggered that at the time. Everything seemed to work though and I forgot it.

Now the last time, it must have been my own doing somehow (in the code). I don't remember that I had done anything special when it happened. Just rotating and zooming the camera view and all of a sudden something snapped and obviously `Camera.perspective` had been turned false, while the rest view and scene camera had their perspectives set to true. The camera view (SWCD againn) started to see behind the camera and during zooming some of the scene was enlarging and some shrinking ... The setting got saved & reloaded with the scene and the fix was simply to set the view to another orientation and back.

Sailsman63:
This discussion is a bit concerning to me, as if we are modifying something without fully understanding it. Just in case it was not clear, my previous comment was specifically referring to 	cff8359. In looking a bit deeper, what your extra return statement is doing is bypassing the 'highlight selection' code when in rendered mode, which we really don't want to do.

I also found through testing (without the changes from commit 	cff8359)  that somehow the render order is off. If one adds items to the scene, the rendered view does not update to display them, but will display the item selection handles. If one then opens a different file, the items from the original scene are rendered, rather than the new scene.

The commit that caused this bug is 079d89d. Have not yet tracked down the root cause.

peteihis:
> Just in case it was not clear, my previous comment was specifically referring to cff8359.

It was not clear... :) Sorry.

> In looking a bit deeper, what your extra return statement is doing is bypassing the 'highlight selection' code when in rendered mode, which we really don't want to do.

I did not realize, that those were supposed to be displayed in a rendered view, but not that you say it of course. This seems to have been a by pass I made for testing purposes at the time. I have a fix ready, should I push it? `SceneViewer" and `ObjectViewer` both have one line removed and one added.

> If one adds items to the scene, the rendered view does not update to display them, but will display the item selection handles. 

To me they appear normally BUT if you have the ambient color set to black and the new object ends up in the shadow of an existing one, then you don't see it. (TroyTools sets ambient color dark.)

I have not touched anything that should affect the rendering order or depth evaluation in the canvas drawers or the renderers.  Though, I have a scene with a scripted object, where on a rendered view the scripted object is some times not shown. That happens with 3.0.3 too. 

> If one then opens a different file, the items from the original scene are rendered, rather than the new scene.

Does not happen to me either. I don't know hot that could even be possible? The rendered view is rendered from the scene that is behind the parent window.

peteihis:
> This discussion is a bit concerning to me, as if we are modifying something without fully understanding it.

To clarify this one thing: The message on the last commit "ViewerCanvas-Camera parameter mismach fixed" is actually a little bit wrong. It only changed the the scroll event handling method -- it did not fix the actual parameter mismatch, that seems to be difficult to repeat.

peteihis:
> This discussion is a bit concerning to me, as if we are modifying something without fully understanding it. 

The message on 3ba16c475963cf901e47f803b25f8a7518cf7fa7 is a bit misleading: It is not fixing the occasional parameter mismatch but instead changes the scroll event handling method, that was affected by it. The discussion is to me more just collecting the pieces, that may be involved...

> Just in case it was not clear, my previous comment was specifically referring to cff835922eeef010c4dedfbe65da3872b99542c1.

It was not clear. :)

> In looking a bit deeper, what your extra return statement is doing is bypassing the 'highlight selection' code when in rendered mode, which we really don't want to do.

OK. I had missed, that the highlights and overlays were supposed to be drawn on the rendered view too. The `return` statement in the `ObjectViewer` lead me to think, that it should be in the `SceneViewer` as well but that was not the case.

The culprit to make the rendered view disappear was the method, that draws the `AuxiliaryGraphics` _(=view cones)_. For now it would have been possible just to disable the drawing of the view cone on a rendered view, but I found the error and model code to doing it right _(=the active view_). I tested the fix to my best understanding and it seems to correct the behavior. Changes pushed: cff835922eeef010c4dedfbe65da3872b99542c1

> If one adds items to the scene, the rendered view does not update to display them, but will display the item selection handles.

This is not happening to me. The handles are drawn first and the image gets updated soon after. If you draw several objects in a row, the rendering restarts every time you draw a new one and finally updates the view after you pause. If the ambient color is set to black and the new object ends up in a shadow, then you won't see it.

I have not touched anything, that should affect the rendering order.

> If one then opens a different file, the items from the original scene are rendered, rather than the new scene.

This one is not happening to me either. I don't see how it could, as the rendered view is rendered from the scene behind the parent window? Though it looks as if the rendering of the rendered view would be delayed sometimes until you do something in that view. I can't explain that.

The commit 079d89dc21b72bea587ae1e17e9c8224d71be099 was the one where the view cone creation was turned into an inner class in ViewerCanvas and that did cause the problem with the rendered view on SceneViewer.

This is how it appears to me now ( .mp4-file):
[AoI-VM-RenderedView-test.zip](https://github.com/ArtOfIllusion/ArtOfIllusion/files/822671/AoI-VM-RenderedView-test.zip)

Sorry for pushing your patience, I sure hope this thing could be wrapped up soon.

Sailsman63:
The fixes seem right. I think that somehow the view cone stuff was messing up render order, but its now functioning correctly.

>Sorry for pushing your patience, I sure hope this thing could be wrapped up soon.

Don't worry about it. We're all learning. Though, this is a good illustration as to why it is good to have a defined goal right from the get-go. Much of the back-and-forth would have been much simpler if we had been reviewing smaller features in isolation.

__I've re-based again,__ to clean out some of the stuff that should not have been added in this branch. To check it out, _instead of_ `git pull` you will want to do `git fetch` and then `git reset --hard e9ed0a4` I'm thinking that the viewmanipulation project is now as done as its going to get. If you agree, I'll merge it tomorrow.

peteihis:
Ouch ... I noticed, that the MoveViewTool was responding only at mouseRelease. This is a fix to that. Just one line that I had forgotten the last time, when I was being extra careful, not to touch anything unnecessarily ... 

> If you agree, I'll merge it tomorrow.

Please do. :)

There are some "features inside features" where some work should be put in but those are minor things in the bigger picture and better handled separately. -- And as it seems, more come up along with the progress, so it's best to do the big merge now.

> .... smaller features in isolation.

At the time I started, it felt easiest to start just by adding the things I had ready, but from project management point of view that was the worst approach ever. It was a bit of a surprise to me, how deep into the existing software I had to dig at times and sometimes the first plan to implement something was not the working one. Next time I'll know better... :) -- And it's probably unlikely, that a make over in this magnitude would happen any time soon.

peteihis:
> I'll merge it tomorrow.

What can I say. Thanks for the 'tomorrow' .... again ... (oops)

I noticed, that I had once again forgotten the "mini windows" like tube creation. Easy fix to have move and rotate back working on those. The `UVMappingWindow` needed some special treatment in the tools. It implements `EditingWindow` but does not seem to obey all the same rules as the rest of them.

And NOW I'm taking my hands off it until it's merged!

Sailsman63:
Well, lets see what the community thinks!

PR #15, opened 2016-11-09
Refactoring: Convert plugin event calls to foreach style in main App class

makiam:
just a code clean

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
+++ b/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
@@ -132,7 +132,6 @@ public class ArtOfIllusion
     classTranslations.put("artofillusion.tools.tapDesigner.TapLeaf", "artofillusion.tapDesigner.TapLeaf");
   }
 
-  @SuppressWarnings("CallToPrintStackTrace")
   public static void main(String args[])
   {
     Translate.setLocale(Locale.getDefault());
@@ -273,7 +272,6 @@ public class ArtOfIllusion
   }
 
   /** Create a new window for editing the specified scene. */
-  @SuppressWarnings("CallToPrintStackTrace")
   public static void newWindow(final Scene theScene)
   {
     // New windows should always be created on the event thread.
@@ -323,7 +321,6 @@ public class ArtOfIllusion
   }
 
   /** Close a window. */
-  @SuppressWarnings("CallToPrintStackTrace")
   public static void closeWindow(EditingWindow win)
   {
     if (win.confirmClose())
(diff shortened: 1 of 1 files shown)


ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
```diff
@@ -371,51 +367,50 @@ public static void quit()
...
     System.exit(0);
   }
 
   /** Execute all startup scripts. */
-
+  @SuppressWarnings("CallToPrintStackTrace")
   private static void runStartupScripts()
   {
     String files[] = new File(STARTUP_SCRIPT_DIRECTORY).list();
+    if(null == files) return;
```
peastman: I know it's just nitpicking, but can we maintain a consistent code style?  I think that helps readability.  That should be

```Java
if (null == files)
  return;
```

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
```diff
@@ -1,4 +1,5 @@
 /* Copyright (C) 1999-2013 by Peter Eastman
+   Copyright (C) 2016 by Maksim Khramov
```
Sailsman63: For clarity, I'd suggest that Copyright notifications for edits should start "Changes Copyright..." as your copyright only applies to the code that you've added or changed. I know it's another nitpick, but better to get things figured out before it gets us in trouble...

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
```diff
@@ -272,7 +273,7 @@ public static void newWindow()
   }
 
   /** Create a new window for editing the specified scene. */
-
+  @SuppressWarnings("CallToPrintStackTrace")
```
Sailsman63: Another thing I was wondering about: @peastman, do we want to get into the habit of suppressing compiler warnings?

peastman: If you're really certain the warning is inappropriate, it's probably a reasonable thing to do.

Sailsman63: And that is the core of my question. I know that we tabled the discussion about using a logger for the time being. Currently, a lot of AOI code uses printStackTrace() at the time an exception occurs. If we want this to be the "Official" standard way to handle exceptions, we should run through the entire codebase, suppress this warning on any method that contains a `catch` block, and then modify the launchers to redirect stderr to a file.

If we decide to do something different for logging, the warnings are good in that they remind us of where some of our legacy code lives. Either way, we should be consistant.

Sailsman63:
Thanks! Merged.

PR #16, opened 2016-11-12
Implement Scene class method to get scene cameras list

makiam:


peastman:
Mostly looks good.  Just a few minor points.

Let's keep the formatting consistent.  That means a space before the parenthesis in `if` and `for` statements.  So `if (cameras.size() <= currentCamera)`, not `if(cameras.size() <= currentCamera)`, and similar in other places.

For test classes, the convention is to have one for each class that's being tested and to call it `Test<classname>`.  So your test would go into a class called `TestScene`.

The test itself would be a bit more rigorous if the scene contained something more than a single SceneCamera.  If you implemented `getCameras()` to simply return `getAllObjects()`, the test would still pass!  And then you don't check to see if it returned the right objects, just that the length of the list is correct.  How about adding several objects to the scene, some of them SceneCameras and some of them other objects, and check that it returns the correct ones?

Sailsman63:
Thank You! Merging.

PR #18, opened 2016-11-17
Move GUI code out of Scene class

makiam:


Sailsman63:
I see where you are going with this. Looking like a reasonable refactor.

 It seems that you created a new branch on top of your 'foreach' branch, rather than the current master branch. This will cause history issues.

In your *local* repository, please:

- check out a new branch *based on the current master*
- cherry-pick your commit b5e6351 on to it
- Git log should then show you the commit history of master as currently shown here on github (including exactly the same commit hashes) and on extra commit for this GUI move

Once that looks good, you can git push -f up to this same branch.

makiam:
As far as I remeber I branch this move-gui-.... from my local master but it alredy contains 4 commits pulled from upstream repository.
What is the proper way to sync repository with upstream repository to avoid this?

Sailsman63:
It looks like you may have merged your for-each branch down to your upstream branch, and then merged master on top of that. 

When branches are merged into master, there are some differences from the original commit signatures, first because a merge commit is added on top, and second because sometimes I rearrange commits a bit - in this case, I squashed all three of your original commits in the for-each branch down to one commit.

Your best bet is to keep your upstream/local master branch for *pulling from repository master only.* Don't make commits to it, and don't merge *anything* into it, even if we merge a pull request. Instead, pull down the merged changes from master, and then only use it as a starting point for new branches.

In the current state, if you local master matches the master on your github fork, you want to get rid of commit 649a956. Don't use `revert.` Instead, you specifically want to:

- `git reset --hard 788b952.`
- From there, you can `git push -f` to update your github copy, if you wish, though strictly speaking, you probably don't need to maintain a github copy of your master branch.
- `git checkout -b <new branch for moving GUI out of `scene`>
- `git cherry-pick b5e6351` to get the GUI move changes into the new branch. Note that the commit in the new branch will have a different signature. This is expected.
- `git push -f` to your github move-gui-out-of-scene branch, to update the pull request.

From this point, the only way that your local master branch changes is to *pull* (not merge) from artofillusion:master. Any other changes will cause issues with history compatibility.

Issue #17, opened 2016-11-14
CreateCurveTool drawOverlay() renders only last point multiple times


makiam:
See code lines from 96 to 101
We iterate over clickPoint array but at line 98 pos value always evaluates from clickPoint.lastElement()

makiam:
Possible solutions:
1. Draw only last element so no for loop needed
2. Draw all point nodes
3. Draw all points and highlight the last one

Sailsman63:
Hmm... 

Drawing all nodes may make visualizing the curve structure easier, but may also imply that the various points can be edited, which is not the case.... I'll have to think about that one a bit.

Sailsman63:
After considering this for a bit, I believe that we should draw all of the points. The last one, of course, should be the "active" color. The rest should be the "disabled" color.

Sailsman63:
Fixed in PR #21

PR #21, opened 2017-01-17
Fix for #17 issue: Wrong curve overlay nodes render

makiam:


ArtOfIllusion/src/artofillusion/CreateCurveTool.java
```diff
@@ -93,12 +101,22 @@ public void drawOverlay(ViewerCanvas view)
...
+      return;
+    }
+    int lastPoint = clickPoint.size()-1;
+    
+    Mat4 cameraWorldToScreen = cam.getWorldToScreen();
+    
+    for (int listIndex = 0; listIndex < clickPoint.size(); listIndex++)
+    {
+      Vec2 screenPos = cameraWorldToScreen.timesXY(clickPoint.get(listIndex));        
+      Color handleColor = (listIndex == lastPoint) ? ViewerCanvas.highlightColor : ViewerCanvas.handleColor;
```
Sailsman63: ViewerCanvas.handleColor should be ViewerCanvas.disabledColor, since the handles cannot be selected or moved.

Otherwise, it looks good. Good catch.

PR #22, opened 2017-02-21
Adding 'Unselect All' to LayoutWindow, Edit and PopUp menus.

peteihis:
This is a feature I have missed many times.

Sometimes on a crowded scene it is almost impossible to find the empty space to click to have nothing selected.


Sailsman63:
I like, I like... 

I might prefer to phrase the English label for this as "De-select all," or even better "Clear Selection." The later, though, overlaps with "Clear," Which in my opinion would be better termed "Delete." @peastman Do you have thoughts on the labeling here?

Give us a couple of days to think through the labeling, but I like this functionality! Thanks!

peteihis:
If you'll allow me to intervene.... As a non-native English speaker. I often use software, where the UI is in English. Some times a Finnish translation may not even be available. Sometimes it is just the house policy.

I think, I have seen the `clearSelection()`-function called things like "De-Select All", "Unselect" or possibly even "Empty Selection", depending on who wrote the English translation. "Clear Selection" though, does seem to have the meaning of "Remove Selected Objects" somewhere in the back.... Which makes it a bit scary to use.

The 'Clear' vs 'Delete' then ... 'Delete' of course has the meanings of 'eliminate' and 'destroy, which may sound a kid of violent to some. Sometimes, on the earlier days, the Finnish translation actually was "tuhoa" (destroy), but nowadays everybody seems to have settled for "poista" (remove). Sometimes maybe "pyyhi" (erase, wipe) _(in text or image processing that'd work)_.

The 'Clear' instead... One online dictionary, that I checked, gives 28 different translations to clear. That includes everything from washing windows to letting go of something or to harvesting wood for forest industry.... Personally, (... and I find this somehow funny, but I can'r help it... ) there is always this small voice somewhere in the back of my mind wondering _"Does this make my objects transparent and shiny?"_ :D

That's my humble view on this subject. :)

makiam:
What about "Select None"?

Sailsman63:
'De-Select' and 'Unselect' are grammatically identical in English, (This point is sometimes disputed, but for practical purposes, we can treat it as so) and both understandable to a native speaker. This is one of those 'irregular' bits of the English language where the preferred form is memorized on a case by case basis for the most part.

For example, we have 'untie' and 'undo' but 'debug' and 'disassemble.' 'Unselect' just _sounds_ wrong to a native English speaker. The other forms that you have seen were almost certainly composed by a non-native speaker. 'Empty Selection' makes sense as an English phrase, but is not common in Software UI.

'Delete' sounds _violent_ to a non-native speaker? That's a surprise to me. It implies making something cease to exist, but without the fuss and cataclysm that 'kill' or 'destroy' would entail.

'Clear' has quite a lot of connotations, depending on context, which is why its tricky to get right. It can be both transitive. "Clear <items> away, ie remove the item" and intransitive, "Clear {things out of- implied} <item>, ie remove things that are interfering with the item." I believe that it for computer program UIs, its more common in a mac environment, so not so surprising that Peter picked it in the first place.

@makiam I think I like 'Select None' - its clear, and does not have the overlap issues that a variation on clear could have.

peteihis:
> 'Delete' sounds _violent_ to a non-native speaker?

Maybe it's just me then recalling some early, not so skillful, translations, years back. I think most understand it as, what you said,"Remove (so it does not exist)" or "Erase".

Actually the documentation of 'clearSelection()' says "Deselect ... " That should have pointed me to the right direction, but I was not paying attention to the comments this time. Even though, I have rewritten or elaborated some of them elsewhere.

Sailsman63:
>I'm not sure but 'Select None' seems faintly familiar too... At least next to 'Select All' it would make sense.

That's because it is a fairly common option. And now that you mention it, we do have `Select All`. I think that `select none` would be the best English translation, and since it came up, it should also be grouped with `Select All` in the menus... easier to understand in context that way.

peteihis:
I just came to think, that in the pop-up menu "Select None" would look a bit isolated. In that sense the "Deselect" might be more universal. 

I think this should be added to the object editors too, incliding the pop-up menus there ans in the edit menus all the selecton manipulaiton should e fenced together (Extend selection, Invert selection, ...)

Sailsman63:
I'm not sure that this would be as important in the object editors, but certainly worth considering. I'd like to consder grouping commands at some point, but I'd like it to be extensible - with plugins, etc, stating that they need to add a command to the 'selection' group of the 'edit' commands, without trying to figure out exactly where in the menu the commands appear.

Sailsman63:
Okay, I had to do some rebase-ing weirdness, as I could not get this to merge cleanly. The changes have been added, however. Thanks for getting started on this one.

peteihis:
Did you leave the addition to th ePopUp window aout on purpose? To my experience it would be more useful there than in the Edit menu.

Anyway, I think I'll "reopen the case" soon an see about renaming the function in the code.

Sailsman63:
Umm... No,I thought all of it was included. I also thought I had tested. I'll re-look later, not sure how that happened. No need to rename the function in code...

peteihis:
> Umm... No, I thought all of it was included

Oops, sorry! 

It was there, of course. I was only looking at the last changes. -- Anyway I need to change the location of it in the popup menu too to where the other selection related commands are.

peteihis:
Oops... Sorry!

It is all there, but somehow the merge showed up on the network graph as just your commit. My branch still appears/ed as a separate unmerged branch.

Anyway, I'm planning on moving the command in the PopUp menu to the same group with the other selection commands, next to Select Children.

Sailsman63:
Ahh... Understandable. I did a local rebase and merge, and then pushed the results directly, rather than `push -f` up to your branch and merging from there. I was having trouble with the merge -the diff was showing as incompatible. For some reason, parts of the system thought that big chunks of `LayoutWindow` had been changed, and it appeared that the file had gotten a whole bunch of `CR` characters at line endings, rather than `LF` or the windows standard `CR+LF`

peteihis:
> For some reason, parts of the system thought that big chunks of LayoutWindow had been changed.

I had that onde earlier resetting and rebasing the viewmanipulation branch. Strangely the parts that tha it suggested to select from were not even the same size, but the end of them was some 70 or so lines different....

> and it appeared that the file had gotten a whole bunch of CR characters at line endings

At some point I did get warning, that new-lines (or something) would be replaced by CRs ... I have no idea why.

Sailsman63:
That may have been because the earlier commits to viewmanipulation were based on a wrong `master`, without the `.gitattributes` file functioning properly?

It's sort of a mess out in the larger world, with no good, universal way to handle things.

PR #23, opened 2017-02-25
A random direction unit vector generator

peteihis:
This class is a  generator, that can produce randomly aimed 3D unit vectors. The distribution is even to all directions in 3D-space unless you wish to "spray" the vectors into a preferred direction.

This avoids the problem with aiming vectors in "Earth coordinates" (360 by -90 to +90 degrees), that the concentration is higher by the poles.

In the naming I followed the "FastRandom" to call it "Vec3Random".

To test it I used and altered this script:

```
//Vec3Random g = new Vec3Random();
Vec3Random g = new Vec3Random(2); // needs to be defined per script if with seed
r = 0.025;
s = new Sphere(r, r, r);
spray = new Vec3(0,0,0);

//g.setSeed(1);
//g.eraseSeed();

//println(g.getSeed());
//println(g.hasSeed());

for (i = 0; i < 100; i++){
	v = g.getVector();
	//v = g.getVector(spray);
	//println(v);

	c = new CoordinateSystem();
	c.setOrigin(v);
	window.addObject(s, c, "dot", null);
}
```

**Small edit above for a more fluent sentence, I hope**

### DIFF
--- a/ArtOfIllusion.xml
+++ b/ArtOfIllusion.xml
@@ -70,7 +70,7 @@
         windowtitle="Art of Illusion Documentation"
 	public="true">
       <doctitle><![CDATA[<h1>Art of Illusion</h1>]]></doctitle>
-      <bottom><![CDATA[<i>Copyright &#169; 1999-2011 by Peter Eastman.</i>]]></bottom>
+      <bottom><![CDATA[<i>Copyright &copy; 1999-2011 by Peter Eastman.</i>]]></bottom>
     </javadoc>
   </target>
 


peastman:
I notice that you're using tabs for indentation.  I strongly recommend using spaces instead, since tabs will get formatted differently on every computer.

peteihis:
Yes, you have a point there.

I use Notepad++ for editing, where the default, tab = four spaces makes it very nice to use and to me very readable too. -- I think, I can put the entire library through a machine, that will take care of the of the substituting.

Another question: I have added portions to the already existing files, where I have used multiples of four spaces instead of two. Is that acceptable or should those be unified with the 'old school style'?

Sailsman63:
You may even be able to set Notepad++ to save tabs as 2 spaces, rather than a tab character. There is also a way to get git to clean this up on commit, but it relies on some things that are platform-specific, and that need to be set up in your local copy of the repository.

I'm not going to ask you to go back and change the spacing on some of your other PRs now. but in future work it would be best to keep to the 2-space indent.

Do you have a particular use in mind for this file, or is this more in the nature of an experiment? I ask, because utility code that is generated without several clear use-cases generally need to be re-written when someone actually does try to use it.

peteihis:
In your consideration, should this be taken care of befor the 'nightlt build'? -- I won't be home until Monday.

> Do you have a particular use in mind for this file, or is this more in the nature of an experiment?

Things like the Patricle Jet, animating explosions, randomizing aim of projectiles, random "star fields" for background ... 

It's been a while since I experiemented with the idea. There was not an exsiting way of producing evenly distrbuted random directions and this principle seemed to work as expected. Now it contains just the very basic case(s), but I have some ideas on adding more 'radiation patterns' an maybe a noise pattern option as well... :)

Sailsman63:
This particular file can wait a bit, no rush. Once I get the automated build system up, we will be able to do a nightly every time we've added something interesting.

peteihis:
> This particular file can wait a bit, no rush.

So, the rest then?

Sailsman63:
No, there have been enough history re-writes with the viewmanipulation project. There are other, legacy bits of the codebase that don't follow the pattern either. I may eventually figure out a way to go back and fix all of them at once, but not right now.

peteihis:
Oops ... I did'n mean that to happen. :( -- Seems that the file was lost in the process too.

Unfortunately I had accidentally deleted all my local copies as well in the "great clean-up", I did a while ago _(had "one too many" local repsitories after all my experiments)_ though I usually keep safecopies of everyting more than enough ... Pffffhhh...

PR #24, opened 2017-03-07
Default Mappings classes constructors

makiam:


Sailsman63:
How do you imagine the parameterless constructors being used? I am having a difficult time imagining a situation where one would want to generate a mapping representation without specifying a texture and object. What would the purpose be?

makiam:
The idea was unify plugins registration code in ArtOfIllusion main class. For future we can change this one-by-one manual registration with more generic scan-and-registry automatically code.

Sailsman63:
We already have an automatic scan for plugins that are actually plugins. I'm not sure that I see a reason to scan the core application. I'm questioning the way that we are storing these references anyway. Something to keep in mind for later.

PR #25, opened 2017-03-13
Declarative Editing Tools initialization.

makiam:
GUI options for EditingTools are configurable with java annotations.
Clear unnessesary method overrides.
Testcases for this changes.


Sailsman63:
The annotation method does seem to be somewhat less verbose, and more elegant. However, I've got some concerns about the PR as implemented:

- I'm not sure that a very specific annotation like this should be declared in a separate package. Would it not make sense for it to be a member of `ModelingTool`, as this is the only class heirarchy that will use it?
- The unit tests for the added annotations are unnecessary, and in fact test the contents of the translation file, rather than the code. They will break if run on a machine with locale set to anything other than english.
- Adding a `setPreferences()` method to facilitate the above really does not seem right. If you're adding a method to allow test access, either something is very wrong with the design on the code under test, or you are testing the wrong thing.

In short, if we want to move to a more declarative arrangement for editing tools and similar, we need to have a discussion about what it should all look like in the end, before starting on the work.

makiam:
> Would it not make sense for it to be a member of ModelingTool, as this is the only class heirarchy that will use it?
I see no reason not to use this annotations for other cases.
> The unit tests for the added annotations are unnecessary,
Created just to check that annotations are properly added to classes.
> They will break if run on a machine with locale set to anything other than english.
I forced to use in these tests Locale.ENGLISH

> If you're adding a method to allow test access, either something is very wrong with the design on the code under test
Looks right. As for now all code which depends of ApplicationPreferences.preferences instance cannot be tested. It's the reason I added this setter

Sailsman63:
What other uses were you thinking of for this annotation? That's actually the lesser part of my concern, as it could be argued both ways.

The more concerning part is regarding the tests. There is absolutely nothing in `ApplicationPreferences` that currently changes things that could reasonably be unit-tested. Testing an annotation by checking for data that is provided by a configuration file is going to be... fragile... at best. 

To test that your annotation does as expected, you would more likely create a small nested class inside your test class. This nested class should be annotated. Perhaps more than one such class, including one that is not annotated. You then `AssertEquals("Value Inserted in the annotation", NestedClass.class.getAnnotation(Tooltip.class).value());` At which point, you can tell that the annotation is working as intended.

makiam:
> There is absolutely nothing in ApplicationPreferences that currently changes things that could reasonably be unit-tested

Every current and future classes which is depends on ArtOfIllusion.getPreferences() call is not unit-testable because has dependency of uninitialized ArtOfIllusion.preferences field.

makiam:
Update AOI preferences initialization code. Remove unused setter for it

makiam:
Replaced with #285

PR #26, opened 2017-03-15
Deselect to editors & small edits

peteihis:
- Added the deselect function to editors
- Added keyboard shortcuts, `Ctrl+I` for invert selection and `Ctrl+D` for deselect. _(`Ctrl+N` is used for New File)_
- Rearranged the menus to show selection commands as a group.
- Added 'intelligence' to enable/disable selection related menuitems according to selection content.

The modified `.gitignore` keeps pushing itself to the commit ...


Sailsman63:
Thanks, Looks good.

I did fix the changes to the `.gitignore` and did `push -f` before merging, so this branch no longer matches what you have in your local history...

Regarding the changed `.gitignore`:

- run `git checkout <your master tracking branch>`
- run `git status`
- if the status output includes `modified: .gitignore` run `git reset --hard` and repeat the status check
- if it still displays as modified, please let me know

peteihis:
No `.gitignore` in the status output.

peteihis:
BTW, How is that .gitignore-business supposed to be handled?

I'd like to keep my own .gitignore in my local repository, but everytime I update the repo to match the upstream-master it is replced by the upstream-one and when I modify/replace the local one it sneaks into the next commit....

I'd be handy to make the last checkup build before git push in the reposityfolder but in my case that means a few .bat and .txt files in the folder.

Sailsman63:
If you are finding the need for a more specific `.gitignore` I suspect that you are using one of the power-tool options for `git add` in a way that is causing you issues.

`git add -A` will stage all changed and untracked files in the working tree to the index for commit. You should almost never need to use this particular option.

`git add -u` is much safer, as it stages any file that is already tracked, but has changes. This one you might use rather often.

`git add path/to/file` will stage the specified file. This is how you normally stage a new file for commit.

In a normal session, as you work up to a commit, you might use variants of `git add` several times. Its also a good idea, before commiting, to run `git status`. If you see a file staged for commit that should _not_ be commited, you can run `git reset HEAD path/to/file` to unstage it.

I'm hoping to refactor the build scripts soon, and part of that would include putting test builds off in a subdirectory, where their clutter will not cause confusion.

peteihis:
> `git reset HEAD path/to/file`

Looks like the command I have been looking for. Thanks. :)

PR #27, opened 2017-03-22
Updated dates and markdown files

johnroper100:


Sailsman63:
Thanks for the update! I hadn't even realized that the markdown parser had been updated. I chose not to keep the Copyright date changes, as they are specifically claiming copyright on the javadoc, which aside from a couple of auto-generated method listings, has not changed since the dates already listed.

PR #28, opened 2017-04-01
MIP-Mapped Image, grayscale losing data bug fix

peteihis:
That's all I could thing about and the fix works for me. 

Induced by the user 'Helge-H' on SourceForge. https://sourceforge.net/p/aoi/discussion/47783/thread/2c90fc28/

### DIFF
--- a/ArtOfIllusion/src/artofillusion/image/HDRImage.java
+++ b/ArtOfIllusion/src/artofillusion/image/HDRImage.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 2001-2009 by Peter Eastman
+   Modifications copyright (C) 2017 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -190,13 +191,13 @@ public class HDRImage extends ImageMap
       {
         if (width[0] < height[0])
           {
-            w = (int) (width[0]*PREVIEW_HEIGHT/(float) height[0]);
+            w = Math.max((int) (width[0]*PREVIEW_HEIGHT/(float) height[0]), 1);
             h = PREVIEW_HEIGHT;
           }
         else
           {
             w = PREVIEW_WIDTH;
-            h = (int) (height[0]*PREVIEW_WIDTH/(float) width[0]);
+            h = Math.max((int) (height[0]*PREVIEW_WIDTH/(float) width[0]), 1);
           }
       }
     int data[] = new int [w*h];
--- a/ArtOfIllusion/src/artofillusion/image/ImagesDialog.java
+++ b/ArtOfIllusion/src/artofillusion/image/ImagesDialog.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 2001-2005 by Peter Eastman
+   Modifications copyright (C) 2017 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -15,6 +16,7 @@ import artofillusion.ui.*;
 import buoy.event.*;
 import buoy.widget.*;
 import java.awt.*;
+import java.awt.image.*;
 import java.io.*;
 
 /** ImagesDialog is a dialog box for editing the list of ImageMaps used in a scene. */
@@ -27,10 +29,13 @@ public class ImagesDialog extends BDialog
   private BScrollPane sp;
   private ImagesCanvas ic;
   private BButton b[];
+  private BufferedImage bgImage;
+  private Color selectedColor;
 
   public ImagesDialog(BFrame fr, Scene sc, ImageMap selected)
   {
     super(fr, "Images", true);
+    createDisplayItems();
     BorderContainer content = new BorderContainer();
     setContent(content);
     parent = fr;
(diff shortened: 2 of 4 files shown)


peteihis:
The very wide but low images (256 x 1 pixels), that I used for testing, appeared as empty placeholders on the image pallette. This last one makes sure, that the preview images will always have at least 1 pixel of height and width.

peteihis:
There's a bug in the latest commit. I'm trying to figure out what it is. Somehow it appears at loading images, but not when loading an AoI-file.

peteihis:
The bug fixed. There seemed to be a timing issue that when the dimensions of the preview were checked right after creating, the getWidth() and getHeight() methods of Image both returned '1' in some caseses. Hence teh automatic calculation could not be ised but the preview dimensions are now precalculated.

Modified the images dialog form this:
dialog303
to this
dialog303plus

So, 
- The preview images are now centered in their boxes
- There is a checkered background
- The selection frame was refit (it was not quite centered)
- The selection color is the highlight color of the current theme

Sailsman63:
Can we hold on this for a bit? I need time to go over the original fix, because something is tickling the back of my brain....

peteihis:
> Can we hold on this for a bit? I need time to go over the original fix, because something is tickling the back of my brain....

You can't let the go of the tickling! :D -- Take your time. Do you want to know more in details, what I did to test it?

peteihis:
I ran some test and I think I know, what is happening. Long story short: **`BufferedImage` and 8-bit (256 level) grayscale don't mix!** Sadly the API documentation does not explain, what everythig is for and why it works the way it does.

It does not matter if the original is **_defined_** grayscale or RGB. If the data actually **_is_** grayscale `BufferedImage` will read it as `TYPE_BYTE_GRAY`, which can only handle 128 levels (7-bit or pseudo "unsigned byte")  and at that point half of the information is lost already. How the 7-bit information is presented is a different story.

The `MIPMappedImage` uses `ImageIcon.getImage()` --> `PixelGrabber(....forceRGB)` which gets all the 256 levels of a grayscale image too. Then, from that point on, `BufferedImage` must handle the information in `TYPE_INT_RGB`,  `TYPE_3BYTE_BGR` or higher modes to prevent data loss.

If you read grayscale data into a `BufferedImage` directly, it behaves as if there was a gamma correction in place. This is a view to both ends of the R, G and B values ontained that way:
~~~~
GRAYSCALE 
INPUT

  0,  0,  0
 13, 13, 13
 22, 22, 22
 28, 28, 28
 34, 34, 34
 38, 38, 38
 42, 42, 42
 46, 46, 46
...
252,252,252
252,252,252
253,253,253
253,253,253
254,254,254
254,254,254
255,255,255
255,255,255
~~~~
Then, if you push the data, collected by `PixelGrabber` into a grayscale `BuffedImage`, you get coarse steps in the black end of the scale, which was the case with `MIPMappedImage`:
~~~~
PXELGRABBER   GRAYSCALE
INPUT         OUTPUT

 0, 0, 0       0, 0, 0	
 1, 1, 1       0, 0, 0	
 2, 2, 2       0, 0, 0	
 3, 3, 3       0, 0, 0	
 4, 4, 4       0, 0, 0	
 5, 5, 5       0, 0, 0	
 6, 6, 6       0, 0, 0	
 7, 7, 7      13,13,13
 8, 8, 8      13,13,13
 9, 9, 9      13,13,13
10,10,10      13,13,13
11,11,11      13,13,13
12,12,12      13,13,13
13,13,13      13,13,13
14,14,14      13,13,13
15,15,15      13,13,13
16,16,16      13,13,13
17,17,17      13,13,13
18,18,18      22,22,22
...
~~~~
In this latter case the `TYPE_INT_RGB` or `TYPE_3BYTE_BGR` would preserve the original values.

Sailsman63:
So, what you're doing is forcing the grayscale to be saved and reloaded as an RGB image, which keeps the dynamic range. Makes me wonder if a `BufferedImage` is really the best fit here, but I don't know what the alternative would be.

Give me time to look over the preview tweaks - those should probably have been a separate PR.

peteihis:
> forcing the grayscale to be saved and reloaded as an RGB image, which keeps the dynamic range.

`MIPMappedImage` was already doing that otherwise (The `findComponentsMaps()` method and the line 669, ´bi.setRGB(i, j, 0xFF000000+(maps[0][0][index]&0xFF)*0x010101)´. The place that was different was the declaration of the BufferedImage, line 664, which unfortunately forced the saving format into 7-bit....

> Makes me wonder if a BufferedImage is really the best fit here, but I don't know what the alternative would be.

I found it strange, that BI doesn't seem to have a way to handle 8-bit grayscale a such... I guess it would originally have been possible to save the data as byte maps into the AoI-file, but BufferedImage and the .png format provide the benefits of built-in methods of Java.

> tweaks - those should probably have been a separate PR.

Next time then :)

Sailsman63:
I like the improvements to the image dialog interface. The main issue seems to be caused by the way that Java handles signed vs unsigned at byte level. (no unsigned primitives)

@peastman, if there is not a more elegant way to handle this, I think that I will merge these changes in a couple of days.

peastman:
Fine with me!  I also like the improvements to the window.

peteihis:
Thank you. :)

PR #29, opened 2017-04-16
Extract scene environment setup dialog to separate class

makiam:


Sailsman63:
This should be two separate PRs, and I have similar concerns about each.

The environment properties dialog only makes sense in the context of a specific `LayoutWindow` and should therefore not be a top-level class. (The class needs to know what scene it is working on. Are there any cases where you would want to launch this dialog from another location?) Creating a class for readability may make sense, but if so, it should be a nested class inside of `LayoutWindow`.

The `ColorSampleWidget` as well... Do you intend to use it in multiple locations? If not, it belongs inside of the ColorChooser.

makiam:
>  Creating a class for readability may make sense, but if so, it should be a nested class inside of LayoutWindow.

The idea is to split very big LayoutWindow class to smaller parts.
BTW... What is the proper way to made changing Environment properties undoable operation?

About ColorSamplerWidget... I create it as replacement for RGBColor code which returns similar widget. As I see color swatches used in many points, not only in ColorChooser.

Sailsman63:
`LayoutWindow` is rather large, and perhaps it could be refactored. However, starting by splitting out code that will always be used by `LayoutWindow` might not be the best starting point.

You might have more of a point with the `ColorSamplerWidget`. One of my longer term goals is to split the core scene and texturing system apart from the user interface. However, we should have a better idea of what the end result should look like before we start adding display classes that we are not currently using

makiam:
Current Widget from RGBColor now used in Directional, Spot, Point light dialogs, in Uniform Texture and Matrial dialogs, in ColorChooser and in two procedural Modules (Spectrum and ImageOrColor)

Sailsman63:
After stewing on this for quite some time, I'm convinced that this is not the direction that we want to go for refactoring.

PR #30, opened 2017-04-20
Xlint code checks

makiam:
Removed unnessesary casts
List declarations now more type-safe.
Access vectors as list's


peastman:
You made a lot of changes like changing

```Java
ArrayList selection = new ArrayList<ObjectInfo>();
```

to

```Java
List<ObjectInfo> selection = new ArrayList<ObjectInfo>();
```

That's a bad thing to do for various reasons.  First, it's less clear.  Different types of lists have different performance characteristics, so when you know what type it is, you should say that.  Also, invoking a method through an interface can be slower.  For example, if the JIT knows an object is an ArrayList, and knows that no subclasses override a particular method, then it can inline that method.  But if all it knows is that the object is a List, it can't do that.

Adding the generics is good.  But it should be an `ArrayList<ObjectInfo>`, not `List<ObjectInfo>`.

ArtOfIllusion/src/artofillusion/MaterialMappingDialog.java
```diff
@@ -39,28 +39,30 @@ public MaterialMappingDialog(BFrame parent, Object3D obj)
...
 
     // Make a list of all material mappings which can be used for this object and material.
 
-    mappings = new Vector();
-    List<MaterialMapping> allMappings = PluginRegistry.getPlugins(MaterialMapping.class);
-    for (int i = 0; i < allMappings.size(); i++)
+    mappings = new Vector<Class>();
+    Material mat = obj.getMaterial();
+    
+    for(MaterialMapping mapping: PluginRegistry.getPlugins(MaterialMapping.class))
```
peastman: Consistent formatting please! :)  `for (` and `if (`, not `for(` or `if(`.

Sailsman63: Thanks for pointing that out. I was going to mention similar. Its fine to use the interface if you are in a situation where several `list` implementations _might_ be supplied, depending on the calling code. In this case, though, its a private member, and you definitely know the concrete type.

makiam:
Effective Java 2nd Edition, Item 52: Refer to objects by their interfaces

        If appropriate interface types exist, then parameters, return values, and fields should all be declared using interface types. If you get into the habit of using interface types, your program will be much more flexible. It is entirely appropriate to refer to an object by a class if no appropriate interface exists.

Sailsman63:
That's true as far as it goes. The flexibility argument _only_ applies when there is a chance that the concrete object is passed in from outside, perhaps by another module. In the cases that we are discussing here, we have `private` local fields. They will never receive another type, unless we are completely re-working the class.

Sailsman63:
@makiam, I'd really like to be able to merge in these changes, but @peastman is right. Private, local fields should be declared for the most specific type possible for readability. In this case, `Arraylist<ObjectInfo> selection = new Arraylist<ObjectInfo>; or similar (I see Vectors and hashmaps that received similar treatment.) Would you be willing to add another commit that reverts to the more specific types?

makiam:
List Type declarations reverted back to specific one.
Cherrypicked commit from #36

ArtOfIllusion/src/artofillusion/MaterialMappingDialog.java
```diff
@@ -39,28 +39,30 @@ public MaterialMappingDialog(BFrame parent, Object3D obj)
...
     this.obj = obj;
     map = obj.getMaterialMapping();
     oldMapping = map.duplicate();
 
     // Make a list of all material mappings which can be used for this object and material.
 
-    mappings = new Vector();
-    List<MaterialMapping> allMappings = PluginRegistry.getPlugins(MaterialMapping.class);
-    for (int i = 0; i < allMappings.size(); i++)
+    mappings = new Vector<Class>();
```
Sailsman63: @peastman Here is another reason to move away from reflection if we can. The reflection call on line 51 will not compile if we try to Parameterise this vector to anything more specific than Class...

Sailsman63:
Thanks! We appreciate your work. I've added some tweaks, just to make sure everything is the way we like it.

Please keep in mind that when you do a null check, its much more readable if you use the form `object == null` rather than `null == object`

PR #32, opened 2017-04-23
ImageFileChooser to match read-supported file types

peteihis:
Changes:
- Removed tif
- Removed svgz -- Needs unzipping, which is not provided by `SVGImage` constructors.
- Added  gif
- Added more  jpeg-type extensios
- Could not verify if pic-format is really supported _(did not find any pic-images)_, but I left it there because it was in `ImageMap.loadImage()` in the same block with hdr.

### DIFF
--- a/ArtOfIllusion/src/artofillusion.properties
+++ b/ArtOfIllusion/src/artofillusion.properties
@@ -1051,12 +1051,14 @@ objectType=Object Type
 # Various File filters
 #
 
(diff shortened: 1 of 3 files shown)


peteihis:
And I don't know what the hell the .gitignore is doing in this PR again. It did not even show up in the committed files....

Sailsman63:
Thank You very much!

Issue #31, opened 2017-04-21
Hide Objects List panel command is not work


makiam:
The problem has two parts:
First one: menu selection code expects that command is under Scene menu. It's easy to fix.

Second one: the proper visibility toggling depends on other panel under same dock. If we detach Properties panel from dock toggling begin to work. If we move score panel to  right dock toggling for Objects panel fails and in this case toggling Score panel becomes invalid too

Sailsman63:
Thanks for reminding me of this. The first part is, as you stated, easy to fix.

The Second part has actually been the case for some time. Not as simple to fix. @peastman I think that this has been a problem since object properties were added as a dockable widget.

peastman:
I'm not sure how much sense that command even makes anymore.  It used to just adjust the position of the divider.  But now that there could be multiple panels in the same group, what should it do?

Sailsman63:
Not sure if we still need it. If we fix & keep, it should be re-named and it should apply to that whole group.

In testing, I found that the `setObjectListVisible()` method functions as expected, but that when there is more than one panel in the group, the command processor and the menu update cannot detect that the group is already hidden.

The buoy code here seems a bit convoluted, and I admit that I have not yet graphed out how all the components come together.

Sailsman63:
Are we okay with just removing this?

Given the lack of support requests on the subject, the average AOI user never notices that this exists. The reports that we have received all came from tech-y people who the to exhaustively test everything in the UI, even if they don't really use it.

If it's not used, is just taking up menu space, and broken UI elements give a bad impression.

peteihis:
Hey, my bad... The first part I mean. (Sorry I did not originally understant, what this was about.)

I forgot to move the command to the viewMenu block. Usually 3D-software have the option of hiding side panels, and since the command is there, I'd suggest to keep it. I can push a fix tonight. The command is necessary, because some plugin-added items (=Preview) in the panel can not be hidden by dragging the edge of the panel.

But (call it '#1b'): I suppose the expectation to be, that you could get the side panel out of hiding by the same menu item? Have to look what it takes, if I can come up something simple...)

And '#1c': The name should probably be something like "Hide side panel".

Looking at the code now, I'm wondering if it actually is necessary to have the menu recognized on the `actionPerformed()`? I'd expect the commands have the same effect, where ever they come form? Just have a long switch in there to handle all of them?

[...]

PR #33, opened 2017-05-14
Prevent unneccessary updateMenus calls on BSplitPane divider changes

makiam:
I remove event links which calls twice to updateMenus on each BSplitPane divider change.
Also small code format


Sailsman63:
Please don't do major whitespace changes piecemeal. If formatting in some files differs significantly from the project standards, open an issue specifically to track the formatting issues, and add filenames to it as you find them. We may eventually go through and fix formatting all at once.

The menu-state updates on splitter change events are used to change the text for Show/Hide score, and Show/Hide object list (The later is broken, as per #31, but the code that you removed is in support of the function)

PR #34, opened 2017-05-23
External Imge

peteihis:
The External Image-type and the UI functions to handle them.

This is a fully functional set, but:
- UI-Translations are missing from ImageDetailsView
- SoftReferences not yet implemented to preview images
- Meta data of old image types not implemented
- Preview dimesions recalculation in other than MIP-mapped images could be revised (if blocks can be replaced by m Math.min/max and 'int' by 'round').
- Some of the UI-icons are just placeholders...
- Some unexpected bahaviour, when a linked image file is moved to another location during session. (The reference may revert to an older reference if it has been changed or the link may actually follow the moved image file. I suspect this to be at least to some extent system dependent.)

### DIFF
--- a/ArtOfIllusion/src/artofillusion.properties
+++ b/ArtOfIllusion/src/artofillusion.properties
@@ -1051,12 +1051,14 @@ objectType=Object Type
 # Various File filters
 #
 
-fileFilter.png=PNG Images (.png)
-fileFilter.jpeg=JPEG Images (.jpg)
-fileFilter.hdr=High Dynamic Range Images (.hdr)
-fileFilter.svg=SVG Images (.svg)
-fileFilter.tif=TIFF Images  (.tif)
 fileFilter.images=All Supported Image Formats
+fileFilter.gif=Graphic Interchange Format (.gif)
+fileFilter.hdr=High Dynamic Range Images (.hdr)
+fileFilter.jpeg=Joint Photographic Experts Group (.jpg, .jpeg, .jpe, .jif, .jfif)
+fileFilter.pic=Pictor PC Paint (.pic)
+fileFilter.png=Portable Network Graphics (.png)
+fileFilter.svg=Scalable Vector Graphics (.svg)
+fileFilter.tif=Tagged Image File Format (.tif, .tiff)
 
 fileFilter.obj=OBJ Files (.obj)
 
--- a/ArtOfIllusion/src/artofillusion/ui/ImageFileChooser.java
+++ b/ArtOfIllusion/src/artofillusion/ui/ImageFileChooser.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 2005 by Peter Eastman
+   Modification copyright (C) 2017 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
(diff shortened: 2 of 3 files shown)


makiam:
IMO is cheaper way to react on Escape is registering keyboard action like this one 
this.getComponent().getRootPane().registerKeyboardAction(this, KeyStroke.getKeyStroke(KeyEvent.VK_ESCAPE, 0), JComponent.WHEN_IN_FOCUSED_WINDOW);

peteihis:
That seems to make sense. Thanks!... I copied the code from some other dialog template in AoI, but I did not check, what it actually was for.

I wonder in Ctrl+Z could be registered the same way? I think that at some point  it might be reasonabe to add a specific line of undo records for images, so that those undos could be accessed only through the image handling dialogs. I would not put image handling undos among the normal scene-undos as changes to the images are generally speaking not visible on the scene view.

I don't know that would how yet, but just thinking...

Sailsman63:
@peteihis Please keep the `escape` handling code in your new files as is. It's using some of the convenience methods that we get from buoy. @makiam's way has a long, complex daisy chain, using direct swing calls. Sometimes you cannot avoid using swing, but in this case, performance would hardly be a legitimate consideration.

peteihis:
Yeah, well....

It didn't look that complex to me, but as I just out of curiosity tested it, it turned out, that `BDilaog` does not implement `ActionListener` which is needed to use @makiam's line.

I also had another promlem: I was trying to add a mouse relesed event to tell, when resizing has ended. I tried adding a MouseReleasedEvent to the dialog and a MouseListener to the RootPane. Both compiled OK, but neither of them reacted. -- It seems that the dialog only notices the mouse, when the cursor is in the ContentPane. The RootPane then has bounds (0,0,0,0), so that could be the reason... but when I noticed that, I had already deleted the MouseListener and the dilaog resizing actually works the same way without an added mouse-up so, it was not relevant after all....

peteihis:
Almost there. :)

- Translations done (English and Finnish)
- `SoftReferences` for previews
- Metadata added to all image types
- A previous bug caught and fixed (related to missing image files)
- Purge UI developed further - -NOw woth a 'ScrollPane' and improved graphic appearance.

What is missing is, that `Convert to local` should have an "are you sure"-check and few more minor things that are bit by bit coming to my mind ... But if nothing big is found broken, I think this would be ready to merge. Some patching of little flaws could be done after.

peteihis:
This time:
- Corrected a logical glitch in filling the `ImagesDialog` with 20 or 21 icons + keeping the selected one in view while resizing.
- Image name no longer drawn over the edge of the icon _(I bet, nobody would have noticed any time soon... ;) )_
- Some small code fixes, not visible to user.

Now all I can think of is a bit smarter "emergency image" for broken links and some work on the `missing_image`-icon.

peteihis:
I noticed, that I had forgotten one tiny thing about the saved metadata for External Images. Nothing big, but that would help with some further work. -- Also found a case when `Refresh` is not working right. The kind of a strange bug, that may go unnoticed and, when it happens, there seems to be no why.... I'll try to get those fixed ASAP.

ArtOfIllusion/src/artofillusion/image/HDRImage.java
```diff
@@ -178,27 +183,20 @@ private void findAverage()
...
   /** Construct the preview image. */
 
-  private void createPreview()
+  private void createPreview(int size)
   {
     int w, h;
-
-    if (width[0] <= PREVIEW_WIDTH && height[0] <= PREVIEW_HEIGHT)
+    System.out.println(height.length);
+    System.out.println(width.length);
```
peastman: Looks like some debugging code got left in.

ArtOfIllusion/src/artofillusion/image/MIPMappedImage.java
```diff
@@ -542,15 +566,126 @@ private void getMapGradient(Vec2 grad, int component, int which, boolean wrapx,
...
+
+  /** Get a scaled down copy of the image, to use for previews.  The dimensions of the 
+      Image will be no larger but may be smaller than size. */
+
+  @Override
+  public Image getPreview(int size)
+  {
+    if (size == lastPreviewSize && preview.get() != null)
+	{
+      return preview.get();
```
peastman: There's a tiny but nonzero chance the soft reference could get cleared between when you check if it's null and when you access it again.  It's safer to write

```Java
Image image = preview.get();
if (image != null)
  return image;
```

ArtOfIllusion/src/artofillusion/image/SVGImage.java
```diff
@@ -373,35 +399,100 @@ public void getGradient(Vec2 grad, int component, boolean wrapx, boolean wrapy,
...
   {
-    return preview;
+    return getPreview(PREVIEW_SIZE_DEFAULT);
+  }
+
+  /** Get a scaled down copy of the image, to use for previews.  The dimensions of the 
+      Image will be no larger but may be smaller than size. */
+
+  @Override
+  public Image getPreview(int size)
```
peastman: Here again you're accessing the soft reference in a dangerous way.  Everywhere you have `return preview.get()` you should make sure it isn't null.

ArtOfIllusion/src/artofillusion/image/ExternalImage.java
```diff
@@ -0,0 +1,363 @@
...
+    setDataCreated(file);
+    lastAbsolutePath = file.getAbsolutePath();
+    // lastRelativePath can not be created without the path to the scene.
+  }
+
+  private Image loadIcon(String iconName)
+  {  
+    try 
+    {
+      return ImageIO.read(ExternalImage.class.getResource("/artofillusion/image/icons/" + iconName));
```
peastman: Instead of doing this, it's better to load icons through `ThemeManager`.  That way they can be customized by themes.

peastman:
Looks nice!  I added a few comments about minor points.  One other minor thing: there are a lot of tab characters scattered through the files, which messes up the formatting.  It would be best to do a find a replace to replace all of them with spaces.

Sailsman63:
I haven't gotten to run this up and test it yet, but in addition to what @peastman stated, I notice that you seem to have gotten some housekeeping commits in the history of this branch. If you start your feature branch from a direct copy of the project master branch, you should never have to merge master into the feature branch.

I'll be able to clean that up, before merging when this branch is ready to go.

PR #35, opened 2017-05-27
External Image and image management tools

peteihis:
Oops.... 

I ended up starting the file pushes over with a clean built repository. Lost the branch including the comments section but I'll reply the comments that are still relevant along with pushing new updates.

Sorry for the mess ... or should I say "Sorry for the clean up"? :)

**EDIT:** 
The earlier discussion: https://github.com/ArtOfIllusion/ArtOfIllusion/pull/34#issuecomment-304421454


peteihis:
@peastman

> Looks nice! 

Thank you! :)

> One other minor thing: there are a lot of tab characters scattered through the files, which messes up the formatting. 

I have been trying to remember to clean those off, but It seems that I forgot something. Notepad++ fills the beginnings of lines with tabs, when ever possible.

```
System.out.println(height.length);
System.out.println(width.length);
```

> Looks like some debugging code got left in.

Only two? ;)

peteihis:
@peastman
```
if (size == lastPreviewSize && preview.get() != null)
{
  return preview.get();
```
> There's a tiny but nonzero chance the soft reference could get cleared between when you check if it's null and when you access it again. It's safer to write

```
Image image = preview.get();
if (image != null)
  return image;
```
Ì was thinking about this a bit: Here `image` could still be null and then he method would have to return something else. On teh other hand, if I write the method(s) like

```
  if (previewSize == size && preview.get() != null)
	return preview.get();
  createPreview(size);
  return preview.get();
```
into 

```
  if (preview == null) // This needs to be added. The first time the SoftReference does not yet exist.
    createPreview(size);
  Image image = preview.get();
  if (previewSize == size && image != null)
    return image;
  while (previewSize != size || image == null)
  {
    createPreview(size);
    image = preview.get();
  }
  return image;
```

but then, in principle, there is a possible infinity loop unless there is a counter or a timer to break it and with the while-loop or not, there should be a alternative value to be returned if the `.get()` returns `null. 

Wouldn't it be clearer just acknowledge, that the method may sometimes return `null` and deal with the occasional null in the receiving end. -- Like `if (image == null) return;` ? That's what I would suggest anyway. Currently the texture editors of AoI, askin for preview images, are probably not built that way, but I wouldn't think the change to be too big a job _(though obviosly something to be handled as a project of it's own)_.

peteihis:
@peastman

```
try 
{
  return ImageIO.read(ExternalImage.class.getResource("/artofillusion/image/icons/" + iconName));
```
> Instead of doing this, it's better to load icons through ThemeManager. That way they can be customized by themes.

I agree, but doesn't that require changes to the ThemeManager? To keep the proportions of this project manageable, I think, I'll leave that idea out of scope this time and be back at it at some time later. -- For now I'll just make the icons better _'fit for all audiences'_.

peteihis:
@Sailsman63
> I notice that you seem to have gotten some housekeeping commits in the history of this branch. 

Sorry. 

The housekeeping stuff stopped bothering me at some point, but I didn't realize, that it would force itself into the later commits. I had probably cloned the wrong URL in the first place(?) and anyway it seems almost imposible to get things back on track after they start to fail.

Now, I don't thing there is anything left there that shouldn't be. -- And I'll not try to make things too handy it this end again  (....but I had to try...)  :D

peteihis:
I added the _image type_ and a setting if the image name is _automatic or user set_ into the saving format of `ExternalImage`. Those help with handling the missing files and (obviously) image name. **Now, what was saved with the version of previous commits can not be opened with this one any more.**

In the dialogs:
- You can move in the preview grid with arrow keys. The `zoomToSelection` does not comply with this very nicely though. Enter and Esc are now OK and Cancel.
- On `ImagesDialog` `cancel` only matters if the dialog has been activated by a texture editor. All changes to images will stay in effect after cancel, but the selection is reverted to the original one. -- I wonder if the cancel should be removed as a button? I was trying to make the button appear only, when it has a meaning, but it did not work out because the image mapped textures editor (which it would help best) sends `LayoutWindow` as a 'parent' to the dialog instead of itself.
- On `ImageDetailsView` the image name can be edited. For external images you can select, if the name is gengerated from the file name or if it is user set.

peastman:
> Notepad++ fills the beginnings of lines with tabs, when ever possible.

It probably has an option for whether to use tabs or spaces.  Take a look in the preferences.

Here's the safe way to handle the soft reference: have `createPreview()` return the Image object directly.  That way you know you have a real object to return.  All responsibility for the soft reference can be handled inside `getPreview()`.  It will look something like this:

```Java
Image image = preview.get();
if (image != null)
  return image;
image = createPreview(size);
preview = new SoftReference(image);
return image;
```

> I agree, but doesn't that require changes to the ThemeManager?

What changes?  Making images customizable is one of the main things it does.

peteihis:
I was about to push this update almost a week ago, but I had a persistant bug there, that I did not quite catch after all, but at least I think I managed to prevent it occurring... _(Related to openig a scene with images missing and refreshig with the images still missing... It did not break anything, but the UI stopped working and the error was unrecoverable during that session)_

I changed the way, that missing image files are handled. Now the thumblails tells, what kind of an image there used to be and the thumbnail image is used to geneate a replacement image, that is used while the reference is empty. -- The notification images are still WIP. I'll update those, when I find he time. 

I also just noticed, that the name of an external image can not be edited while the link is broken. That is not an intentional feature.

@peastman
> Here's the safe way to handle the soft reference:...

Thanks. I see that I was cutting corners with it a bit.

> What changes?

Oops. I was probably recalling, when I was looking at the possibility of adding theme dependent colors. :)

Anyway, I tested so much, that `ThemeManager` was able to get the images into the software and there were no complaints about any errors, but somehow the icons did not appear on the canvas of `ImagesDialog`. The difference to start with is, that the information icons are currently read and handled as `Image` whereas the `ThemeManager` returns `ImageIcon`. I did not go tracking, what happened after `ImageIcon.getImage()` or if that even was the right thing to do...

**Another issue yet:** Now, when the images change, the views don't change the texturing of objects right away. You have to go editing the textures to get them updated. This does not affect rendering the scene, but it is a bit of an annoyance. I don't think it is directly related to the image management but rather to some other minor issue with textures: Texture previews are not alwas up to date.... The preview is usuall updated if you open and close the mapping editor. This is a place for a different study though.

peteihis:
Is there are'n any serius problems with this one, I consider this PR ready.

There are things that could still be done:
- Icons under theme control
- Template and reference images under the same management with the rest of the images

IMO those should go into their own branch(es) and pull request(s). I wouldn't like to keep this one waiting for those.

>  the name of an external image can not be edited while the link is broken

Now it can. ...trying and catching.. :)

peteihis:
When you hear a perfectionist say "It's final", start to prepare to the next change... 

One minor thing I forgot in the code and some tinkering with a couple of the icons. Now, I promise: No more changes!

Sailsman63:
Thanks for posting. I think that this feature will be really great. My next step is to take a more detailed look through the actual code. I may have some feedback for you.

I already feel like the interface could use some polish. All of those buttons in the image manager just make me think that there has to be a better way. That can probably wait for some feedback from our other users, though.

peteihis:
> All of those buttons in the image manager just make me think that there has to be a better way.

I a kind of agree. I tried a few different things and so far the button group is, what I was least unhappy with. 

One thing you might ask is, if the Cancel button is needed? It only has a meaning, when the dialog is launched by a texture editor and texture editors have their own cancel buttons. -- I'd keep the function though for closing the window by "the red X" and Esc-key. 

I actually tried to make the Cancel button appear only when relevant, but that was not possible with the current ImageMapped textures editor _(which sends the LayoutWindow as the parent for the dialog, unlike the others)_ and I did not want to escallate the project any more than I had to.

peteihis:
How about this:

- Keep the `ImageMap` as the super type, that textures communicate with.
- Have two direct sub types: `ExternalImage` and `MappedImage` (or such).
- The actual maped image types (HDR, SVG, MIP...) would be sub types of `MappedImage`.
- `ExternalImage` would wrap a `MappedImage`, _not_ another `ImageMap`.

That would make a clear separation between the classes and get rid of any overlapping and it seems a relatively simple task to do.

Earlier Peter said something about the ExternalImage _implementing_ ImageMap but after a short test I went back to extending (don't remember the details, but it seemed simpler that way). On general level I was trying follow the example of `ObjectWrapper`, where both the wrappers and the objects are extending Object3D. But...

Currently ImageMap has (at least) the method, that chooses which image type is going to me mapped, that ExternalImage does not use.t. On the other hand, if `ExternalImage` were to be merged with `ImageMap`, then all the functions, that are needed to handle the external link _and what ever else comes with that_, would be unnecessary to the internally mapped images and would actully need to be turned off or change their function in that case.

How does that sound to you?

ArtOfIllusion/src/artofillusion/image/ExternalImage.java
```diff
@@ -0,0 +1,479 @@
...
+import java.awt.image.*;
+import java.io.*;
+import java.util.Date;
+import javax.imageio.*;
+
+public class ExternalImage extends ImageMap
+{
+  private File externalFile;
+  private String loadingError, imageType;
+  private ImageMap imageMap;
```
peteihis: How about this:

- Keep the `ImageMap` as the super type, that textures communicate with.
- Have two direct sub types: `ExternalImage` and `MappedImage` (or such).
- The actual maped image types (HDR, SVG, MIP...) would be sub types of `MappedImage`.
- `ExternalImage` would wrap a `MappedImage`, _not_ another `ImageMap`.

That would make a clear separation between the classes and get rid of any overlapping and it seems a relatively simple task to do.

Earlier Peter said something about the ExternalImage _implementing_ ImageMap but after a short test I went back to extending (don't remember the details, but it seemed simpler that way). On general level I was trying follow the example of `ObjectWrapper`, where both the wrappers and the objects are extending Object3D. But...

Currently ImageMap has (at least) the method, that chooses which image type is going to me mapped, that ExternalImage does not use.t. On the other hand, if `ExternalImage` were to be merged with `ImageMap`, then all the functions, that are needed to handle the external link _and what ever else comes with that_, would be unnecessary to the internally mapped images and would actully need to be turned off or change their function in that case.

How does that sound to you?

peastman: I don't see a problem with the current implementation.  ExternalImage is an ImageMap that delegates many of its functions to another ImageMap.  That seems simple enough.  It's performing a specialized function: creating a link between two different files doesn't belong in the parent ImageMap class.  But the image loaded from the external scene could be of many types, so it also needs a reference to that other object.  This is exactly the same way ExternalObject works.

Sailsman63: I see your point. Still not my favorite design, but I do understand the logic. I'll try to get this merged in the next couple of days. I want to get some of the commented-out code snips removed for real.

peteihis: > I want to get some of the commented-out code snips removed for real.

I think that should be my job... I forkot to clean those off in the end. I can look at those tonight, it should not be a very big thing to do.

peteihis:
There's one thing I meant to ask about the `writeToSteam()`.

Now, the method takes just the output stream as an argument for the imageMaps that have been existing before, but for `EsternalImage` there is a specal call that also takes the scene as an argument. (`Scene.java` 1568 - 1573).

It might be clearer if the method would be the same one for all of them, though only `ExternalImage` uses the scene. This current arrangement is more or less the outcome of me trying not to make too may changes to the existing practices early in the project, but it does not quite seem right to me after all.

peteihis:
>>I want to get some of the commented-out code snips removed for real.

>I think that should be my job... I forgot to clean those off in the end. I can look at those tonight, it should not be a very big thing to do.

Done that & a couple of minor fixes. Did not change the `writeToStream()`s at this time.

Sailsman63:
Thanks! I think that this one is ready for user testing.

Sorry for the delay in getting it merged -  my life went crazy here.

PR #37, opened 2017-06-05
Adding unit() to Vec3.

peteihis:
Adding this method, that I have been missing.  Discussed on SF https://sourceforge.net/p/aoi/discussion/47784/thread/dd7c3020/

### DIFF
--- a/ArtOfIllusion/src/artofillusion/math/Vec3.java
+++ b/ArtOfIllusion/src/artofillusion/math/Vec3.java
@@ -179,7 +179,7 @@ public class Vec3
       }
   }
   
-  /** Crate a copy of this vector, scaled to the length of 1. If the length of this vector is 0, 
+  /** Create a copy of this vector, scaled to the length of 1. If the length of this vector is 0, 
       a new vector with length 0 will be returned. <p>
       
       This is like <pre>normalize()</pre>, but returns a new vector.*/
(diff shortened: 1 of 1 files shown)


ArtOfIllusion/src/artofillusion/math/Vec3.java
```diff
@@ -178,6 +178,20 @@ public final void normalize()
         z /= len;
       }
   }
+  
+  /** Crate a copy of this vector, scaled to the length of 1. If the length of this vector is 0,
```
peastman: Crate a copy?  Like, put it in a crate? :)

peteihis: I thought it might be handy.... :)

This thing is genetic, and the usual "be more careful" and "read it once more over" don't do it for me. It is not uncommont, that when I write a few lines of code, it takes more time of Ant to get through the spell checking phase that I took me to write it. -- So, what can I but to add one more of the usual comments: "Be patient...." :)

Sailsman63: Or perhaps you are starting to learn the Rust language? They put their code in crates... :D

peteihis: Hmm... No racing, no garbage, stacks, boxes, crates, cargo... Everything will turn Rust eventually? :O

Sailsman63: Well, many Rust fans want you to think that. They may well eventually replace chunks of the system libraries that everyone depends upon, but there will probably still be a place for languages like Java. Future JVMs, though, might be written at least partly in Rust....

peteihis: Cleverly picked name at least...

peteihis:
@Sailsman63: On my behalf this bit is ready.

PR #38, opened 2017-06-07
Dynamic material, texture and mappings probe

makiam:
Made getTypeName method non static for Materials and Textures
Made getName method non static for Material and Texture mapping classes
Create base legalMapping implementation and needed overrides for Mappings classes

This allows to replace calls to this methods through reflection with direct calls.


ArtOfIllusion/src/artofillusion/material/LinearMaterialMapping.java
```diff
@@ -45,7 +45,8 @@ public String getName()
     return "Linear";
   }
 
-  public static boolean legalMapping(Object3D obj, Material mat)
+  @Override
+  public boolean legalMapping(Object3D obj, Material mat)
```
Sailsman63: I sort of get where you are going with this, but these are an API-breaking change. Any plugins or scripts that call `XYZMapping.legalMapping()` would be broken if this change was released. We're not really in a position to fix that breakage quire yet, so I want to approach this more carefully.

We also need to make a larger design decision here: Are we going to keep our list of mappings as `Class` objects, or as instances of the specific mapping types? (See also #24, and the differences in how parts of that list are generated) In general, I think that moving away from reflection, if there are clean ways of doing without, would be preferable.

ArtOfIllusion/src/artofillusion/ObjectTextureDialog.java
```diff
@@ -145,15 +144,7 @@ public ObjectTextureDialog(LayoutWindow parent, ObjectInfo objects[], boolean in
     
     for (Texture texture : PluginRegistry.getPlugins(Texture.class))
     {
-      try
```
Sailsman63: Noticed that you have removed a lot of try blocks in this PR. Some of them certainly look as if they are not guarding anything, but are you _sure_ of that? In particular, what happens if you open a file that contains a texture that relies on a plugin, and you do not have the plugin?

@peastman would it even be safe to edit and save the file in such a case? I looked at the scene loading code, and do not see where unreadable texture data would be retained?

ArtOfIllusion/src/artofillusion/MaterialMappingDialog.java
```diff
@@ -1,4 +1,5 @@
 /* Copyright (C) 2000-2004 by Peter Eastman
+   Changes copyright (C) 2017 by Maksim Khramov
```
Sailsman63: Please do this in the commit where you actually edited the file. If you forget, `git commit --amend` and `git rebase -i` are your friends. In this case, please leave things as they are until we can complete reviewing your changes. If we decide to merge the PR, I'll clean up the commit history on my end.

ArtOfIllusion/src/artofillusion/material/LinearMaterialMapping.java
makiam: Looks like we need a development plan

ArtOfIllusion/src/artofillusion/ObjectTextureDialog.java
makiam: I see only reflection-dependent exceptions can be thrown there. But there are no more such code.

ArtOfIllusion/src/artofillusion/material/LinearMaterialMapping.java
makiam: Do we actually have such plugins/scripts which implements custom Mapping/Material/Texture?

ArtOfIllusion/src/artofillusion/ObjectTextureDialog.java
peastman: I think those try blocks are just there because reflection requires them.

ArtOfIllusion/src/artofillusion/material/LinearMaterialMapping.java
peastman: I agree, we shouldn't make incompatible changes unless we have a really good reason to.  What's the reason in this case?

ArtOfIllusion/src/artofillusion/ObjectTextureDialog.java
Sailsman63: If we eliminate the use of reflection, then they would not be necessary.  I'll have to take a little more time to consider the other comments on this PR.

ArtOfIllusion/src/artofillusion/material/LinearMaterialMapping.java
Sailsman63: I _think_ that this is part of trying to get out of using reflection. Instead of `instance.getClass().getMethod("legalMapping").invoke(obj, mat)` we would call `instance.legalMapping(obj, mat)` directly.  For this to work, the method must not be static, as static methods get called based in the _reference type_ (ie, we would always get what is defined in `MaterialMapping`, even if what we really have is a `LinearMaterialMapping`)

Regarding use in Plugins, It does not appear that any of the current major plugins create custom mappings. Polymesh has a class that __should__ be, but is defined outside the inheritance tree...

I have a thought in mind that _would_ want to use this as an extension point, but it does not exist yet.

All in all, I do think that this way is cleaner. I'm not sure why reflection was being used here in the first place. Even the `PluginRegistry` actual object instances to work from.

peastman: But is there any benefit to it other than simplifying a very small amount of code (a few dozen lines total)?  That doesn't seem like a good reason to change a public API in a way that breaks compatibility.

Sailsman63: In this case, I think that there may be, mostly because this _is_ intended as a public API, a designed extension point. It was always intended that plugins would add other mapping types, and I have at least one in the back of my mind that would do so.

I'd expect an intentional API to try to be as natural as possible. (What some call the 'principle of least surprise')

Think of it from the viewpoint of someone developing a plugin or script. The natural thing to do, if you don't realize that `legalMapping()` is overridden by every subclass, (or don't realize all of the implications) is to call it in `instance.legalMapping()` style. We could argue about whether Java should allow that to happen, but it _does_ allow it. The developer would probably get away with it for a while, especially with a Material mapping. As soon as someone adds a new type of mapping, however, things will break in unexpected, probably hard to de-bug ways.

I'm not saying that we need to change this tomorrow, but we should consider it while we still have a plugin ecosystem that will mostly not even notice.

What I would like to do is start a list of breaking API changes/improvements that would be nice to have. We can add to them as we continue reviewing the code and cleaning up lint points, etc. Once we have a full list, we could make all of the changes in one version increment, so that we have a well-defined API version change.

PR #39, opened 2017-06-23
Division by zero-bug fixed

peteihis:
This was the rootcause of camera views sometimes going blank.

To repeat the bug on on 3.0.3:
- Make sure you are using the `SoftwareCanvasDrawer`
- Open a new scene. Don't move or rotate views.
- Draw something on the scene for reference (and select it).
- Switch to 3D-manipulator.
- Select Camera 1 on the item list.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/view/SoftwareCanvasDrawer.java
+++ b/ArtOfIllusion/src/artofillusion/view/SoftwareCanvasDrawer.java
@@ -168,7 +168,7 @@ public class SoftwareCanvasDrawer implements CanvasDrawer
       if (drawFocus){
         //pixel[i+bounds.width] = pixel[index1+i-bounds.width] = line;
         pixel[i+bounds.width] = pixel[index1+i-bounds.width] = Color.GRAY.getRGB();
-	  }
+      }
     }
   }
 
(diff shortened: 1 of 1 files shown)


Sailsman63:
Good detective work. I think that your fix is probably not quite right, there may be a few more edge cases in there. I'll have to spend time on that later... Good to know that there is a fairly simple explanation for it though.

peteihis:
Take your time :)

I checked so much, that I had the ViewerCavas draw a vertical and a horizontal line for no other reason but to test this fix. Those at least came through correctly.

Having slept one night over it, it also came to my mind that the conditions like

~~~~java
if (Math.abs(dx) > Math.abs(dy) && dx != 0)
~~~~

would be better with the simpler check first:

~~~~java
if (dx != 0 && Math.abs(dx) > Math.abs(dy))
~~~~

After that I'd remove the check if both of the variables are zero.

peteihis:
Ok ... Went a little deeper.

The error originally happens in the `Compont3Dmanipulator`, where in this case `screenY.x = NaN` and `screenY.y = -Infinity`. When these are converted to `int`, they become `0` and `-2147483648`.

Now, as it happens, the largest possible POSITIVE integer is only `214748367` so the `Math.abs` of the negative number causes overflow by `1`, returning again `-2147483648`. That's how the `dx = 0` got through the test with the `Math.abs`es. -- Not that it mattered much as then a similar case would be left with the `dy`.

So, it probably does not hurt to vaccinate `SWCanvasDrawer` against the aritmetic error, but the actual fix should happen in the manipulator. I have not yet checked the code, but the error must be in the 3D to 2D conversion. A fix coming to my mind woud be to just change any NaN-Infinity-pairs that come out of the matrix equations, into 0's.

peteihis:
You were right about those edge-cases. With the zero value cases ruled out, some ghost lines started to appear on the screen and some lines were not drawn at all or at least not correctly. 

This one now takes care of the actual numercal problem in SWCD. 

In the 3D-Manipulatior the `worldToScreen` matrix is used a lot (for both ends of every tiny element of the drawn manipulator). The negative infinity values appear in the perspective viewing, when the camera z-distance to a 3D-point is 0. It is virtually impossible to catch all the cases, when those might occur. Also who can tell tha there would not be other sources for `Integer.MIN_VALUE`?

Another thing is that currently the manipulator is drawn in perspective mode even, when it is behind the camera ... I'll add a check for those cases at least. :)

peteihis:
Realized, that the exact same structure is used in `renderLine()`. I'm not sure if there is surrently anything could put `Integer.MIN_VALUE` in there, but I don't think, it can be guaranteed that it would never happen. :)

Sailsman63:
Thanks for the fix! That one had been happening _forever_ and could be really annoying.

PR #40, opened 2017-06-28
Unit tests for Scene class

makiam:


Tests/src/artofillusion/SceneTest.java
```diff
@@ -74,16 +74,15 @@ public void testCreateScene()
...
       int sceneObjects = scene.getNumObjects();
 
       scene.addObject(new Cube(1d,1d,1d), new CoordinateSystem(), "Cube", (UndoRecord)null);
       ObjectInfo so = scene.getObject("Cube");
       
       assertEquals(++sceneObjects, scene.getNumObjects());
-      
-      assertNotNull(so);
+
       assertEquals("Cube", so.getName());
```
Sailsman63: Thanks for your update. I think you missed my point about redundant asserts, though. I'm specifically referring to lines like:

```
assertEquals("Cube", so.getName());
```

You just created this cube object, so there should be no need to check on its name.

Sailsman63 (dismissed):

Tests/src/artofillusion/SceneTest.java
```diff
@@ -10,32 +10,768 @@
...
+    }
+    
+  /**
+   * Test to remove image used through material from scene
+   * Checks that image deletion is rejected
+   * As no way to add image to Procedural Material as Image module programmatically test fails
+   * @throws InterruptedException
+   */
+  @Test
+    public void testAttemptRemoveImageUsedInMaterial() throws InterruptedException
```
Sailsman63: Generally, you won't use images in a `Material` at all. Not sure that this one is testable, or even makes sense.

Try adding using an `ImageModule,` perhaps in conjuction with a `ProceeduralTexture2d`. You're really trying to test if the image removal fails _when using the proceedural system_ anyway.

Tests/src/artofillusion/SceneTest.java
```diff
@@ -10,32 +10,768 @@
...
+      assertEquals(1, listenerFireCount);
+    }
+    
+  /**
+   * Add material to scene at given position.
+   * Check that material listener event is triggered
+   * Check that material is inserted at expected position
+   */
+  @Test
+    public void testAddMaterialAtGivenPos()
```
Sailsman63: This method is failing with the current code. Please check your logic, and sanity-check by adding a material through the script console, and then retrieving and printing the material list.

Tests/src/artofillusion/SceneTest.java
```diff
@@ -10,32 +10,768 @@
...
+      assertEquals(2, scene.getNumTextures());
+    }
+    
+  /**
+   * Add texture to scene at given position.
+   * Check that texture listener event is triggered
+   * Check that texture is inserted at expected position
+   */
+  @Test
+    public void testAddTextureAtGivenPos()
```
Sailsman63: And this one as well. It seems to have the same logic as the above, and shows the same failure.

Tests/src/artofillusion/SceneTest.java
makiam: Both testAddMaterialAtGivenPos and testAddTextureAtGivenPos test cases fails because of code contract violation as discussed in this [thread](https://sourceforge.net/p/aoi/discussion/47784/thread/7aa9f80e/)

Tests/src/artofillusion/SceneTest.java
makiam: Idea is to test that image used in Material prevent to be removed from scene, similar to image used in Texture which tested with testAttemptRemoveImageUsedInTexture(...). (This checks Scene's removeImage(...) method). 
But as no way to add image to Material programmatically I stub this test with fail(...)

Sailsman63: But you have not tested the general case of "Image used in `Texture`." You have tested the specific case of "Image used in `ImageMappedTexture`."

That is a perfectly good thing to test. The other thing to test is the case of "Image used in `Procedure`." This would have to be done with some extra work, perhaps by generating a special Object3d class in the test case that implements `ProcedureOwner`.  This would cover both the `ProceduralTexture` and `ProceduralMaterial` cases, as its the behavior of the `ImageModule` itself that would be of concern.

Tests/src/artofillusion/SceneTest.java
Sailsman63: Thats right. I'd forgotten.

PR #41, opened 2017-07-08
Adding unit() to Vec2 too.

peteihis:
Working on the 3DManipulator I realized that I had forgotten Vec2, when I added unit() to Vec3, so here it is. :)


Sailsman63:
Thanks. Merged.

PR #42, opened 2017-07-09
Tests for ObjectInfo class

makiam:


Tests/src/artofillusion/object/ObjectInfoTest.java
```diff
@@ -0,0 +1,852 @@
...
+    assertTrue(target.getDistortion() instanceof DistortionImpl);
+    
+  }
+  
+  /**
+   * Test objectInfo copy data from other objectInfo and points to same geometry
+   * Checks that empty distortion data overwrites existed one
+   */ 
+  @Test
+  public void testCopyInfoWithNullDistortionOverExisted()
```
Sailsman63: This test fails with the current codebase. There is an explicit check to make sure that the distortion is _not_ null before copying it over existing data. See:

https://github.com/ArtOfIllusion/ArtOfIllusion/blob/7623fc4a7d3d58434c90b294ad499479f210d55c/ArtOfIllusion/src/artofillusion/object/ObjectInfo.java#L136-L139

@peastman do you know what the reason is here for _not_ overwriting with a null?

peastman: No reason I can think of.  I suspect I just forgot to handle the case where it's null.

Sailsman63: Makes sense to me.

Sailsman63:
Thanks for the tests and quick fix!

PR #43, opened 2017-07-15
Redundant calls to scene updateSelectionInfo() clean.

makiam:
Redundant calls to scene updateSelectionInfo() clean.
Simplify empty scene creation code
Added not undoable addObject(...) method


ArtOfIllusion/src/artofillusion/Scene.java
```diff
@@ -455,27 +454,36 @@ public void addObject(Object3D obj, CoordinateSystem coords, String name, UndoRe
   public void addObject(ObjectInfo info, UndoRecord undo)
   {
     addObject(info, objects.size(), undo);
-    updateSelectionInfo();
   }
 
+  /** 
+   * Add a new object to the scene.  No undo record is added
+   */
+  public void addObject(ObjectInfo info)
```
Sailsman63: I'm not sure that we want another method here. Adding an object without an `UndoRecord` should be rare, and the odd thing to do.

ArtOfIllusion/src/artofillusion/Scene.java
```diff
@@ -455,27 +454,36 @@ public void addObject(Object3D obj, CoordinateSystem coords, String name, UndoRe
...
   public void addObject(ObjectInfo info, int index, UndoRecord undo)
   {
     info.setId(nextID++);
     if (info.getTracks() == null)
-      {
-        info.addTrack(new PositionTrack(info), 0);
-        info.addTrack(new RotationTrack(info), 1);
-      }
+    {
+      info.addTrack(new PositionTrack(info), 0);
```
Sailsman63: There may be some cases where an object should not have any default tracks, such as a dummy object that is holding data for a plugin.

ArtOfIllusion/src/artofillusion/Scene.java
makiam: Fixed

ArtOfIllusion/src/artofillusion/Scene.java
makiam: This addObject method is only way to add object to Scene as objects field is private access only. And this method is always checks and and adds 2 tracks to object. In my mind adding this tracks to every obejct is not necessary at all

Sailsman63: Sorry, I thought that you had added these lines. I see now that they have been in the codebase for a very long time.

PR #47, opened 2017-07-26
Remove dependency of Windows System and Scene model in Preferences

makiam:


peteihis:
> "...dependency of Windows System.."

How so? I don't see any depencency on Windows system. The list of open `EditingWindows` was named "`windows`".

makiam:
I mean dependency of EditingWindow

Sailsman63:
What is the intent with the change listener structure? How would you see this used outside of the application preferences class?

makiam:
I see two pro's for this.

- The properties system is no more depends of aoi EditingWindow and Scene classes
- Every aoi code which depends of app properties can listen for the it's changes

peteihis:
My question would be, what is the effect on practical level?

If I get this right, currently the only connection to `EditingWindow` is, that the `interactiveSurfaceError` is made effective immediately after changing it. If I understand your idea right, you would change property changes event driven? But doesn't that mean, that then every element, that has any user changeable properties should have an event link to capture preferences changes ... which currently is not the case?

I have been thinking, that there really should not be a reason, why AoI, in most cases, needs to be restarted after user preferences have been changed. From the user's point of view it would be an improvement if the changes could happen immediately and basically using events sounds like the way to do it but, on general level, I think we'd need a bigger plan how to have all of the code comply at the same time. 

_A bit out of topic, currently (I mean the current Git HEAD) the background color change happens immediately. That's because of, how I built the `ViewAnimation`. Unfortunately there is no mechanism to have the rest of the UI colors adapt. I think I should, for the time being, consider trying to change that back to "restart mode"._

makiam:
For now this change has no immediate parctical effect. In the future applying same style to Locale propety will allow us to switch UI language immediately without restart.

peastman:
> I have been thinking, that there really should not be a reason, why AoI, in most cases, needs to be restarted after user preferences have been changed.

In most cases it doesn't need to be.  There are a few preferences that only affect newly opened windows, like choosing whether to use OpenGL.  But most of them should take effect immediately.

peteihis:
> In most cases it doesn't need to be.

As funny as it might be .... **

I had a look at that color issue, I commented above and now I know how to have the background-dependent colors adapt.

_**(it continues, "..the one and only man in town was me.")_

Sailsman63:
I agree with Pete. A change like this (Adding an eventing system) has been discussed off and on, but we need to have a plan.

- What things should fire an event?
- What event channels should we have?

I'm closing this for the time being. Please feel free to continue discussing, either here or on Sourceforge. Please also keep in mind that changing the name of a preference property will cause annoyance for some users unless you also check under the old property name.

peteihis:
One other thing: There is still something odd going on with `LayoutWIndow` as again this time GitHub sees an about 1300 line chunk being changed, though the changes are probably just a few lines. 

I wonder if there could be some hidden characters somewhere, that the text editors ignore but the GitHub comparison catches? Should we run the file though a machine, that performs some kind of a byte level check on it?

Sailsman63:
Somehow, we still have some windows & mac line endings in the repository. If a text editor then saves with linux style endings, the repo shows a huge block change. That's one of my projects for *soon*, but I want to clean up some of the existing PRs first.

peteihis:
I ran a small check. See my findings _(and the sorry attempt to fix things)_ in PR #52.

PR #48, opened 2017-08-05
Catch exceptions for missing properties

peastman:
#30 changed this from `Exception` to `NumberFormatException`, which meant that if any other type of exception was thrown, it wouldn't be caught.  Most importantly, if a preference simply isn't present (because you've just upgraded from an older version), it throws a `NullPointerException`.  This is a case where it's safest to just catch all exceptions, since returning the default value is always the best we can do in that case.


Sailsman63:
Thanks for catching this. I remember seeing that at the time, but it kind of slipped my mind due to other concerns with that PR.

PR #51, opened 2017-08-23
Standalone tests config

makiam:
New ant project to run junit tests


Sailsman63:
I don't think that we want to include the JUnit libraries in the repository.

I'm intending to re-arrange the tests a bit, and add test targets to the build scripts. #45 is a good start on this. (I'm also considering refactoring the scripts a bit, so hang in there!)

In the future,  if you are considering adding libraries or substantial changes to architecture, **Please** discuss it with the team, either on SF, or opening an issue here, before putting in the effort to make the code changes.

makiam:
I want also to add some tests for LayoutWIndow UI, but these need to use jemmy library to compile and run

Sailsman63:
Jemmy? Link please?

makiam:
On my own I take this lib from my Netbeans distributive but [this](http://bits.netbeans.org/nexus/content/groups/netbeans/org/netbeans/external/jemmy-2.3.1.1/RELEASE82/jemmy-2.3.1.1-RELEASE82.jar) is the same

Sailsman63:
I meant a link to documentation...

makiam:
http://wiki.netbeans.org/Jemmy

PR #52, opened 2017-08-23
LayoutWindow clenaup

peteihis:
Don't know if this helped. 

Anyway I wrote a script, that checks the file for certain characters. The findings were:

~~~~
Lines  3009
LF     3004
CR     1434
~~~~

Some times there were two CRs on the same line and somehow the amouts of LFs and lines don't match. The script actually checked a couple of other things too, but nothing else to report was found. 

I ran another script to clean the line ends. Now the check said:

~~~~
Lines  3009
LF     3009
CR     0
~~~~

This file is the cleaned one but, when I `git add -u`, git repots `LF will be replaced by LFCR`, so I don't know what happens now, but at least it looked good in this end of the wire.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -120,8 +120,8 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
       theView[i].addEventLink(MousePressedEvent.class, listen);
       theView[i].addEventLink(KeyPressedEvent.class, keyListener);
       theView[i].setPopupMenuManager(this);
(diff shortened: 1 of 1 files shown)


Sailsman63:
Thanks for making the effort. There are standard Linux command line utilities for things like this, (I settled on a combination of grep and sed) which will catch several other source files, and some of our XML files, etc.

The warning just means that the next time you check those files *out* on Windows, the working directory will have Windows style line endings. This is intentional, all windows checkouts already do this for 'correct' files. The problem is that currently the *repository* has some Windows line endings. Let me finish up reviewing the current PRs tonight, and I'll get this fixed for real.

peteihis:
Well, that was just an experiment out of curiosity anyway but did reveal some interesting things. (The fix script was not much of an effort on top of that.) :)

PR #56, opened 2017-09-06
Build refactor

Sailsman63:
A bit of an overhaul to the build scripts. Nothing earth-shattering, but should be a little more ergonomic to use.

- We now have a standard `build.xml` which is used by default by ant. No more `ant -f ArtOfIllusion.xml [target]` stuff. The individual sub-projects can still be called on their own by name, if neccessary.
- The working application (.jar files and script/plugin directory structure) is now built in its own subdirectory, `Live_Application/`by default, rather than being intermixed in with the source directories. The exact location is specified in a property that can be overridden at the command line.
- Default `docs` target now includes javadoc for all of the stock plugins, not just the core application classes. Default location for the combined javadoc is `docs/Javadoc/`, but this can be overridden.
- Tartgeted java version can be overridden on the command line. This should make doing experimental builds for newer JVMs much less painful.
- Build Script help: run `ant help` to see available targets, and the overridable properties.

There's more that could be done:
- Pass compiler flags at runtime, at least for `-Xlint` options.
- refactor some of the sub-build logic (most of the files are essentially identical, and that can be extracted into a file) - perhaps a little OCD.

Heading this way would make things like automated travisci builds much simpler to put together. Any comments or concerns before I merge this into master?


Sailsman63:
We also need to add an Ant target for the unit tests. This will require some refactoring in test file locations.

makiam:
Isn't a good idea to move other build scripts to corresponding project folders? So in every subproject we will have own build.xml ant file. Base build script this way just call something like
      <subant target="">
         <fileset dir="." includes="*/build.xml"/>
      </subant>

Sailsman63:
@peastman @peteihis Do either of you have any concerns about these changes, or things to suggest before I merge?

I'd also probably jump the default target version to java 1.6, as [this post](https://sourceforge.net/p/aoi/discussion/47784/thread/df034b7c/#8042) showed that we already do depend on Java 6 runtime classes.

peteihis:
I'm no XML-expert, but looks alright to me. I also second the 1.6, for one thing because of the improved `switch`.

peastman:
Moving to 1.6 is fine with me too.

peteihis:
Just curious, but what with the 1.666666 in OSSpecific?

Sailsman63:
Typo. Just fixed. __Don't__ run `git pull` on this branch, use `git fetch`, then `git reset --hard {latest commit hash}` (2e3e925)

peteihis:
~~~~
C:\Users\Petri\AoI_Branches\ArtOfIllusion>git fetch buildRefactor
fatal: 'buildRefactor' does not appear to be a git repository
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
~~~~

Sailsman63:
you need to add my personal AOI fork, `github.com/user/Sailsman63/ArtOfIllusion.git` as a third remote, then call `git fetch Sailsman63` or whatever tag you give me. Then you can do `git checkout -b {name for local copy of branch} Sailsman63/buildRefactor`

I forget that, since you do not have push access, the github UI does not give you short instructions on how to do this. I'm just about ready to merge these, though. A few more tweaks.

peteihis:
I can wait....  :)

build.xml
```diff
@@ -61,6 +61,32 @@
    </javadoc>
   </target>
 
+  <target name="test" description="Build and run the JUnit test suite.">
```
makiam: Why not to set this target depends on 'dist' target? This will reuse compile task from dist...

makiam:
BTW... Why not to set target jdk to JDK8?

build.xml
Sailsman63: Because we don't need the entire build. As our test suite grows, I intend to refactor the tests to be built on a per-project basis anyway.

Sailsman63:
@peastman has some concerns about depending on java 7 or 8 on AppleOS. We are certain that java6 works well on our supported platforms, and we don't yet have a clear use case for Java 8 features at this point.

makiam:
I try to use this new build system and I have an issues with it.
My scenarios:
1. Run ant on clean repository... This works OK. 
2. Try to run tests on clean repo:
run ant test
Target fails with lot of javac errors like: package does not exist or cannot find symbol. 
This occurs on first compile subant task - OSSpecific.

3.Clean repository and run ant again.
Run ant test again
task failed as no target folder Tests/build
Ok. create target folder manually and run ant test again
Test failed again because it cannon find junt classes...

Looks like on @Sailsman63 development host junit libraries are placed under ant/lib or defined vith CLASSPATH environment property

Sailsman63:
The `test` target does indeed depend on JUnit being installed correctly, and correctly referenced in your ant install. Please refer to [the ant user manual](https://ant.apache.org/manual/install.html#optionalTasks) on how to configure this if your OS does not have a package manager that takes care of it for you.

I see that the `test` target does not generate the Test/build directory correctly. I'll fix that in a bit.

makiam:
IMO is better solution is to embed junit into project (probably also other test libraries ex. jemmy mentioned #51, mocking libs, performance like jmh etc etc...). This also allows to use CI tools in the future...

Sailsman63:
Fixed the `test` target to properly generate the `Tests/build` directory. You'll need to do a fetch/reset update in this case as well.

> better solution is to embed junit into project

Absoulutely not! junit and similar are part of a developer's toolkit. They should be installed and configured locally. In fact, I'm hoping in the future to reduce the number of embedded libraries in hte source tree, and pull them in through something like IVY.

RE: CI tools; Any decent CI toolkit will allow you to download and install additional tools at the start of the build process. We will almost certainly be using TravisCI, which allows full `apt-get` access for the underlying OS.

> Looks like on @Sailsman63 development host junit libraries are placed under ant/lib or defined vith CLASSPATH environment property

Actually, they are listed in `{path_to_ant}/etc/ant.conf` which is not well documented on the apache site, but is the better way to specify such jars (Used by Linux package management systems like pacman)

If the file does not exist, you can create it, and add the line: `OPT_JAR_LIST={path_to_junit.jar}:{path_to_hamcrest-core.jar}`

PR #57, opened 2017-09-10
Use GLJPanel instead of GLCanvas

peastman:
Fixes layout problems using Oracle Java on Mac.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/view/GLCanvasDrawer.java
+++ b/ArtOfIllusion/src/artofillusion/view/GLCanvasDrawer.java
@@ -32,7 +32,7 @@ import buoy.event.*;
 public class GLCanvasDrawer implements CanvasDrawer
 {
   private ViewerCanvas view;
-  private GLCanvas canvas;
+  private GLJPanel canvas;
   private GL2 gl;
   private Rectangle bounds;
   private Mat4 lastObjectTransform;
@@ -57,7 +57,7 @@ public class GLCanvasDrawer implements CanvasDrawer
   {
     this.view = view;
     GLCapabilities caps = new GLCapabilities(GLProfile.get(GLProfile.GL2));
-    canvas = new GLCanvas(caps);
+    canvas = new GLJPanel(caps);
     canvas.addGLEventListener(new CanvasListener());
   }
 


peteihis:
> In principle that hurts performance a little bit

Actually just the opposite. Arranged a little test and it was faster. Marginally but still.

peastman:
Interesting.  How did you measure it?

peteihis:
With a little practical hack to ViewAnimation:

- `timer.setCoalesce(true)` so all frames will be processed
- At the end of the move&rotate `start`-method (just before the `restart`)  save curretTimeMillis to `t0`
- At the beginning of `endAnimation` save milliseconds to `t1` and print the difference to System.out.

Then I set the framerate (preferences) to 120 fps _(vs my 60 Hz HW)_, had a scene with just one cube on it and pressed on the number pad 5,4,3,2,1 always waiting, that the animation had stopped, before pressing the next key. ...And I cleared the folders before each compile. _(For a side track, the coalesce set to false does have an effect, but it cannot handle the 120 fps on 60 Hz gear .. ... Not the way I wrote it anyway ...)_ 

The results from above are below. With different settings there could be some variation between two silimar cases, but this time it produced a perfect series. :D

~~~~txt
GLCanvas
Steps 61, time 1000
Steps 61, time 1000
Steps 86, time 1406
Steps 61, time 1000
Steps 86, time 1406

GLJPanel
Steps 61, time 969
Steps 61, time 969
Steps 86, time 1359
Steps 61, time 969
Steps 86, time 1359
~~~~

peastman:
If I understand you, what that's showing is just that it has no trouble hitting 60 fps in either case?  That's not surprising, given that your scene just contained a single cube.  I'd be a little worried if it couldn't!  But also, the overhead from using a GLJPanel is independent of the scene complexity, so that suggests even on a much more complicated scene, it will have little impact.

Sailsman63:
Worked without issues on both my Windows native and Linux VM. I think this one is safe enough to release to our Mac users in the next nightly.

peteihis:
> ...that it has no trouble hitting 60 fps in either case? 

Mostly so. The idea was that there ae least would not be any waiting extra waiting time between the frames.

But I came up with a different plan: I tested with a scene with a 17x17x17 array of spheres and captured nanotime before and after `view.repaint`. The sum of all used nanoseconds with the procedure before was  8 908 682 with GLJPanel and 15 854 936 with CLCanvas. -- Then I'm not sure how much of that is creating the image and how much somethog else, but at least the panel seems to be doing pretty well in that comparison. :)

PR #60, opened 2017-09-15
Performance fix in 'mouseMoved()'

peteihis:
- A performance fix in `mouseMoved()` method: Update only the view where the mouse is (Object viewer. line  513 and SceneViewer, line 710). -- The update is stilll rather heavy thingking of the need. 
- Refactoring: Cleand tabs while I was at it.


peteihis:
Hold on, I found a side effect...

PR #62, opened 2017-09-18
Small math and loop opimization for GLCanvas drawer

makiam:


peastman:
Does this actually make a measurable difference?  These are all optimizations that I'd expect the compiler to be handling automatically.

peteihis:
I don't know how smart compilers are, but I have always thought, that creating a variable and using it just twice is not any better than calculating the value twice unless the calculation is very complex... But that is just a rule of a thumb from somewhere in the ancient history. This probably does not hurt either.

peastman:
Common subexpression elimination is a standard optimization in any compiler.  If you're doing the same calculation twice, it will recognize that and consider storing the result.  But it does a lot of sophisticated analysis to decide whether that's actually a good idea or not.  If you're short on registers, it might decide it's better to recalculate the value than to waste a register on it, or write it to memory which can be very slow.  So I'd be surprised if this change actually makes the code run faster.

makiam:
I try to measure a bit drawboxes() code running it under profiler. 
Not shure that I do this 100% correct and sharp... 
moving isPerspective() call out of loop gives us reducing it calls 8x times.
This gives us on 50x times attempts 0,058 ms in old code vs 0,009 mswith new one for it
All Self time for drawBoxes() is 4,98 ms with old code vs 9,32 ms with new one

peastman:
It's best to use `System.nanoTime()` to measure it directly.  Running in a profiler will often disable a lot of compiler optimizations.

I could imagine a benefit from moving `isPerspective()` out of the loop.  The rest of it seems unlikely to help, and might hurt.  So try measuring the effect of each change separately.

Sailsman63:
The other question is: How much of each repaint is actually spent in this code? If its only a small fraction, you could double the speed here, and have very little effect on the responsiveness that the user sees....

makiam:
found odd... when there are some selected objects drawBoxes() calls on every mouseMoveEvent over canvas...

peteihis:
>  every mouseMoveEvent

Unfortiunately you are right.

In the current Git HEAD mouseMoved calls repaint on every view, every time the mouse is moved. That definitely should not be the case. 

MouseMoved is the only way to get instant feedback on where the mouse is. Currently there is not other way to get the feedback on the screen but do full `updateView()`. At the moment there is only one thing, that needs this feature, and it would do fine, updating just the active view but I have something coming that should be updated to all views at the same time.

I think there is a way to handle it so that a full update is not needed. I'm have tested the idea a bit: https://sourceforge.net/p/aoi/discussion/47784/thread/62f50608/?limit=25#e71e ... I think it should not be very overwhelmingly difficult to make work.

peastman:
> At the moment there is only one thing, that needs this feature

What's that one thing?

peteihis:
>> At the moment there is only one thing, that needs this feature

> What's that one thing?

It is the steering cues in Fly and Drive modes. Iäm not entirely happy with those in their current form, but I think something of  that kidn should be there. Those travel type navigationmodes generally speaking are hard to handle. But then the next thing coming would be the 3Dmanipulatior, which hilights the handle, that the mouse is over.

makiam:
I do a new measurements with JMH benchmark.
Moving code out of loop has no effect...
Introducing new floats - gives execution time increase....

peastman:
> It is the steering cues in Fly and Drive modes.

Ok, I've got to ask.  What are those modes???  I just spent some time messing with them, and I have no clue.  Whenever I move the mouse some lines and circles that track the mouse position briefly appear, then disappear again.  They don't seem to have any connection to what tool is currently selected.  I can't figure out what they're supposed to tell me.

Anyway, if you just want to redraw one view, call repaint() on it.  There's no need to redraw all four views for that.  But you also should only repaint a view if there's actually a reason to.  For example, if this is only used in Fly and Drive modes (whatever those are), don't trigger a repaint when you're in Space or Tray modes (whatever those are).

Sailsman63:
@peastman Those are all a result of the View Manipulation changes. Basically, they are distinct "Modes" for moving an interactive view. They operate through the shortcut keys only, not through "Move" and "Rotate" ToolPalette options. 

Try Alt+drag, Ctrl+Alt+Drag, and scroll.

The controls work like this:

- Alt + Left Drag = Center Drag = Rotate view about (View) X and Y axes
- Ctrl + Alt + Left Drag = Ctrl + Center Drag = Rotate View around (View) Z axis
  - This control displays a "Clock Dial" cue graphic when in motion
- Ctrl Right Drag = Scroll Wheel = Move along (View) Z axis in perspective mode, Zoom in Parallel mode
- Right Drag = slide along (View) X and Y axes (Strafe)

- Space and Tray modes center view rotation around a point in front of the view. 
- Alt + Left Click = Center Click = pick center of rotation. 
  - If click on object, Object is now center
  - If not click on object, center is projected at same z-depth as old center
- Z-axis move takes the rotation center along with it, at constant distance from viewpoint
- Space rotations are unconstrained
- Tray rotations lock out Z-axis rotation and constrain results of other rotations so that (Scene) Y always projects as parallel to (View) Y

- Fly and drive modes center rotation at the viewpoint.
- These are the ones with the crazy cue graphics
- Z-Slide in center of cue "Sight Ring" moves straight along (View) Z.
- Z-Slide outside cue center, but inside cue border, adds rotation in direction of mouse cursor.
  - magnitude of rotation is dependent on distance from center
- Fly is unconstrained, Drive is constrained similarly to Tray
  - Z-Slide outside of cue shifts to rotation in direction of mouse cursor in fly mode
  - Z-Slide in drive shifts to rotation around Y axis/Slide along Y axis, depending upon position of cursor. I'm not exactly a fan of this mode.


I'd say that this is the final point that shows that these cues need serious work. I'd hoped for some community feedback, but have not heard anything substantial until now.

I'm sorry if these are a surprise to you. The various options were in fact the entire thrust of the View-manipulation project. (It did grow much bigger in the doing, and Pete and I both learned a lot about breaking something down into smaller chunks.)

peteihis:
> Anyway, if you just want to redraw one view, call repaint() on it. There's no need to redraw all four views for that. But you also should only repaint a view if there's actually a reason to.

That is the plan. 

I have been workin on a way to update just the overalays, not to repaint all of the view and as said, I'm also workin on something (other than the fly/drive modes) that needs overlays updated on all the views, not just one. That can be done without a full repaint, but it regures some new addition to the CanvasDrawers.

The navigation modes:

When I (and everybody else at the time, some time in 2003 or 2004) first encountered view modes like fly, walk, drive, we did not even realize, that there was such a thing as different rotation/navigation modes. Just wondered, why the view was all of a sudden behaving very strangely and, _"What the heck are those idiotic arrows and aim-crosses appeaing on the screen?"_ -- And we did not know, what the small icons with a car and an airplane were for, nor paid attention to the fact, that he parallel/perspective switch was disabled if you pressed the airplane icon. Nobody had told us. I doubt, if even our CAD instructor knew.

The idea of drive and fly modes is, that when you have a model of a large environment like a building or a city block you can place yourself in it an move "naturally". Those modes make more sense with wider fov-angles, than what is good for modelling work (think of the viewing of games) and of course, when you don't know what's happening, it is confusing But they come in handy in designing camera drives and so forth.

> The various options were in fact the entire thrust 

To me the most important driver to get started was the very frustrating fact, that if I did not have something selected and wanted to rotate the view, the software selected a rotation center on it's own and the detail, that I was working on disapperad from the view. -- So in fact the software was making decisions, that the user did not want.

This is actually a problem in most of the mainstream state of the art CAD-software even still today. If you want to rotate aroud a point that you are working on, you have to know to select the correct tool for that, which alway breaks your work flow. They have chosen a way to do it 20 year ago and haven't changed it though already 20 years ago there was the modeler friendly way, that kept your _point of focus_ on the view without having to worry about it.

peteihis:
About the scroll cues: I had it in my mind that there should be a text that says something about using the scroll wheel but.... The GLCanvasDrawer can not produce textlabels with a transparent background.

I really should get the introduction video done, but the UI still should tell you, what to do and what to expect....

peastman:
But why do the cue overlays appear every time I move the mouse, then disappear again a moment later?  If these only relate to manipulating the view with modifier keys, they shouldn't be drawn when I'm doing something unrelated.

I think there also are some bugs, because the modes don't actually work as you described.  For example, in Space and Tray modes, right drag moves the view and center drag rotates the view (as you said).  But in Fly and Drive modes, center drag moves the view, exactly like a right drag.

Sailsman63:
>But in Fly and Drive modes, center drag moves the view, exactly like a right drag.

@peastman No, when in fly/drive mode, the camera rotates around the _camera_ origin. At default FOV, and typical object sizes, this _looks_ similar to a sideways slide. To test, zoom one of the other views further out, until you can see the entire frustum when manipulating the view.

Or, try using the mode to manipulate the view through a scene camera, and watch the object properties.

>why do the cue overlays appear every time I move the mouse, then disappear again a moment later?

I think that the thought was that the z-slide is always available through scroll wheel, so the cue should remind you every time that you show interest in a view. Perhaps this needs a revisit, among other things.

peastman:
> I think that the thought was that the z-slide is always available through scroll wheel, so the cue should remind you every time that you show interest in a view.

I don't think users need quite *that* much reminding. :)  They can remember that the scroll wheel exists without needing to be reminded of it every single time they move the mouse!

Besides, I don't really see the connection to the cue graphics.  What information is it supposed to convey to them that isn't already immediately obvious?  Where the mouse is?  Where the center of the view is?  That's really all it shows.  And it's very distracting having it constantly popping up and disappearing, especially when it's usually irrelevant to what you're doing.

Sailsman63:
The graphics are an aiming device, because in these modes, the effect of the z-slide depends on the _exact_ position of the mouse. The inner ring is a dead zone, straight z-slide. Outside the outer ring is pure rotation. In between is a blend of the two, depending on the distance from the two rings.

Regarding the always-on cues, they were also intended to make the above more... Organically discoverable.

Given that we are having this discussion, (which probably deserves its own issue topic) I'd say that it's not doing that. I'm quite open to changing that behavior, and indeed replacing the graphics entirely, if we can find a better alternative. (These were always intended as a first draft. I've been waiting for constructive feedback, and yours is the first.)

Sailsman63:
Seems that the original code changes in this PR did not live up to expectations. Closing.

PR #63, opened 2017-09-18
Coordinate axis symbol fixes / changes

peteihis:
Changes:
- The changing size in Camera view stabilized.
- Default axis length 50 -> 30
- Drawing the labels with better precision.
- Made `axisLength` adjustable by set/get per view. (Considered making it static, but ended up with this.)

Could not prevent the flickering of the symbol one pixel right-left-up-down. I don't see, where it is coming from.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -1334,17 +1334,26 @@ public abstract class ViewerCanvas extends CustomWidget
     drawLine(centerPoint, new Point((int) Math.round(screenZ.x), (int) Math.round(screenZ.y)), lineColor);
 
     // Draw the labels.
+	// The drawers handle the 'drawline' with minor differences.
 
+	boolean isGL = (drawer instanceof GLCanvasDrawer);
+	
     if (screenX.minus(screenCenter).length() > 2.0)
     {
       Vec2 dir = screenX.minus(screenCenter);
       Vec2 labelPos = screenX.plus(dir.times(8.0/dir.length()));
       int x = (int) Math.round(labelPos.x);
       int y = (int) Math.round(labelPos.y);
-      drawLine(new Point(x-3, y-3), new Point(x, y), lineColor);
-      drawLine(new Point(x+3, y-3), new Point(x, y), lineColor);
-      drawLine(new Point(x-4, y+4), new Point(x, y), lineColor);
-      drawLine(new Point(x+4, y+4), new Point(x, y), lineColor);
+	  if (isGL)
+	  {
+        drawLine(new Point(x-4, y-3), new Point(x+3, y+4), lineColor);
+        drawLine(new Point(x+3, y-3), new Point(x-4, y+4), lineColor);
+	  }
+	  else
+	  {
+        drawLine(new Point(x-3, y-3), new Point(x+4, y+4), lineColor);
+        drawLine(new Point(x-4, y+4), new Point(x+3, y-3), lineColor);
+	  }
     }
     if (screenY.minus(screenCenter).length() > 2.0)
     {
@@ -1352,7 +1361,10 @@ public abstract class ViewerCanvas extends CustomWidget
       Vec2 labelPos = screenY.plus(dir.times(8.0/dir.length()));
       int x = (int) Math.round(labelPos.x);
       int y = (int) Math.round(labelPos.y);
-      drawLine(new Point(x-3, y-3), new Point(x, y), lineColor);
+	  if (isGL)
+        drawLine(new Point(x-4, y-3), new Point(x, y+1), lineColor);
+	  else
+	    drawLine(new Point(x-3, y-3), new Point(x, y), lineColor);
       drawLine(new Point(x+3, y-3), new Point(x, y), lineColor);
       drawLine(new Point(x, y), new Point(x, y+4), lineColor);
     }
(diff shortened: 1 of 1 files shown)


Sailsman63:
Can you clarify a little what issues you were experiencing? I see the shorter axes, which do feel *"better"* to me, but otherwise, I'm not seeing the difference.

peteihis:
> Can you clarify a little what issues you were experiencing?

May be minor ones to some. :)

Mostly the thing, that got me on it was the camera view, where the size of the coordinate axis symbol did not stay constant. It was changing with both the fov and the view size. Now lines 1309 -1314 take care of that. I can also use that same fix on the 3D-manipulator.

Another thing was, that something looked a bit off with the "Y". The arms were different. So I ended up adjusted them all. 

The position of the Y-label was also traveling left and right (=flickering) by one pixel, when rotating horizontally _(and I believe, so were the rest of them in other directions)_ which sometimes lead me believe, that the entire symbol was shaking. _(In some of my test versions it actually was shaking)_.

The size adjustment a kind of just came along.  :)

peteihis:
OK.. I know, people have come up with inventive names to people who spend time on this kind of thing... Now dedicated labels for both of the drawers.

Sailsman63:
Thinking that this would be much simplified if we got the concerns in #66 sorted out first.

peteihis:
> Thinking that this would be much simplified if we got the concerns in #66 sorted out first.

The `drawCoordinateAxes()` is still exactly the same, what ever happens with #66. And the size issue with SceneCamera and the drawers' different handling of the labels is not addressed there either.

To draw the labels with real fonts the GLDrawer would need to be taught to draw text with a transparent background, but it would not look pretty at all because text would require antialiasing to transparency and the GLD cannot do that in it's current state. It can do transparency, but it is entirely an on/off type of a thing.  -- In that sense you have a point, that if the #66 would produce a propper layer system, then the coordinate symbol could be drawn on a layer that does not have to go through GL.

peteihis:
To clarify a bit: 

I don't think that tha #66 should be a part of the next release. There are many already many new features and bug-fixes that are ready to be released. -- And I feel a bi guilty about the condition of a couple of the navigationmodes.

Here are a couple of the reasonf to do this:

First, the coordinate symbol is jumpy. It is worst in 4-view mode with the SW renderer. With GL-only the letters travel.
coord_jump

Second, as the X Y and Z are drawn as lines each renderer interprets the information in it's own way:
xyz_symbols

And third, the coordinate symbol of the camera view changes size, when you switch between 1 an 4 views. It's just that it has been botherin the small perfectionist inside me for a very long time, how unprofessilal that looks... :D

Neither of the drawers actually seem to follow a very solid logic in how they handle the `drawLine` in pixel level. _(If at #66 the coordinate symbol is going be drawn by the Java AWT tools, then it's again a different story entirely.)_

If you feel that something should be changed, please let me know. (Changing the default size is now just changin one number, resizing the axis markers a bit more work.)

Sailsman63:
I've seen the flickering issue, and agree that it needs to be fixed eventually. Looking at it now, I think that the software drawer is, perhaps, _truncating_ the line coordinates that it receives, rather than rounding them, or perhaps uses slightly different rounding rules? Something to check.

I'm concerned about the `if isGL()` stuff. Is this just to make the letters exactly the same shape in the different drawers? `ViewerCanvas` and company should have no idea about which drawer is being used. The fact that the letters tend to float a bit makes me think that this is another area where various bits of the software make different assumptions about rounding of coordinates...

We really should not need extra code to compensate for camera FOV. I'm not sure why it has an effect on the indicator in the first place. Perhaps finding the root cause of that would make more sense?

Looking closer, we could possibly get this cleaned up without waiting for #66.

RE: the view manipulation cues/modes: Don't worry about it. We do need to do some cleanup, but we both learned a lot about trying to keep projects manageable. (That's why I've become so ultra-cautious about PRs. I'm trying to compensate for my own limitations, and make sure that the changes we make don't have unexpected consequences)

peteihis:
> I think that the software drawer is, perhaps, truncating the line coordinates that it receives, rather than rounding them, or perhaps uses slightly different rounding rules? Something to check.

Actually the truncating happend in the `drawCoordinateAxes()` code. In the process the axes start as a 3D-entity, but are drawn as pixels. The conversion from 3D to 2D happens by `Camera.getWorldToScreen`. Then to use that for drawing pixels, `double`s need to be turned into `int`s. That's where I put `Math.round()` instead of just `(int)`. That took care of the flickering of the axis lines. The flickering was depending on the view size too. I guess it was (at least partially) about the horizontal pixel count being odd or even.

>the if `isGL()` stuff. Is this just to make the letters exactly the same shape in the different drawers?

Yes.

The letters are drawn as pixels without the 3D-side interfering.  Unfortunately ther is no `drawPoint()` but `drawLine()` has to be used. The drawers just read the same `drawLine` code in a different way. You can actually see this in a lot of places, but the letters are something where you pay attetion to it. And BTW using `Math.round()` to calculate the location of the letters tamed down the frickerin of the Y in GL-drawer too.

> `ViewerCanvas` and company should have no idea about which drawer is being used. 

Generally speaking I agree. The small carvings on the screen are a bit of a special case though, hence the hacking. :)

If you look at the lines 1349 to 1355 in the modified file, the 3's and 4's do not make much sense but somehow they manage to produce the same out come. I actually found those simply by brute trial and error. Going into the drawers to unify their behavour would probably be a really big task and not really worth it. The differences are basically only a pixel here and another one there, which many of us never notice.

> We really should not need extra code to compensate for camera FOV. 

Actually we do. Explaining it in just words completely would ... ehh .. _not happen ?_ ...  But I'll try to shed some light on it:

The usual editing views use two parameters `scale` and `distToScreen` to define how big the objects appear on the screen. In perspective mode, these have a fixed ratio (actually fixed values) and the view size is entirely ignored. You can see, that the perceived perpective or _the size_ of the objects are not affected by switching 1 to 4-view or rescaling the windwow. So thinking of FOV, FOV does not exist as a parameter, but a larger view is just _behaving_ like it had a larger FOV. However...

The `ScenerCamera` does have FOV as a parameter. This changes everything. 

When you switch1-4-viewing or rescale a camera view, the effect is like recalculating `scale` and `distToSceen` to match a fixed FOV with the new view size. Further more, changing the FOV itself has the same effect again. This is fine as long as the objects are _supposed be scaled_ along with the view. Now the coordinate symbol actually is a 3D-object in a 3D-space, similar to the scene. So it is subject to the same perspective and scaling that happen to the objects in the scene. That's why the compensation of size has to happen somewhere.

Basically you could rescale the axes before the `worldToScreen` operation, you could change the (hypotetical) distance between the symbol and the camera or you could rescale the 2D-vectors after the conversion. .. or you could have a special 3D-space for the axes either "flat" or with a fixed FOV -- I see this kind of behavior in other software... In this case I found it easiest to use the same space and rescale the length of the axes before the 3D-2D conversion. Also this way the conversion matrix is already there and the perspective of the axes stays similar to the scene even with large FOVs.

Of course there is a lot more to it  but I hope this is enough for now. :)

peteihis:
> RE: the view manipulation cues/modes: Don't worry about it.

Last weekend, I finally got to have look at a few things. I'll suggest some simplifications, as "soon" as I have something working.

> make sure that the changes we make don't have unexpected consequences

Keep up the good work. :)

peteihis:
... Or the `SceneCamera` should have a `getDistToScreen()` that would calculate the parameter from the FOV and the `Camera.distToScreen()` should then be matched with the SceneCamera parameter at all times when a SceneCamera is present ... 

That would actually simplify quite a few other things as well, but the effects are probably scattered to all over the code ... I seem to have had this idea for a while already, but never got to making it happen ...

peteihis:
This on has been waiting too long.... I have, more compact solution on the way.

PR #65, opened 2017-09-30
Bug fix to an 'addObject()'-simplification

peteihis:
This simplification from 3.0.3 causes an error if objects have parent/child relations in the object tree and new objects no longer appear in the object list.

The assumption has been, that the lenght of the object tree would be equal to number of objects in the scene. However this is true only as long as all objects are in the root level of the object tree. Otherwise the length of the object tree is the number of the stand-alone parent objects.

This modification changes the method to, what it is in 3.0.3.

_I don't remember, that I would have done this change myself, but I just as well might have...._

### DIFF
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -1167,10 +1167,9 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
     {
       public void run()
       {
-	    System.out.println("@run");
         itemTree.setUpdateEnabled(true);
         for (int i = 0; i < theView.length ; i++)
-          theView[i].rebuildCameraList();
+             theView[i].rebuildCameraList();
         theScore.rebuildList();
       }
     });
@@ -1190,10 +1189,9 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
     {
       public void run()
       {
-	    System.out.println("@run");
         itemTree.setUpdateEnabled(true);
         for (int i = 0; i < theView.length ; i++)
-          theView[i].rebuildCameraList();
+             theView[i].rebuildCameraList();
         theScore.rebuildList();
       }
     });
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -122,8 +122,8 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
       theView[i].addEventLink(MousePressedEvent.class, listen);
       theView[i].addEventLink(KeyPressedEvent.class, keyListener);
       theView[i].setPopupMenuManager(this);
-	  theView[i].setViewAnimation(new ViewAnimation(this, theView[i]));
-	  theView[i].setNavigationMode(1,false);
+      theView[i].setViewAnimation(new ViewAnimation(this, theView[i]));
+      theView[i].setNavigationMode(1,false);
     }
     theView[1].setOrientation(2);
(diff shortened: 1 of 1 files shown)


peteihis:
> the last position in scene.

The problem is not with the scene. It is with the **object tree list** on the right side of the view. 

The `addObject` in question was calling the next `addObject` using `theScene,getNumObjects()` as the index, where to add the new item in the **object tree**.

When at least one object is made a child to another, the tree is shorter than the number of objects in the scene. Then adding a new item results `ArrayIndexOutOfBoundsException` at `artofillusion.ui.TreeList.addElement` and though you can draw more objects on the view and they will be added to the scene, the object tree does not get updated any more.  You can easily verify this yourself: Draw a few objects, move one of them under another in the object tree and draw some more objects.

In it's 3.0.3 form the method simply adds the new item into the end of the object tree,

Sailsman63:
@peteihis This one might be worth a test case in `Tests`.

Unfortunately, I've not been able to duplicate this problem through the interface, which is probably because `LayoutWindow` does not call this method at all... probably only happen with custom tools or scripts.

peteihis:
> probably only happen with custom tools or scripts.

Nope. Just with the bare core AoI. I have not tested with scripts. See this 
https://www.youtube.com/watch?v=RLvW9OOWcIo&feature=youtu.be (set private).

I cloned the master just for that and there aren't any extras installed. You might also notice, that the tool handles that are supposed to indicate the selection (latest addition) freeze after I move the 2nd Shere under the first.

peteihis:
Just noticed another effect of this problem: When the ObjectTree has stopped working, drawing curves goes into "eternity mode". Drawing curves never stops thoug I try to select a different tool. It is possible to finish a curve, but the next click just makes a copy of it and continues drawing ....

Sailsman63:
>See this
https://www.youtube.com/watch?v=RLvW9OOWcIo&feature=youtu.be (set private).

You need to set this as `unlisted` to allow others to see it.

peteihis:
Sorry. Done. -- I don't kow why YouTube have made it so extremely difficult to find my own uploaded videos ... but as I keep clicling the mouse and switching accounts eventually I seem to get there...

Sailsman63:
Thanks! I thought that I had tried that. Its even worse than you thought... those (unlisted) objects can be selected and moved in the scene, but they cannot be deleted!

Sailsman63:
Unfortunately, this change rolls back _all_ of the performance gains from #59 when adding objects. I'm going to take a closer look, it seems that perhaps this is not the right place for this update loop to be run.

peteihis:
> Unfortunately, this change rolls back all of the performance gains from #59 when adding objects.

OK, I did not realize / pay attention to that.  Sorry.  I dug a little deeper...

In `Scene` there are three `addObject`s and  addObject_#1 is calling addObject_#2, which is calling addObject_#3. The last one uses an index number and that's fine because tode addObjects don't call anything else.

Now after #59, LayoutWindow is using that same structure, but it does not take to account the fact that it is communicating with a `Scene` and a `TreeList` whose lengths mignt not match. 

In 303 the `LayoutWindow.addObject_#2' calls the methods of  `Scene` and `TreeList` _without_ the index number allowing _them_ to decide where to put the additions. The only call to the `LayoutWindow.addObject_#3' _with_ index, that I colud find is in `UndoRecord`, where it seems to be working corrently. I'd suggest to add a warning there that using the method in other contexts is not safe and have `addObject_#2` call `Scene` and `TreeList` w/o an index as before but with the performance improvement included.

I did some testing and, as far as I could tell, that seems to work as expected. -- I'll Push an update soon.

Sailsman63:
Thanks! looks good. I did remove a couple of `println()`s that you must have used for diagnostics.

Issue #64, opened 2017-09-28
Changes to navigation modes


peastman:
This is to continue the discussion that was started at #62.  I think the new navigation modes could use some changes.  Let me go through some of the issues I encountered when working with them.

In Fly and Drive modes, there are cue graphics that appear every time you move the mouse, then disappear after a few seconds.  I see several issues with this:

- It's really distracting having the overlay constantly appearing and disappearing!  It gets in the way of trying to do other things.
- The constant redraws hurt performance.
- It's only relevant to a very specific function (moving the camera along the z axis), so having it appear when you're doing other things isn't useful.
- It's not obvious what the graphics mean.  For example, in Fly mode you get inner and outer circles, connected by a line passing through the mouse position, and a third smaller circle around the mouse position.  Apparently the inner circle is to indicate the central dead zone, but what about the rest of them?
- It's not even clear to me that the graphics are useful.  Users should be able to just think of it as, "Use the scroll wheel to zoom in on whatever the cursor is pointing at."  And mostly, that seems to work pretty well.  Perhaps having a central dead zone makes the behavior feel a little more natural, but there's no need to draw a circle around it.

Another question is whether we can make it more intuitive to the user what these modes mean.  When I first saw a menu with the word "Tray" above every view, I didn't know what it meant.  It wasn't obvious that had anything to do with navigation.

When you go into Fly or Drive mode, it disables the parallel/perspective menu and forces it to perspective mode.  Why?  Is there a reason these modes can't be used in parallel view?

In any case, there's a bug in the implementation of forcing it to perspective mode.  Try the following:

1. Start out looking at a front view, parallel, tray mode.
2. Switch to fly mode.  It disables the parallel/perspective menu and shows a perspective view.
3. Use the scroll wheel to zoom in and out a few times at different places.
4. Switch back to tray mode.  It reenables the menu and claims to be in parallel mode, front view, but the  view is obviously not correct.  For example, I can see the sides of an axis aligned box!
5. Reselecting either parallel mode or front view doesn't fix it.  But if I first select a different view direction (back view, for example) then select front again, that does fix it.

Sailsman63:
A user [on sourceforge](https://sourceforge.net/p/aoi/discussion/47782/thread/2c1a2f05/?limit=25#5b81) recently tried to run AOI under Java 9. In addition to issues with the wrapper `.exe`s,

We've found out that there are issues with the view controls under java 9 on Windows.
[...]

PR #66, opened 2017-09-30
Drawig overlays independent of 3D-repaint made possible

peteihis:
Quite a few files edited, but that's, what it took to get the basics work.

Now there is a possibility to update views without repainting the scene on the view. For example tool handles and the 3DManipulator can be drawn without a full repaint.

At every repaint the CanvasDrawer saves a snaphot of the view after painting the scene. If only overlays need updating, calling `updateOverlays()` has `updateImege()` use the cached snapshot instead of recalculating the scene image.

In future this should enable things like selection pre-hilights and drag-ghost shapes on the sceen.

Currently the cached images are kept "hard" in the memory.  They probably should be handled as SoftReferences, unless there is somethign more to it than just that? _(There are some even more advanced ways of handling data in the Canvas Drawers)_

This PR also contains the changes in #65.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -989,6 +989,16 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
       }
   }
 
+  /** Update the overlays displayed in all of the viewport. 
+      @param view The view whose overlays don't need updating. If null, update all. */
+
+  public void updateOverlays(ViewerCanvas view)
+  {
+    for (int i = 0; i < numViewsShown; i++)
+      if (theView[i] != view)
+        theView[i].updateOverlays();
+  }
+
   /** Update the state of all menu items. */
 
   @Override
@@ -1162,6 +1172,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
   public void addObject(ObjectInfo info, UndoRecord undo)
   {
     theScene.addObject(info, undo);
+<<<<<<< HEAD
     itemTree.setUpdateEnabled(false);
     itemTree.addElement(new ObjectTreeElement(info, itemTree));
     uiEventProcessor.addEvent(new Runnable()
@@ -1174,6 +1185,12 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
         theScore.rebuildList();
       }
     });
+=======
+    itemTree.addElement(new ObjectTreeElement(info, itemTree));
+    for (int i = 0; i < theView.length ; i++)
+      theView[i].rebuildCameraList();
+    theScore.rebuildList();
+>>>>>>> 09dd8fd... Drawig overlays independent of 3D-repaint made possible
   }
 
   /** Add a new object to the scene.  If undo is not null,
@@ -1375,7 +1392,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
       which[i] = theScene.indexOf((ObjectInfo) sel[i]);
     setUndoRecord(new UndoRecord(this, false, UndoRecord.SET_SCENE_SELECTION, new Object [] {getSelectedIndices()}));
     setSelection(which);
-    updateImage();
+    updateOverlays(null);
   }
 
   private void displayModeCommand(CommandEvent ev)
@@ -1481,7 +1498,6 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
     theScene.clearSelection();
     itemTree.deselectAll();
     theScore.rebuildList();
-    updateImage();
     updateMenus();
   }
 
@@ -1893,7 +1909,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
       which[i] = i;
     setUndoRecord(new UndoRecord(this, false, UndoRecord.SET_SCENE_SELECTION, new Object [] {getSelectedIndices()}));
     setSelection(which);
-    updateImage();
+    updateOverlays(null);
   }
 
   public void preferencesCommand()
@@ -2501,7 +2517,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
         info.setLocked(locked);
       }
     setUndoRecord(undo);
-    updateImage();
+    updateImage(); // Does something really change?
     itemTree.repaint();
   }
 
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -1172,7 +1172,6 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
   public void addObject(ObjectInfo info, UndoRecord undo)
   {
     theScene.addObject(info, undo);
-<<<<<<< HEAD
     itemTree.setUpdateEnabled(false);
     itemTree.addElement(new ObjectTreeElement(info, itemTree));
     uiEventProcessor.addEvent(new Runnable()
@@ -1185,18 +1184,12 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
         theScore.rebuildList();
       }
     });
-=======
-    itemTree.addElement(new ObjectTreeElement(info, itemTree));
-    for (int i = 0; i < theView.length ; i++)
-      theView[i].rebuildCameraList();
-    theScore.rebuildList();
->>>>>>> 09dd8fd... Drawig overlays independent of 3D-repaint made possible
   }
 
   /** Add a new object to the scene.  If undo is not null,
       appropriate commands will be added to it to undo this operation. <P>
-	  
-	  NOTE! This method is only used by 'UndoRecord'. Using it in any other context is not safe. */
+
+      NOTE! This method is only used by 'UndoRecord'. Using it in any other context is not safe. */
 
   public void addObject(ObjectInfo info, int index, UndoRecord undo)
   {
--- a/ArtOfIllusion/src/artofillusion/MoveViewTool.java
+++ b/ArtOfIllusion/src/artofillusion/MoveViewTool.java
@@ -65,7 +65,7 @@ public class MoveViewTool extends EditingTool
   {
     Camera cam = view.getCamera();
 
-	selectedNavigation = view.getNavigationMode();
+    selectedNavigation = view.getNavigationMode();
     controlDown = e.isControlDown();
     clickPoint = e.getPoint();
     clickPos = cam.convertScreenToWorld(clickPoint, view.getDistToPlane());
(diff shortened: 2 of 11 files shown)


Sailsman63:
There should be no need to cache data like this.

It is entirely possible to render a (transparent) `JPanel` directly on top of another one, and have the renderers deposit overlays, etc. in the top layer. This would be more of an optimization that is handled at the `CanvasDrawer` implementation level, and should be a lot less intrusive to the codebase.

As an aside, please do not include changes from another active PR in something like this. It makes things messy if we needed to accept this PR without accepting the other one.

peteihis:
> This would be more of an optimization that is handled at the CanvasDrawer implementation level, and should be a lot less intrusive to the codebase.

That was the general idea.... It happens inside the CanvasDrawer. What happens outside is just the decision whether to call for a `updateImage()` or `updateOverlays()`.

True enough that `JPanel` / `JLayeredPane` looks like the right tool for the job, especially thinking of the possibility of splitting the job into even more layers, but right now I wouldn't know how to make that happen... It would seem to me, that using layered panels would actually take the work at least in some part to ViewerCanvas, and would start by changing the component to a JLayeredPane.  (Much the same way, that setting GLDrawer does now and changes it to the GLJPanel of the drawer.... )

Would that sound like the way to go?

> if we needed to accept this PR without accepting the other one

In this particular case it might be just the other way around... :)

peteihis:
Current tool was drawn over the view border. I don't think it was meant to be that way. Affected 4 files. Removed an unused private method.

Sailsman63:
> Current tool was drawn over the view border.

Can you clarify? I'm not seeing a practical difference. Perhaps a graphic of what you mean?

peteihis:
Ok. This probably is only seen, if the current tool happens to be the 3D-manipulator.

In this forst picture the view border is drawn as it should, after the object, but the tool is, againd all logic, drawn after the border.

over_the_edge_303

On this other one the tool is drawn before the border. 

under_the_edge

For some reason, the CanvasDrawers first call `updateImage()` of the (sub class of) `ViewerCanvas` and then call `currentTool.drawOverlay()`, so the tool graphics get drawn _after_ the view border. When the call is moved to the updateImage method before the drawBorder, it shows up where I would expect to see it. Also I have no idea, why that call would be the only drawing call in the CanvasDrawer, when all the rest of the drawing is done by the updateImage methods...

Another thing, that you might notice is that the GLCanvasDrawer draws the border in a wrong place. It is one pixel down and one pixel left. The SWCD draws it correctly.

peteihis:
I seem to be missing something.

I tried to build a layer system out of JPanels and a JLayeredPane but I only get blank views wihout even a bacground color.

The compiler didn't see anything wrong with it and it was (more or less) working as a script but in the code it turned out, that the `paint` method (in SWCD) or the `updateImage` were never even visited. It seems that the `CustomWitget` and `GLJPanel` have something that a `JPanel` does not and what ever is wrong, is handled silently in the back. 

I probably should sepatarte some things, that are not that directly related to the overlay drawing into their own PRs.  (Though it feels nicer to fix all sorts of thing in a file, that I'm working on, featurewise that may not be the best approach...)

peastman:
The class you want is `OverlayContainer`.  See https://sourceforge.net/p/buoy/code/HEAD/tree/src/buoy/widget/OverlayContainer.java.  The documentation describes how to use it for this purpose.

peteihis:
How do I connect it with the ViewerCanvas? So far I the reults are just the same as with the J-parts.

I can see that the OverlayPanel is there. Changing it opaque or not has an effect, but that's all.

peastman:
The OverlayContainer has two children: the ViewerCanvas (lower layer) and a CustomWidget (upper layer).  Anything drawn on the CustomWidget will appear in front of the ViewerCanvas.

peteihis:
>  the ViewerCanvas (lower layer) and a CustomWidget (upper layer).

That seems to explain something ... My original problem was, that had missed the `layoutChildren()` which makes the layers visible (doh...)

So far I managed to test, that the background including the template image on one layer and the rest on another works fine. Now I'd have to teach the drawer to draw overlays on their own layer. and I can think of uses for a few more lauerd, but let's get those some time later.

One question: If I get this right, the drawers use 7-bit coding for the view images? Any reason for that?

peastman:
> If I get this right, the drawers use 7-bit coding for the view images?

What do you mean?

Sailsman63:
> Now I'd have to teach the drawer to draw overlays on their own layer. 

I'm wondering if it might be better to have `ViewerCanvas` itself extend or contain an `OverlayWidget`, with a couple of nested `CustomWidget`s, and then explicitly pass the appropriate widget to the `CanvasDrawer` at the time of the draw call.

This would prevent the need for the drawer to know anything about layering, and would make it easy to change layering behaviors by simply extending or modifying the `ViewerCanvas` (Such as adding the ability to render a transparent copy of an object into the overlay layer when moving, rotating, or scaling operations are going on.)

Sailsman63:
>I probably should sepatarte some things, that are not that directly related to the overlay drawing into their own PRs. (Though it feels nicer to fix all sorts of thing in a file, that I'm working on, featurewise that may not be the best approach...)

A bit of a side note, but yes, please. This also helps us in the future if we wonder why, exactly, a line of code is written the way that it is. The the commit in which it was added should tell us exactly what was being worked on when that code was written.

peteihis:
> I'm wondering if it might be better to have ...

OK, it took me a while, to understand what Peter was acually saying before, :D --  So far I actually put the `OverlayContainer` into the `CanvasDrawer`, and used that (the _component_ of it actually) as the "canvas" that replaced the component of ViewerCanvas, which again lead to having to write a new `repaint()`method to `ViewerCanvas`. 

Now there is a choice to be made how the drawer knows on which layer to draw. I could be done so that each method draws on a specific layer, but it would be more flexible to select/set the active layer for each set of actions.

Basically having the layers in the `ViewerCanvas` rather than the `CanvasDrawer`should add more flexibility, like having more than one drawers for different purposes, maybe as plugins in future...

Let's see where this leads to, I'll just make one first.

>> If I get this right, the drawers use 7-bit coding for the view images?

> What do you mean?

The `GLCanvasDrawer` uses `UNSIGNED_BYTE`. at least in some of the operations. I thoink this is a 7-bit/byte format If I remember right and my eyes are seeing, what I think they are seeing this made smooth color gradiens (like of a template image) appear contoured, only every second colordepth used. I'll have to revisit the subject.

The `SoftwareCanvasDrawer` uses `TYPE_INT_ARGB_PRE`, which actually is an 8-bit/byte format, but when I used wrong parameters to copy `theImage` into another `BufferedImage` the default, was that the alpha channel would have been set to 127, which reduced the color depth to 7-bit/byte again... That can be handled better, but if the background is on a different layer than the renderd image, then it seems that the transparent mode to work right will require the alpha channel to be `real`. -- A step closer to real transparency, but still not quite. (What happens with GLCD and transprency I don't know yet...)

peteihis:
I was experimenting with the SWCanvaDrawer. 

The layersystem seems to work nicely with four layers: backgound, scene, overlay and mouse feedback. The last one is for the dragged selection frames and alike. _Later on I'd probably want to add one for dragging-ghosts between 'scene' and 'overlay'. This would be to avoid rgenerating the scene layer while objects are being dragged. Should help with large scenes._

Now obviously there are things, that do not much benefit from GL and are common to both of the drawers, so it would seem logical to convert the CanvasDrawer interface into a (abstract) super class and put the common functions in there. -- Or maybe create a third canvasDrawer to handle dragged shapes on screen texts and so on? -- At the moment the super class seems like a better idea to me. Any comments on that?

Now I, of course, don't yet know, how the GLJPanel works with this stuff but, no doubt, I'm bound to find out....

Sailsman63:
>four layers: background, scene, overlay and mouse feedback.

Do keep in mind that each layer is going to require memory to hold its image. I don't see any case where you would render the background separately from the scene, other than with a Template Image, which I consider obsolete. They can share a layer.

Most non-mouse overlays should be fairly cheap, so they can share a layer with the mouse queues. Drag-ghosts would probably fit in this layer as well.

peteihis:
> Do keep in mind that each layer is going to require memory to hold its image.

Yes. definitely.... This is about finding the balance between memory usage and performance. .... Which thinking a kind of loses it's relevancy by it self: As the HW gets more powerful it's easier to use more memory to improve performance, which is anyway improved as the HW is more powerful.... 

Basically two layers can do the job, but adding more of them would further reduce the need of redrawing things, that are not changing, which impoves reaction times. The drawDraggedShape is a bit of a renegade here, that does not seem to work well with the layers (at least in the SWC). It could be combined into the drawing process of overlays, but with a cost, that I'm not entirely happy with...

One thing that bothers me is that in the GL-drawer transparency has been made an entirely on-off feature. That actually limits the options quite a bit. I am NOT suggesting painting the _scene_ with 'real' transparency _(I have not seen a faultless implementation yet even in my daily tools)_, but allowing alpha channel blending for some functions would help with a lot of things. -- It's just about adding the informative value of the graphics without disturbing the view too much.

One thing, I wonder is, if the layers can exchange information about the z-distance? In the SW-drawer, that can happen by reading the z-buffer, but in GL that information is a litte deeper than that... (And now, I'm beginning to the the cachedSnapShot as an option again.)

peastman:
> This is about finding the balance between memory usage and performance.

I think it's more a matter of premature optimization. :)  Going beyond two layers isn't likely to have any meaningful impact on performance.  The background almost never changes, except when the scene also changes, so there's no benefit to being able to update it separately.  The overlay and mouse feedback are both super quick to draw.  You could probably hit 1000 fps for either one of them.  So avoiding a redraw of one won't make any difference.

> The GLCanvasDrawer uses UNSIGNED_BYTE at least in some of the operations. I think this is a 7-bit/byte format.

No, bytes have 8 bits. :)  Anyway, I just searched for UNSIGNED_BYTE in GLCanvasDrawer, and the only place it's used is in drawing images, to specify the image format.

Sailsman63:
Trying to trade z-data back and forth between layers would get complicated. Realtime 3d engines, such as for games, spend a lot of time and optimization on this, and we shouldn't be worrying about the types or artifacts that they are trying to avoid.

I'd say that grids, etc. should be rendered on the back layer with the scene. Everything else in the front layer, with the following z-order, back to front:

- XYZ orientation marker
- Mouse handles, like the 3dManipulator
- Nav. cues, dragged selection lassos, etc.

If we decide to add other things to the interaction, we can re-vamp the order at that time.

peteihis:
I'm a little stuck wit this thing.The SoftwareCanwasDrawer works OK with two layers and I separated the `updateOevrlay()` into itäs own method, which wirs tells the drawer to switch to the appropriate layer and then tells the tools to draw what ever they fell like drawing. -- This happens without an event, when a full repaint is not needed.

But the GLCanvasDrawer is givng me a bit of a headace. I can add GLJPanels into an OverlayCantainer using `AWTWidget`, but only one of them works. The `addGLEventListener` does nothing and the second GLJPanel can not produce a `GL2` to draw on. --> NullPointerException --> blank screen.

> Realtime 3d engines, such as for games, spend a lot of time and optimization on this, and we shouldn't be worrying about the types or artifacts that they are trying to avoid.

Nothing of that sort. Just hoping to add some 3D-feel to wire frame objects that are drawn onto the screen. The idea would be to draw the parts, that are inside or behind solid 3D objects in "hidden line style" and the rest normally. This is a pretty common thing to do in a graphic software and I would not expect it to be very difficult. (In the case of the SWCD the tools seem to be there already...)

Sailsman63:
What is the full trace for the NPE? And when is it thrown? At startup, or some other time? I suspect some issues with shared contexts....

>Just hoping to add some 3D-feel to wireframe objects, that are drawn onto the screen. The idea would be to draw the parts, that are inside or behind solid 3D objects in "hidden line style"

This has nothing to do with sharing z-data across layers. hidden-line erasure/backface culling should be implemented in the code that draws the wireframe. You just have to draw it to the appropriate layer. If you are doing something like using a wireframe for a drag indicator, you would render it on the upper layer, because you want it visible even when dragged _behind_ other items in the scene...

peteihis:
In the consrtuctor I have:

~~~~java
  public GLCanvasDrawer(ViewerCanvas view)
  {
    this.view = view;
    GLCapabilities caps = new GLCapabilities(GLProfile.get(GLProfile.GL2));
    canvas = new OverlayContainer();
    canvas.add(sceneLayer   = new AWTWidget(scenePanel   = new GLJPanel(caps)));
    canvas.add(overlayLayer = new AWTWidget(overlayPanel = new GLJPanel(caps)));
    scenePanel.addGLEventListener(new CanvasListener());
    overlayPanel.addGLEventListener(new OverlayListener());
    canvas.layoutChildren();
    view.addEventLink(RepaintEvent.class, this, "paint");
  }
~~~~

The  `OverlayListener` is an experiment, where I made a copy of the `CanvasListener` to see if that would help, but the `init` method there is never used. I also tried to use an extra check that would create  the GL2, but that does nothing either. I realize that I have not tried to create a dedicated `GLCapabilities` for the latter layer.

The NPE the happens every time when something shoul dbe drawn on the layer, sarting with the CoordinateAxes. Here's the error trace.
[SystemOutError.txt](https://github.com/ArtOfIllusion/ArtOfIllusion/files/1405746/SystemOutError.txt)

> You just have to draw it to the appropriate layer.

That would a kind of ruin the idea of separating the layers. If I draw a temoprary wireframe on the scene layer, then I have to regenerate all of the 3D to erarse it. -- The dragShape is of course just on the top layer. Anyway this is not the very top-most thing on the priority list now.

peteihis:
On more general level, I probably don't understand too well, how the GL is supposed to be working. 

For example, as long as the GLJPanel was set the component of the ViewerCanvas `redraw`, affeceted the GLJPanel directly, without an EventLink and it was havin the the view dispatchi a RedrawEvent _after_ drawing, but now that the GLs are wrapped in the buoy, I had to write an EventLink and a `paint`-method and remove the dispathcing part....

Sailsman63:
The exception is originating in AOI code, and from the line numbers, it looks like you've modified the file. I'll have to take a look at your actual current code to track down what's happening. It sort of sounds like things are getting convoluted, and I'm not sure if they need to be.

------------------------------------------------
>That would a kind of ruin the idea of separating the layers. If I draw a temporary wireframe on the scene layer, then I have to regenerate all of the 3D to erarse it. -- The dragShape is of course just on the top layer. Anyway this is not the very top-most thing on the priority list now.

I think I need to clarify what I mean: A 'temporary' wireframe would be rendered in the overlay layer, and it should be rendered over the top of the _entire_ scene, even if some scene object is technically in front of it in z-depth.

Trying to interleave z-info is exactly the kind of stuff that game engines spend a _lot_ of code on.

peastman:
I think we need to pause, step back, and do a proper design for this.  The implementation seems to be getting ahead of what we're actually sure we want to do.  Let's answer the questions, 1) what problem are we trying to solve, and 2) what are the requirements for the solution?

The problem, as I understand it, is that we may sometimes want aspects of the image to change that are very cheap to draw, and change at times when the main rendered view doesn't.  Rerendering the whole view is wasteful and slows down the UI.

The above qualifiers are very important.  This is only relevant if the changing aspects of the image are themselves fast to draw.  If they're expensive, then the relative speedup from not rendering the scene is much smaller.  It also is only relevant if the scene doesn't need to be rendered anyway.  For example, I don't think this applies to anything drawn during mouse drags.  If you're dragging the mouse, we're probably rerendering the scene anyway.  It also is only significant if the changing aspect of the image changes frequently.  If it only changes rarely, there's little benefit to optimizing it.

So far as I know, the only thing this really applies to is cues that change as you move the mouse.  Can we make a list of all such cues?  The only ones we currently have are for the navigation modes, and I'm still not convinced we really want those at all.  As described in #64, I find them more distracting than useful.  So we need to decide what we're going to do with those.  What other types of mouse cues might we also want to add?

Once we identify them, the next question to answer is how they should be drawn.  For example, are they 3D objects that are best rendered with OpenGL, or are they 2D graphics better drawn with the Java graphics APIs?  Answering that question will dictate a lot about how we want to implement this.

peteihis:
> 1) what problem are we trying to solve

I recognize this question as the question that expensively paid R&D-consultants like to present as the key indicator to justifying any development ... And I hate the question. We can try to make things better, without necessarily seeing the current state of things as a problem. :)

Having said that, my point here was to enable fast drawing and erasing of short lived graphics without having to re-render the entire 3D-scene. _Currently_ there is not much, that uses it but if the possibility is not enabled, then features that could rely on it can not be implemented. The order of progress needs to be to _first_ build the ability or "rediness" and _then_ add the things that use it.

A few things, that would be using this are:
- Selection cues (hilight elements on mouse-over). This is a standard feature in the tools that I use.
- On-screen warnings, like "zoom too close". Especially in perspective mode this is a very real possibilily, that I have seen to happen to almos everybody some times: Zoom get's stuck and the user does not know why, and a pop-up window there would be an over-kill...
- Drag ghost graphics. This is another standard feature nowadays _(actually has been as long as I remember)_. Instead of regenerating the graphics of the entire scene you just draw the moving parts (possibly highlighted and transparent) while they are being dragged. At mouse relese it is time to redraw the entire scene.
- All the above in four views.

I can think a few more, but those are the very basic features that I'm missing.

> As described in #64, 

I'll be happy to rethink those cues. They really are not nearly as nice as I had hoped for but I still think _(and I base my opionion on my past experiences on similar modes)_ something is needed to make sure that the user knows, that the behaviour of the view has changed from the usual. Just seeing the selection on the menu bar is not enough. I have a few ideas, but let's discuss those in the other thread.

The above would make a start, but on top of those I'd like to add the ability to handle transparent graphics on the overlays layer (or what ever it would be called). As the amount of graphic information increases, then keeping it somewhat subtle is important and transprency is a very good way of doing that. An example from my world: I find it disturbing, when a drag ghost is drawn in pure white with about 80% opacy. I'd like to use the specific color, (let it be white), but I also need to see scene behind the chost while I drag. -- I could not edit the transparency level, so I had to turn the painted highlight off and use just the wireframe highlight, which then begins to look messy, when the object has a lot of geometry in it.

...And to add on the pile, as the next thing would come the depth checking for the 3D-wireframes, that belong to tools, but that'd be the last thing on this list. A point of modernization though.

Sailsman63:
>I recognize this question as the question that expensively paid R&D-consultants like to present as the key indicator to justifying any development ...

Say rather, anyone who needs to prioritize the allocation of limited resources (such as our own free time) to a project.

>The order of progress needs to be to first build the ability or "rediness" and then add the things that use it.

This does sometime morph over into 'premature optimization.' Question: do we intend to add any of the 'things that use it' before the next release? If not, this project gets dropped to a much lower priority.

Regarding your list:

1. I like. selection in AOI can be a bit frustrating, especially if there is a lot of geometry close together.
2. I'd rather fix the zoom. This should not happen.
3. Another good one. We already get lag from dragging objects in very complex scenes.
4. Sort of goes without saying.

>depth checking for the 3D-wireframes, that belong to tools,

As I've said before, very complex, and could lose most of the benefits of splitting rendering panes in the first place. Tool wireframes belong on top. Any other sorting gets messy.

peteihis:
> allocation of limited resources (such as our own free time)

OOT, but ... Basically my  point was, that I would not like to spend my time under the shadow of negativity cast by problem, that I don't even see as a problem but rather a chance of improvement. -- Regardless of my time being paid or free.

> Question: do we intend to add any of the 'things that use it' before the next release? 

I was hoping to add selection cues to the 3D-manipulator. .... I can suppress it or limit to the active view though. I do recognize, that there are things with higher priority.

> I'd rather fix the zoom. This should not happen.

That was the firt example coming to my mind but there might be others as well. 

In the case of zoom ... There really is nothing to fix, the zoom is working as it should, but the user may "panic", when he does not recognize the situation. Maybe the best fix would be that the software suggests an action. _(For example the idea of the software actively scanning for a new rotation center is one that would prevent this, but I have recently started to use a tool that does that and I'm desperately trying to find a way to switch the scaning off...  )_

peteihis:
BTW, I wouldn't mind scrapping this PR and starting over but I felt that I had to get the thing started. Basically the first commit is capable of what I'd like to see in the first phase of implementaion and it has the functional priciple that would seem logical to me: Call `updateView` to update everything and `updateOverlay` to erase and draw the lighter graphics.

Sailsman63:
>the zoom is working as it should, but the user may "panic", when he does not recognize the situation. Maybe the best fix would be, that the software suggests an action.

I disagree. If the zoom gets stuck, that is a _bug._ If the code can recognize the problem, it can get itself back out, but would be better not to have to add the extra code in the first place.

 What is the simplest situation that can reproduce the zoom getting stuck? Perhaps we just need to define the expected behavior a little more precisely.

peteihis:
Sorry for the long silence, my life seems to have turned into ground-hog-day lately... (though this one would be an autumn version of it)

> If the zoom gets stuck, that is a bug.

Actually it does not, but let's have a different discussion about that. -- I'll try to do a little demo to explain it...

> are they 3D objects that are best rendered with OpenGL, or are they 2D graphics better drawn with the Java graphics APIs? 

Mostly, what I was originally missing were lines, text and an occasional on-screen icon -- think of the 3D-manipulator.  By it's nature it is a 3D-object, but it is drawn entirely by 2D-methods. That and all the currently existing overlays could be handled perfectlty well by the current SW-Drawer but there are things on the list that would eventually be drawn more efficiently by GL. 

Based on that, the approach could be, that as the first step drawing overlays would be directed to the SW-Drawer and the GL-would be used as the "HW-accelerator" for rendering the 3D-scene on the back. The AWT-methods have the advance that hey can handle RGB colors out of the box. There are cases where that would make the overlays less disturbing. The rotation cue of the manipulator to mention one.

Then as futher steps, for examle to draw dragging ghosts, it migh be worth to have the GL do that, but what that takes is another story. Might after all require another layer and possibly the capability of handling the alpha channel.

Sailsman63:
>my life seems to have turned into Groundhog Day lately... 

I hear you. I'll probably be working 50+ hour weeks through the end of December (Seasonal Business) So it may take a while before I can review recent contributions properly.



>    If the zoom gets stuck, that is a bug.

>Actually it does not, but let's have a different discussion about that. -- I'll try to do a little demo to explain it...

Yes, that should be an entirely separate discussion, not in this thread.

peastman:
> there are things on the list that would eventually be drawn more efficiently by GL.

Efficiency probably isn't a big concern here.  First, because we're only talking about things that are really fast to draw (otherwise there's less benefit from separating them out, as noted above), and second because the Java2D APIs are actually implemented with OpenGL anyway.

So it sounds like we just want a CustomWidget for the overlay, and we'll draw into it using the standard methods of Graphics2D?

peteihis:
>> there are things on the list that would eventually be drawn more efficiently by GL.

>  Java2D APIs are actually implemented with OpenGL anyway.

Well, I meant the parts that need 3D but let's worry about those later.

> So it sounds like we just want a CustomWidget for the overlay, and we'll draw into it using the standard methods of Graphics2D?

That woud pretty much be, what I had originally sketched in my mind, but how about the guidelines how to have that happen?

My original idea, some time last winter was, that there would be an "OverlayDrawer" that would be extended from the SWCanvasDrawer. It shoud have methods for drawing arrows, texts, predefined shapes with parametric control, etc. and accept RGBA-colors, which AWT handles by default.

Then I began to wonder if it would be better just to add those to the SWCanvasDrawer. For clarity probably not. 

_(Using or Extending the SWCD, would take care of at least one thing. AWT and GL interpre the same (2D) graphic information with a slight difference, that probably not everybody pays attention to but I, for one, find sometimes a bit disturbing....)_

And then: I suppose the 2-layes system is the way to go? And the bottom layer (='media' to draw the 3D on) is defined/created, when the main canvas drawer is selected... And the overlay layer is deifend the same way the SWCD does it now?

peastman:
I don't think it should have anything to do with SoftwareCanvasDrawer.  It can just be a completely ordinary widget.  You can draw whatever you want using the standard Graphics2D methods.  That's already a much more powerful API.

PR #68, opened 2017-10-05
Set window modified on import file

makiam:
Once import performed and new window opened it marked as unmodified. If after that user opens other window or create other new window, the imported scene is silently closed.
I add code to force set window mofified on import

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
+++ b/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
@@ -269,17 +269,17 @@ public class ArtOfIllusion
     newWindow(theScene);
   }
 
-  /** Create a new window for editing the specified scene. */
-  public static void newWindow(final Scene theScene)
+  /** Create a new window for editing the specified scene and set window modified. */
+  public static void newWindow(final Scene scene, final boolean modified)
   {
     // New windows should always be created on the event thread.
-
+    
     numNewWindows++;
     SwingUtilities.invokeLater(new Runnable() {
       @Override
       public void run()
       {
-        LayoutWindow fr = new LayoutWindow(theScene);
+        LayoutWindow fr = new LayoutWindow(scene);        
         windows.add(fr);
         
         for (Plugin plugin: PluginRegistry.getPlugins(Plugin.class))
@@ -296,6 +296,7 @@ public class ArtOfIllusion
         }
         fr.setVisible(true);
         fr.arrangeDockableWidgets();
+        if(modified) fr.setModified();
 
         // If the user opens a file immediately after running the program, close the empty
         // scene window.
@@ -308,7 +309,14 @@ public class ArtOfIllusion
               closeWindow(win);
           }
       }
-    });
+    });    
+  }
+  
+  /** Create a new window for editing the specified scene. */
+  public static void newWindow(final Scene scene)
+  {
+    boolean modified = false;
+    newWindow(scene, modified);
   }
 
   /** Add a window to the list of open windows. */
--- a/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
+++ b/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
@@ -269,8 +269,8 @@ public class ArtOfIllusion
     newWindow(theScene);
   }
 
-  /** Create a new window for editing the specified scene and set window modified. */
-  public static void newWindow(final Scene scene, final boolean modified)
+  /** Create a new window for editing the specified scene. */
+  public static void newWindow(final Scene scene)
   {
     // New windows should always be created on the event thread.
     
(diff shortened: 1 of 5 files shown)


peteihis:
Interesting that before this the `isModified()` method returns false but the only saving option for the new window is "Save As". I would understand, that with an empty scene, but in this case there is something in it. :)

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
```diff
@@ -269,17 +269,17 @@ public static void newWindow()
     newWindow(theScene);
   }
 
-  /** Create a new window for editing the specified scene. */
-  public static void newWindow(final Scene theScene)
+  /** Create a new window for editing the specified scene and set window modified. */
+  public static void newWindow(final Scene scene, final boolean modified)
```
Sailsman63: Please do not make changes to the public API of major classes without discussing with the core team.

Sailsman63:
I see your original issue. Not sure that we want an additional method signature added here, though.

This gets into how AOI knows whether its safe to close a window, hence the check with `isModified`. The original issue may, in fact, be intended behavior. After all, you've simply opened the data contained in the external file, and not changed it at all.

Perhaps `importFile(BFrame)` in the importer should set a name for the new scene, based on the original file name. You are only losing the data because the imported scene is untitled.

peteihis:
I hope you don't mind me asking, but why would you name an untidled file "Untitled". AoI calls it that automatically already and. _(BTW "Untitled is not even translated to the local language, but that is a different issue.)_ :) 

What would help, woud be to get the _name of the imported file_ and use that as the name of the scene. That is pretty much the common practice in the rest of the world: When files are imported/exported with format changes, the name of the file/scene/content is retained, when ever possible.

peteihis:
Better, but now it reads the file name with the extension, so next time you save it as .aoi, the name that it suggest is something like "SceneName.obj.aoi". 

I'd suggest to use `file.getName().substring(0, file.getName().lastIndexOf('.'))`, which is used in ExternalObject and ExternalImage.

makiam:
I update code to set imported scene name from file name. This prevents close imported scene on new window open as this code checks modification flag AND sceneName null equality. However imported scene still silently closes when user call to exit App. This code calls confirmClose() from LayoutWindow and as window modification flag is not true the window is closes....

Moreover IMO the modification flag should be the Scene attribute instead of window...

peteihis:
> Moreover IMO the modification flag should be the Scene attribute instead of window...

I was wondering the same thing. Sometimes it would be simpler to to get the information in, if it were in the scene, as in some cases there is no connection from the tools to the window, that is hosting the scene. (For example _somewhere_ in handling of procedural textures, the arrangement is that just opening a dialog sets `isModified(true)` and it does not matter if you modified anything or not.) 

On the other hand, the modified flag is not the kind of information that would be saved with the scene, -- In other words, it is not an actual scene property,  so I a kind of understand the logic in keeping it in the LayoutWindow.

Translators/src/artofillusion/translators/OBJImporter.java
```diff
@@ -32,23 +34,19 @@
...
     // Create a scene to add objects to.
 
     Scene theScene = new Scene();
     CoordinateSystem coords = new CoordinateSystem(new Vec3(0.0, 0.0, Camera.DEFAULT_DISTANCE_TO_SCREEN), new Vec3(0.0, 0.0, -1.0), Vec3.vy());
     ObjectInfo info = new ObjectInfo(new SceneCamera(), coords, "Camera 1");
-    info.addTrack(new PositionTrack(info), 0);
-    info.addTrack(new RotationTrack(info), 1);
-    theScene.addObject(info, null);
+
+    theScene.addObject(info, (UndoRecord)null);
```
Sailsman63: Why the Cast to `UndoRecord`?

ArtOfIllusion/src/artofillusion/util/FileUtils.java
```diff
@@ -0,0 +1,34 @@
...
+
+package artofillusion.util;
+
+import java.io.File;
+
+/**
+ *
+ * @author MaksK
+ */
+public final class FileUtils
```
Sailsman63: Do you have plans for other file utils?

I'd rather not create a new class for a single method. I also think that unrelated refactors, such as the changes to `POVExporter` do __not__ belong in this PR.

Translators/src/artofillusion/translators/OBJImporter.java
makiam: Just a copy-paste from ArtOfIllusion.newWindow()

makiam:
I see couple of points to replace code with getNameNoExtension():
ScriptedObjectEditorWindow class line 142 & setScriptNameFromFile() method
ExecuteScriptWindow class line 139 & setScriptNameFromFile() method
As for me ArtOfIllusion class method loadFile() is a good candidate to move to FileUtil too

peteihis:
> `return (pointPos == -1) ? fileName : fileName.substring(0, pointPos);`

I'll leave it to others to decide about the class and the method, but at least that this is a good check. The default value for the `lastIndexOf()` is `-1`and if the `fileName` doesn't not have an extension, the `subString()` causes `arrayOutOfBounds`.

PR #69, opened 2017-10-22
Move some duplicated code to separate class

makiam:


ArtOfIllusion/src/artofillusion/texture/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,163 @@
+/* Copyright (C) 2017 by Maksim Khramov
+
+   This program is free software; you can redistribute it and/or modify it under the
+   terms of the GNU General Public License as published by the Free Software
+   Foundation; either version 2 of the License, or (at your option) any later version.
+   This program is distributed in the hope that it will be useful, but WITHOUT ANY 
+   WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A 
+   PARTICULAR PURPOSE.  See the GNU General Public License for more details. */
+
+package artofillusion.texture;
```
Sailsman63: These common files are not specific to textures, therefore this class should not be in `artofillusion.texture`. `artofillusion.procedural` would probably be a better home.

ArtOfIllusion/src/artofillusion/texture/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,163 @@
...
+import artofillusion.TextureParameter;
+import artofillusion.image.ImageMap;
+import artofillusion.math.RGBColor;
+import artofillusion.procedural.IOPort;
+import artofillusion.procedural.ImageModule;
+import artofillusion.procedural.Module;
+import artofillusion.procedural.OutputModule;
+import artofillusion.procedural.ParameterModule;
+import artofillusion.procedural.Procedure;
+import artofillusion.procedural.ProcedureEditor;
```
Sailsman63: Please collapse imports when using most of a package! `import artofillusion.procedural.*`

ArtOfIllusion/src/artofillusion/texture/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,163 @@
...
+{
+  private ProceduralTextureCommon()
+  {    
+  }
+  
+  /**
+   * Create a Procedure object for texture.
+   */
+
+  public static Procedure createProcedure()
```
Sailsman63: This method is the odd one out, as it is specific to _Texture_ procedures. Perhaps call it `createTextureProceedure()`? Perhaps it does not really belong here, despite the fact that two types of texture use the same code.

ArtOfIllusion/src/artofillusion/texture/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+    return params;
+    
+  }
+  
+  /** Determine whether given procedure texture has a non-zero value anywhere for a particular component.
+   *  @param proc the Procedure to check 
+      @param component    the texture component to check for (one of the Texture *_COMPONENT constants)
+  */  
+  public static boolean hasTextureComponent(Procedure proc, int component)
```
Sailsman63: Also perhaps not belonging here, as it only works on _texture_ procedures?

ArtOfIllusion/src/artofillusion/texture/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,163 @@
...
+  
+  public static boolean procedureUsesImage(Procedure proc, ImageMap image)
+  {
+    for (Module module: proc.getModules())
+      if (module instanceof ImageModule && ((ImageModule) module).getMap() == image)
+          return true;
+    return false;
+  }
+  
+  public static MaterialPreviewer getPreview(ProcedureEditor editor, Texture texture)
```
Sailsman63: ditto...

ArtOfIllusion/src/artofillusion/texture/ProceduralTexture3D.java
```diff
@@ -235,6 +185,14 @@ public Texture duplicate()
     tex.antialiasing = antialiasing;
     return tex;
   }
+  
+  /** Get the list of parameters for this texture. */
+
+  @Override
+  public TextureParameter[] getParameters()
```
Sailsman63: You've moved this method from above `duplicate()` to below. Any particular reason?

ArtOfIllusion/src/artofillusion/texture/ProceduralTexture2D.java
```diff
@@ -385,22 +294,22 @@ public void disposePreview(Object preview)
     UIUtilities.findWindow((MaterialPreviewer) preview).dispose();
   }
 
-  /** Determine whether the procedure may contain Parameter modules. */
+  /** Determine whether the procedure may contain View Angle modules. */
 
   @Override
-  public boolean allowParameters()
+  public boolean allowViewAngle()
```
Sailsman63: You've also switched the positions of `allowViewAngle()` and `allowParameters()`. This makes the diff harder to read and understand. Again, is there a reason for this?

ArtOfIllusion/src/artofillusion/material/ProceduralMaterial3D.java
```diff
@@ -361,8 +324,8 @@ public void acceptEdits(ProcedureEditor editor)
   {
     initThreadLocal();
     int i = editor.getScene().indexOf(this);
-    if (i > -1)
-      editor.getScene().changeMaterial(i);
+    if(i == -1) return;
```
Sailsman63: This is not a good idea. I know that you do not like the single-statement stye:
```
if(cond)
   doThis();
```
But the inverted condition with an early return is worse. Please __do not__ do this. The style in the block above is correct.

As a second point, style changes outside of the methods that you are editing __Should Not__ be done, as the make the diff harder to understand. If something is seriously difficult to read/needs a re-format, please open a separate issue to discuss.

ArtOfIllusion/src/artofillusion/texture/ProceduralTexture2D.java
```diff
@@ -416,8 +325,8 @@ public void acceptEdits(ProcedureEditor editor)
   {
     initThreadLocal();
     int i = editor.getScene().indexOf(this);
-    if (i > -1)
-      editor.getScene().changeTexture(i);
+    if (i == -1) return;
```
Sailsman63: Same.

ArtOfIllusion/src/artofillusion/texture/ProceduralTexture3D.java
```diff
@@ -417,8 +325,8 @@ public void acceptEdits(ProcedureEditor editor)
   {
     initThreadLocal();
     int i = editor.getScene().indexOf(this);
-    if (i > -1)
-      editor.getScene().changeTexture(i);
+    if (i == -1) return;
```
Sailsman63: Same

ArtOfIllusion/src/artofillusion/texture/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,163 @@
...
+  }
+  
+  public static TextureParameter[] getTextureParameters(Object texture, Procedure proc)
+  {
+    Module modules[] = proc.getModules();
+    int count = 0;
+
+    for (Module module1 : modules)
+    {
+      if (module1 instanceof ParameterModule)
```
Sailsman63: Perfect spot for the indented if style.

ArtOfIllusion/src/artofillusion/texture/ProceduralTexture3D.java
makiam: I do this to made this and ProceduralTexture2D methods order the same to be easily compared. Same for allowParameters() and allowViewAngle()...

Sailsman63: Don't make changes like that. This is sort of like reformatting - it does not belong in a PR that is focused on something else.

ArtOfIllusion/src/artofillusion/procedural/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,151 @@
...
+{
+  private ProceduralTextureCommon()
+  {    
+  }
+  
+  /**
+   * Create a Procedure object for texture.
+   */
+
+  public static Procedure createTextureProcedure()
```
Sailsman63: It does make sense to associate this code with the texturing system, rather than in the procedure system. I'm just not sure that reducing the amount of duplicated code is enough reason in this instance to have an entirely separate helper class.

ArtOfIllusion/src/artofillusion/procedural/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,151 @@
...
+      }
+    return params;
+    
+  }
+  
+  /** Determine whether given procedure texture has a non-zero value anywhere for a particular component.
+   *  @param proc the Procedure to check 
+      @param component    the texture component to check for (one of the Texture *_COMPONENT constants)
+  */  
+  public static boolean hasTextureComponent(Procedure proc, int component)
```
Sailsman63: Same here. Putting these in a separate class adds some friction to reading and _understanding_ the Procedural Texture code. Not sure if the reduced line count is enough to warrant that.

ArtOfIllusion/src/artofillusion/procedural/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,151 @@
...
+  
+  public static boolean procedureUsesImage(Procedure proc, ImageMap image)
+  {
+    for (Module module: proc.getModules())
+      if (module instanceof ImageModule && ((ImageModule) module).getMap() == image)
+          return true;
+    return false;
+  }
+  
+  public static MaterialPreviewer getPreview(ProcedureEditor editor, Texture texture)
```
Sailsman63: This one I'm not sure of. It may, in fact, work best as a Factory method inside `artofillusion.MaterialPreview`. Something like `getGenericPreviewDialog()` but with a better name. It certainly does not belong in the same class as the others, though.

ArtOfIllusion/src/artofillusion/procedural/ProceduralTextureCommon.java
```diff
@@ -0,0 +1,151 @@
...
+  {
+    final BDialog dlg = new BDialog(editor.getParentFrame(), "Preview", false);
```
Sailsman63: BTW, You've got a bug here. Since you never pass a material, the preview never displays anything when you are editing a `ProceduralMaterial`

makiam: MaterialPreviewer can display Texture and/or Material. So for Procedural textures material is null.
See    private void initObject(Texture tex, Material mat, ObjectInfo objInfo) at MaterialPreviewer line 79

ArtOfIllusion/src/artofillusion/procedural/Procedure.java
```diff
@@ -21,6 +25,77 @@
 
 public class Procedure
 {
+
+  public static TextureParameter[] getTextureParameters(Procedure proc, Object texture)
```
Sailsman63: Looking better. We're getting there. This method should not be static. Rather: `public TextureParameter[] getTextureParameters()` and internally replace `proc` with `this` (actually, only used on line 31, which can be `Module[] modules = getModules();`)

ArtOfIllusion/src/artofillusion/procedural/Procedure.java
```diff
@@ -21,6 +25,77 @@
...
+        return output[3].inputConnected(0);
+      case Texture.BUMP_COMPONENT:
+        return output[9].inputConnected(0);
+      case Texture.DISPLACEMENT_COMPONENT:
+        return output[10].inputConnected(0);
+    }
+    return false;
+  }
+
+  public static boolean procedureUsesImage(Procedure proc, ImageMap image)
```
Sailsman63: Same thing about static vs. instance.

ArtOfIllusion/src/artofillusion/procedural/Procedure.java
```diff
@@ -21,6 +25,77 @@
...
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Create a Procedure object for texture.
+   */
+  public static Procedure createTextureProcedure()
```
Sailsman63: Still not sure about the correct location for this one.

ArtOfIllusion/src/artofillusion/procedural/Procedure.java
```diff
@@ -21,6 +25,77 @@
...
+      }
+    }
+    return params;
+  }
+
+  /** Determine whether given procedure texture has a non-zero value anywhere for a particular component.
+   *  @param proc the Procedure to check
+  @param component    the texture component to check for (one of the Texture *_COMPONENT constants)
+   */
+  public static boolean hasTextureComponent(Procedure proc, int component)
```
Sailsman63: Or this one.

ArtOfIllusion/src/artofillusion/procedural/Procedure.java
```diff
@@ -274,4 +278,74 @@ public void readFromStream(DataInputStream in, Scene theScene) throws IOExceptio
...
+        params[count] = ((ParameterModule) mod).getParameter(texture);
+        ((ParameterModule) mod).setIndex(count++);
+      }
+    return params;
+  }
+
+  /** Determine whether given procedure texture has a non-zero value anywhere for a particular component.
+      @param component    the texture component to check for (one of the Texture *_COMPONENT constants)
+   */
+  public boolean hasTextureComponent(int component)
```
makiam: This method now look not good for me as it implicitly depends of order and count of procedure output modules...

Sailsman63: It is specific to `Texture` procedures. As long as the procedure is a `Texture` procedure, it will work fine. That is why I think that this method and the `createTextureProcedure()` method do not belong in this class. I'm not sure where to put them though. For the time being, I would remove them entirely, and leave the original code in the Texture classes.

PR #72, opened 2017-10-29
Made artofillusion.procedural package accessible for scripting engine

makiam:
Made artofillusion.procedural package accessible for scripting engine


peteihis:
I have ben wondering why the animation package is not visible to scripting? There probably is a reason, but I'd like hear, what the thinking behind it is?>

Sailsman63:
Essentially, `artofillusion.procedure` contains _only_ procedural modules. These modules are intended to be manipulated through the graphical interface, not code. Trying to access them in a script makes no sense. and could lead to some very strange behavior.

If you need to create a texture in a script, you are better off writing the entire texture in code.

makiam:
Classes from artofillusion.procedure are still accessible with FQN. Adding this package to imports gives only a bit comfortable way to access classes.

Sailsman63:
That's true, but part of the point is that the extra friction should make you stop and ask if what you are doing makes sense.

PR #73, opened 2017-11-07
Fix to blank view on mouse drag in SceneViewer

peteihis:
This fixes the problem that Konrad reported on SourceForge. 

The bug was, that the views could go blank, when you pressd and dragged on a view, that was set to perspective viewing. The cause was, that an object that was in the same location as then view camera could be selected at mouse press and then moved by drag, which caused an uncaught error in `findDragVector()` of `Camera`.

It is still possible to select a "wrong object" using the object list, but this fix deselects those when the illegal move is attempted.

I have not tested object editors for a similar case, but I believe the same phenomenon to be present there as well, though less likely to happen.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/Camera.java
+++ b/ArtOfIllusion/src/artofillusion/Camera.java
@@ -338,7 +338,7 @@ public class Camera implements Cloneable
   
   public Vec3 findDragVector(Vec3 p, double dx, double dy)
   {
-    Vec3 v1, v2;
+    Vec3 v1, v2, v;
     Vec2 v3, v4, p1;
     double a, b;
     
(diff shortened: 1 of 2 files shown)


peteihis:
In an object editor the exact circumstances when the bug emerges seem to be less likely to happen, but when the wrong combination happens, the effects are even worse than on the scene view: All the views turned blank so it was not possible to fix the sitution and after returning to SceneViewer it got blank too.

This fix takes care of that. It has the `findDragVector` return a `new Vec3()` if it has a "NaN-vector" as a result. This actually is the correct answer: When the point gets closer to camera the drag vector gets shorter too, hence a 0.0 for distance should the result 0.0 as drag distance as well.

Sailsman63:
I've been playing with this one a bit, and have not been able to duplicate? Was this for standard left-click drags, or something else?

peteihis:
Strangely I can now repeat it only partially. 

On the latest head, the problem is still there in the Scene Views. To me the problem occurs, when you set the top right or lower left view to Camera 1. Just normal left button clik & drag knocks out the coordinates of Light 1 and then Camera 1 _(and sometimes first any other objects that may be at [0, 0, 20])_ if everything is in their default positions.

On editor windows I could not make it happen. On the 3.0.3 some kind of a milder version occurs sometimes, but nothing as disasterous as I had happening at the time I wrote the fix and I can not reproduce the phenomenon with 100% certainty either ... I have no idea, what I did differently earlier.

As I recall the sympthoms that Konrad described were also slightly different from, what I could produce.

Sailsman63:
Okay, Duped. For me, it's all but the lower right pane.... Weird. I'll take a closer look.

Sailsman63:
Just pushed a cleaned version, with no whitespace changes. Will continue testing.

Sailsman63:
Looks Good!

PR #74, opened 2017-11-28
Navigation fixes

peteihis:
A PR for fixes to the view manipulation.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -1247,6 +1247,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
   public void setTime(double time)
   {
     theScene.setTime(time);
+    theScore.setTime(time);
     theScore.repaint();
     itemTree.repaint();
     for (SceneViewer view : theView)
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -563,7 +563,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
     
     toolsMenu = Translate.menu("tools");
     menubar.add(toolsMenu);
-	
+    
     for (ModellingTool tool: modellingTools)
     {
       BMenuItem item = new BMenuItem(tool.getName());
@@ -1178,8 +1178,8 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
 
   /** Add a new object to the scene.  If undo is not null,
       appropriate commands will be added to it to undo this operation. <P>
-	  
-	  NOTE! This method is only used by 'UndoRecord'. Using it in any other context is not safe. */
+      
+      NOTE! This method is only used by 'UndoRecord'. Using it in any other context is not safe. */
 
   public void addObject(ObjectInfo info, int index, UndoRecord undo)
   {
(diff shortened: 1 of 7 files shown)


peteihis:
Added a propper undo to the scroll wheel moves of the SceneCamera. Should have done the tab-character removal first, but better late than never I guess?

peteihis:
Scrollwheel travel checks if orientation changes during scroll. If the orientation is set to "Other" it does not check if the orientation would match a preset. The check could be added, but a (re)match would be rather an unlikely case.

peteihis:
ViewAnimation and a navigation change will now update ObjectProperties if an affected object (a camera) is selected.

peteihis:
Ooh rats! .. I just discovered a bug in this branch... I have no idea why, but the timemarker on the score does not move during preview. I'll be back...

peteihis:
And found it. Fix added to the previous push.

Sailsman63:
Thanks for taking a look. I'll have to do some testing, etc. when I get a chance.

As a note, though, please __do not__ do tab removal when working on a file. This adds a lot of noise to the pr/commit history. Instead, make a note of files that seem to have a lot of tabs. In the future, we should go through and fix tab/whitespace issue _in a single, separate PR_.

peteihis:
> do not do tab removal

OK...

>I'll have to do some testing, etc. when I get a chance.

There are still things missing from this PR. It seems that I'll have to study a few things a bit deeper. Trying to fix some rather minir things seem to produce side effects, that I was not anticipating and I need to map out what is happening.

peteihis:
This fixes the behavior of the navigation mode selection in the case that an animation track tilts the camera.

Some tab replacement happend, but it could not really be helped this time.

Sailsman63:
Okay, Finally got a chance to clean up the whitespace issues. Take a look at #76. See how much easier it is to see the _important_ changes?

peteihis:
Sorry for the yet another long silence -- Hoping that I can get back on track soonish...

So,  I'll rebase my branch on yours now?

Sailsman63:
>Sorry for the yet another long silence

That's okay. I've been mostly distracted by other things as well.

>So, I'll rebase my branch on yours now?

No. That won't do what  you think. I've mostly kept the branches open just in case @peastman noticed anything else. I'll review my branch again later today, (It has _all_ of your changes, with you credited as the author...) and if everything looks good, I'll merge it into Master.

peteihis:
> I'll merge it into Master.

That's, would have been my next suggestion :D ... I have a few more things I was testing, but what is already is should work as it is.

There are still some things "a little bit" off with the handling of light cameras, that I'd like the get under control but, when I was testing it, solving one issue seemed to arise others ... I might leave that for later. _(I think I'll end up making a change to the priciple the parallel/perspective modes are handled internally. It should make things easier afterwards, but there seems to a lot of places that would be affected. -- The fast forward way would be to go `if (...)`, but that'd lead to special cases over special cases, which is not a small project either and adds unnecessary complexity --> Probably has to wait past 3.1)_

Other than that I think, I'll just try to cut out some features, that have been found confusing and focus on getting the 3.1 out. -- Then plan a modernization round with the layers and the `mouseMoved`-functions as those two are pretty closely related.

Sailsman63:
>when I was testing it, solving one issue seemed to arise others

Sounds like some of the code does things that you don't quite understand... or interacts with other code in complex ways. "You keep changing that code. I do not think it does what you think it does"

>I'll just try to cut out some features, that have been found confusing and focus on getting the 3.1 out.

We need to decide _what_ we need to do. Mostly to make the control modes easier to discover. Please discuss that over on issue #64.

Issue #50, opened 2017-08-18
Standardize logging style


Sailsman63:
#7 pointed out that AOI does not really have a logging system. `System.out.println()` calls scattered just about anywhere.

A simple logging framework would seem to be in order, even if just a home-rolled class that handles driecting output to file or otherwise, and diferentiating between emergency messages and "Include the Kitchen Sink" debug level messages. Suggestions on exactly what to use are welcome.

makiam:
Minimal code to setup java standard logger:
logging.properties file at sources root:
```
handlers= java.util.logging.FileHandler,java.util.logging.ConsoleHandler
java.util.logging.FileHandler.pattern = %h/.artofillusion/ArtOfIllusion.log
java.util.logging.FileHandler.limit = 1000000
java.util.logging.FileHandler.count = 5
java.util.logging.FileHandler.formatter = java.util.logging.SimpleFormatter
```
this writes log to console and to log file under user's home
and some code at application startup:
```
            InputStream is = CJLTestApp.class.getResourceAsStream("/logging.properties");
            LogManager.getLogManager().readConfiguration(is);
```

peteihis:
It came to my mind, that if there is a logging system, could there be a log/console-window that would in (somewhat) real time show what is happening. 

Even better if, in the graphic presentation, there would be some kind of a nesting system to view certain things,  like for example a `NullPointerException` colud show just a headline and the part of the program, where it came from and, then by "clicking the +"  (or something like that) it would show, the longer trace...

That may be in some ways a bit of an anchient idea, but just a thought :) And maybe not so useful in normal use but as I run a compiled program, there naturally is no console available and the `System.out`, `printStackTrace` and so on are recorded in a text file, which I have to keep reopening to see, where I had a problem...

Sailsman63:
@peteihis Should be doable as a plugin, if we have the logging system set up first. The Collapse/expand could be a little ticklish, but not impossible. The log plugin sort of does this already for serious exceptions, and AOI does little to help it.

@makiam We can also configure the log system through code/preferences, and provide an in-app interface for changing them.

Sailsman63:
I've been poking around a bit looking into other logging options, and I'm sort of becoming enamored with [SLF4J](https://www.slf4j.org/).

[...]

PR #75, opened 2017-12-31
Update OSSpecofic plugin code and build script not to use reflection calls

makiam:
Use orangeextensions lib at compile time to allow cross compile mac specific code

### DIFF
--- a/OSSpecific.xml
+++ b/OSSpecific.xml
@@ -3,9 +3,10 @@
 <project name="OSSpecific" default="dist" basedir=".">
 
   <!-- set global properties for this build -->
-  <property name="src" value="OSSpecific/src" />
-  <property name="build" value="OSSpecific/build" />
-  <property name="docs" value="OSSpecific/docs" />
+  <property name="src" value="${ant.project.name}/src" />
+  <property name="build" value="${ant.project.name}/build" />
+  <property name="docs" value="${ant.project.name}/docs" />
+  <property name="lib" value="${ant.project.name}/lib" />
   <property name="dist" value="Live_Application" />
   <property name="assemble" value="${dist}/Plugins" />
   <property name="aoijar" value="${dist}/ArtOfIllusion.jar" />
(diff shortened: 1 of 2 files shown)


OSSpecific.xml
```diff
@@ -3,9 +3,10 @@
 <project name="OSSpecific" default="dist" basedir=".">
 
   <!-- set global properties for this build -->
-  <property name="src" value="OSSpecific/src" />
-  <property name="build" value="OSSpecific/build" />
-  <property name="docs" value="OSSpecific/docs" />
+  <property name="src" value="${ant.project.name}/src" />
```
Sailsman63: These changes do not belong in this PR. If you would like to make some style changes to the build scripts as a whole, please raise an issue for discussion. This has nothing to do with removing the reflection code from `OSSpecific`.

OSSpecific/src/artofillusion/osspecific/MacOSPlugin.java
```diff
@@ -15,16 +15,21 @@ Changes copyright (C) 2017 by Maksim Khramov
 import artofillusion.ui.*;
 import buoy.event.*;
 import buoy.widget.*;
+import com.apple.eawt.Application;
+import com.apple.eawt.ApplicationEvent;
 import java.awt.*;
 import java.io.*;
-import java.lang.reflect.*;
+import java.util.logging.Level;
```
Sailsman63: We have not decided on a logging strategy. (#50)

Please refrain from adding logging code that is specific to a certain logger configuration, especially as the core concern here has nothing to do with logging.

Sailsman63:
@peastman: When you get a chance, can you test this branch? IIUC, newer OSX systems that have Oracle Java 7+, but not AppleJava, will not have the extension classes available. I would expect, though, that this would mean that the reflective version of the code would not work either...

Sorry, don't have a mac to test on.

peastman:
Like @Sailsman63 I'm confused about what this change is for.  What problem is it trying to solve?

makiam:
This update to more clean, easy to understand and update code. For future... Now only Quit and Preferences menu are moved to mac-specific menu. All other menu items also may be moved to it

makiam:
I compile and run this with JDK 8 update 131 on my Mac. As expected Preferences and Quit menu items moved to mac menu

Sailsman63:
>I compile and run this with JDK 8 update 131 on my Mac.

Okay, but does you Mac also have leftover bits from an AppleJava installation? What we need tested is whether it works on a Modern mac, with _no traces_ of AppleJava code.

>As expected Preferences and Quit menu items moved to mac menu.

Does this mean that they are _not_ moved to the mac menu in the current master code?

makiam:
> Okay, but does you Mac also have leftover bits from an AppleJava installation

I have fresh elCapitan 10.11.6 installation on it. There are no Apple java by default

> Does this mean that they are not moved to the mac menu in the current master code?

Means no regression in UI

Sailsman63:
@peastman Hate to bug you about this one again, but can you clarify something for me?

- Does the current `master` version of `OSSpecific` have UI issues, such as some standard menu items not appearing where a mac user would expect them?
- If so, does this change fix that?

peastman:
I spent a little while testing the master branch, and didn't find any obvious problems.  At any rate, the most obvious Mac-specific things seemed to be correct.  The menu bar is at the top of the screen, the Preferences and Quit menu items are both in the application menu, and it shows the marker when the scene is modified.

Sailsman63:
@peastman Okay, thanks. In that case, I'm dropping this one. While I sort of sympathize with the intent to reduce reflection, I don't think that its worth another dependency in this case.

(The other option for dropping the reflection code would be to compile `OSSpecific` on mac only. This would be doable through travis.)

peastman:
That sounds good to me.

PR #77, opened 2018-01-06
Update set and get app Locales. Restrict locale set to registered only.

makiam:
Once user's default locale is differs from registered and user has no stored options opening Properties window set selected language option to first registered one (It's africaanas). After that new windows are opens with this new language option.
I restrict Translate. setLocale to set only registered values. Otherwise set to US

### DIFF
--- a/ArtOfIllusion/src/artofillusion/PreferencesWindow.java
+++ b/ArtOfIllusion/src/artofillusion/PreferencesWindow.java
@@ -168,10 +168,11 @@ public class PreferencesWindow
     });
     toolChoice.setSelectedIndex(prefs.getUseCompoundMeshTool() ? 1 : 0);
     localeChoice = new BComboBox();
+    Locale defaultLocale = Locale.getDefault();
     Locale languages[] = Translate.getAvailableLocales();
     for (int i = 0; i < languages.length; i++)
     {
-      localeChoice.add(languages[i].getDisplayName(prefs.getLocale()));
+      localeChoice.add(languages[i].getDisplayName(defaultLocale));
       if (prefs.getLocale().equals(languages[i]))
         localeChoice.setSelectedIndex(i);
     }
--- a/ArtOfIllusion/src/artofillusion/PreferencesWindow.java
+++ b/ArtOfIllusion/src/artofillusion/PreferencesWindow.java
@@ -168,11 +168,11 @@ public class PreferencesWindow
     });
     toolChoice.setSelectedIndex(prefs.getUseCompoundMeshTool() ? 1 : 0);
     localeChoice = new BComboBox();
-    Locale defaultLocale = Locale.getDefault();
+    Locale currentLocale = prefs.getLocale();
     Locale languages[] = Translate.getAvailableLocales();
     for (int i = 0; i < languages.length; i++)
     {
-      localeChoice.add(languages[i].getDisplayName(defaultLocale));
+      localeChoice.add(languages[i].getDisplayName(currentLocale));
       if (prefs.getLocale().equals(languages[i]))
         localeChoice.setSelectedIndex(i);
     }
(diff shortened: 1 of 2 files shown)


Sailsman63:
> Once user's default locale is differs from registered

Please clarify.

- What do you mean by "registered?"
- What OS/version/related documentation?
- How would a user get into this state?

makiam:
- Registered means list of available locales in Translate i.e Translate.getAvailableLocales()...
- I have Windows 10 with Russian locale. java's Locale.getDefault() returns ru_RU for me
When user runs app first time or have no preferences file bu some reason and opens Preferences dialog.
The list of available languages is populated from  Translate.getAvailableLocales(). As my locale is not match to any of this list the selected option is first one. 
On click OK this state is saved in preferences

ArtOfIllusion/src/artofillusion/PreferencesWindow.java
```diff
@@ -168,10 +168,11 @@ void processEvent()
     });
     toolChoice.setSelectedIndex(prefs.getUseCompoundMeshTool() ? 1 : 0);
     localeChoice = new BComboBox();
+    Locale defaultLocale = Locale.getDefault();
     Locale languages[] = Translate.getAvailableLocales();
     for (int i = 0; i < languages.length; i++)
     {
-      localeChoice.add(languages[i].getDisplayName(prefs.getLocale()));
+      localeChoice.add(languages[i].getDisplayName(defaultLocale));
```
peastman: Does this mean that no matter what language you select, the list of available languages will always be displayed in the one returned by `Locale.getDefault()`?

Sailsman63: Yes, it does. Not sure what I think about that. It seems orthagonal to the purpose of this PR, in any case.

makiam:
I think this may be useful to see at least languages list in native language in case of unexpected UI langauge switch

peastman:
I'd assume their native language is the one they've selected.

ArtOfIllusion/src/artofillusion/ui/Translate.java
```diff
@@ -69,22 +89,7 @@ public static Locale getLocale()
   
   public static Locale [] getAvailableLocales()
   {
-    return new Locale [] {
```
Sailsman63: What issue is moving this block supposed to solve?

makiam: Idea is to have single time array initialization

makiam: More important... At line 69 I need this list alrady initialized to set proper 'locale' variable value

Sailsman63: Any reason why line 69 could not simply call `getAvailableLocales()`? Again, refactors to your preferred style _in_ a bugfix are not a good idea.

ArtOfIllusion/src/artofillusion/PreferencesWindow.java
Sailsman63: I really don't like this one. I'd prefer the locale options to be listed in the selected language. If we really wanted to get fancy, they would each be rendered in their own locale, but that would be a whole different feature discussion.

ArtOfIllusion/src/artofillusion/ui/Translate.java
makiam: getAvailableLocales() returns array but availableLocales is a set and it has contains() method to check item exist in it

ArtOfIllusion/src/artofillusion/PreferencesWindow.java
makiam: >I really don't like this one. I'd prefer the locale options to be listed in the selected language.

So if Your cat accidentally switch locale to ex, Japan You prefer to see next time languages list in Japan locale and try to understand which hieroglyphs line is English rather than see it in host default locale?

Sailsman63: If your cat manages that, she needs her own user account... :)

Seriously, though, I'd _prefer_ that each language choice was rendered _in that language_, which is way out of scope for this PR. Your proposed change here is also out of scope for fixing the bug that you found, so does not belong in this PR.

Sailsman63:
Okay, Looks good for now. Thanks.

PR #78, opened 2018-01-06
Catch for NoClasDefFoundError

makiam:
I try to install GLRenderer plugin with SPManager. 
Once application is restarted it stuck on showing TitleWindow.
I see NoClassDefFoundError as this plugin expected classes from old JOGL

### DIFF
--- a/ArtOfIllusion/src/artofillusion/PluginRegistry.java
+++ b/ArtOfIllusion/src/artofillusion/PluginRegistry.java
@@ -204,7 +204,7 @@ public class PluginRegistry
         registerResource(info.type, info.id, jar.loader, info.name, info.locale);
       }
     }
-    catch(LinkageError | Exception ex)
+    catch(Error | Exception ex)
     {
       new BStandardDialog("", UIUtilities.breakString(Translate.text("pluginLoadError", jar.file.getName())), BStandardDialog.ERROR).showMessageDialog(null);
       System.err.println("*** Exception while initializing plugin " + jar.file.getName() + ":");
--- a/ArtOfIllusion/src/artofillusion/PluginRegistry.java
+++ b/ArtOfIllusion/src/artofillusion/PluginRegistry.java
@@ -1,6 +1,6 @@
-/* Copyright (C) 2007-2009 by Peter Eastman
+ /* Copyright (C) 2007-2009 by Peter Eastman
    Some parts copyright (C) 2006 by Nik Trevallyn-Jones
-   Changes copyright (C) 2018 by Maksim Khramov
+   Changes copyright (C) 2018-2021 by Maksim Khramov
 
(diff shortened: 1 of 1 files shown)


Sailsman63:
Not sure why this error is being thrown, rather than `ClassNotfoundException`

Perhaps our load routine needs a little work. I think this is probably a little to targeted as a fix. Next time you find something like this, please file it as a bug report.

Sailsman63:
@peastman how would you feel about catching all (`LinkageError`s)[https://docs.oracle.com/javase/7/docs/api/java/lang/LinkageError.html] here, just to insulate ourselves from any sort of oops that the plugin code generates?

I'd also like to see the Plugin repo updated to mark the current version of `GLRenderer` as incompatible with 3.0.3 and up, as there is a certain type of user that always starts every new install with 'doonload all teh pluginz'

peastman:
That sounds reasonable.  I wonder if we should just catch `Error`?  Even if it's something really strange that leaves the JVM in a broken state, we might at least be able to display a message telling the user which plugin is causing the problem.

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -203,6 +204,12 @@ private static void processJar(JarInfo jar, Map<String, JarInfo> nameMap)
         registerResource(info.type, info.id, jar.loader, info.name, info.locale);
       }
     }
+    catch(NoClassDefFoundError ncdfe)
```
Sailsman63, line 207: As previously mentioned, I think we do need to broaden this, at least to `LinkageError` if not all the way to just `Error`.

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -203,6 +204,12 @@ private static void processJar(JarInfo jar, Map<String, JarInfo> nameMap)
...
+    {
+      new BStandardDialog("", UIUtilities.breakString(Translate.text("pluginLoadError", jar.file.getName())), BStandardDialog.ERROR).showMessageDialog(null);
```
Sailsman63, line 209: As previously mentioned, I think we do need to broaden this, at least to `LinkageError` if not all the way to just `Error`.

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/PluginRegistry.java
Sailsman63: Thanks for the update. I did some further testing with an (easy to make) broken plugin. We do need to just broaden all the way to `Error` here.

I can make the change myself, if you prefer.

Sailsman63 (approved):

Sailsman63:
Thank you!

PR #80, opened 2018-03-13
Fix to View Orietation menu when several cameras have the same name.

peteihis:
This fixes the problem with the Orientation menu, that if there are several cameras with the same name in the scene, the first on the list one always gets selected, which ever the user tries to select.

Also discussed here https://sourceforge.net/p/aoi/discussion/47784/thread/8ba666ec/


peastman:
This looks fine to me.  Look good to you @Sailsman63?

Sailsman63:
Looks good! Thanks.

PR #81, opened 2018-03-25
View handling development

peteihis:
Hi.

This PR is still WIP, but I thought I'd send this for commenting. So far 10 files affected. Most of these changes are really too closely related to be separated. -- _And this time I tried to avoid white space editing,... ;)_

What I have been working on is mostly to redesign the graphic information on view manipulation. It's still there, but I tried to tame it down a bit.

Changes:

- I removed the `mouseMoved()` methods from views for the time being. There also was some interaction with `mouseMoved` and `mouseScrolled`, that emerged after the `mouseMoved` was gone, so I simplified the scroll handling back to, what it was before. 
- Using the scroll wheel sets now that view active, which was not the case before.
- The cue graphics in travel modes are static and they can be switched off in the preferences. 
- The active view visualization in the other views is hopefully less annoying and can be switched off too.
- The colors of the two above are blended closer to the color of the background.
- I tried to [edit] use a combination of [/edit] `renderLine` and `drawLine` to add a 3D-effect on displayed view frustum but I was not happy with the result. Some Issues with both of the drawers so that  the results were a bit messy and really, alpha channel transparency would be the tool to make things happen there.
- The largest changes (mostly in `ViewerCanvas` are, that I turned the view frustum drawing inside out -- now the active view draws it's picture on the other views basically the same way that tools do. 

And of course some new clitches turned up: 

- The switching from tiltable modes back to y=up stopped working. What exactly happens seems to have some connection to drawing the active view. 
- Camera zoom jumps back at the next move or rotate command.
- Light cameras still have issues.

Those aren't anything I haven't seen before, but the causes may be new ones. I'll keep digging.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
+++ b/ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
@@ -1,5 +1,5 @@
-/* Copyright (C) 2002-2009 by Peter Eastman
-   Changes Copyright (C) 2016 by Petri Ihalainen
+/* Copyright (C) 2002 - 2009 by Peter Eastman
+   Changes Copyright (C) 2016 - 2018 by Petri Ihalainen
    Changes copyright (C) 2017 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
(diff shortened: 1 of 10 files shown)


peteihis:
Oh and for the bigger picture: If this was not obvious yet, let's put the #66 on hold now. I'd like to get get all the new fetures and fixes out next. The `mouseMoved` which is needed for teh selection cue hilights needs to wait for the layeres views. The 3Dmanupulator ... I'll see what I can do -- I probably took it too far, and I need to take a few steps back, but let's not delay the 3.1-release for that (rather make a new one sooner)... :)

peteihis:
That took care of the navigation mode issue and the camera jumps (lust silly mistakes, like a missing `else` and a wrong variable sent to a method). Also improved things with lights as cameras but did not quite solve everything ...

peteihis:
Seems that nobody has tested these and now the head is probably too far ahead for a reasonable rebase. I'm taking this down and I'll split it to smaller PR's, where possible.

PR #82, opened 2018-05-10
Separate filters as plugin

makiam:
Separate image filters as plugin. Made this classes store and build as separate subproject. Less hard-coded initialization in AOI main class


Filters/src/extensions.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" standalone="yes"?>
+<extension name="Image Filters" version="3.0">
+    <author>Maksim Khramov</author>
+    <date>2013</date>
```
Sailsman63, line 4: For clarity, it might be nice to say "Based on code extracted from AOI core" in the author field

Also, check the date here...

Sailsman63 (changes requested):

Sailsman63:
This one might be an easy "Okay". I can probably fix the conflict on merge.

Filters/src/extensions.xml
makiam: Updated date and description

Sailsman63:
Actually, I can't get this to build. Please rebase on top of current head, and then we can troubleshoot.

Sailsman63 (approved):

Sailsman63:
Bump. This branch needs to be rebased to be built with the current build-scripts. If you are not comfortable with doing that yourself, I can do it from my end.

makiam:
изображение
This already rebased?

Sailsman63:
I rebased and `git push -f` last night.

On thing I'm thinking about: `ImageFilter` is supposed to be a built-in category. I'm wondering if we should keep the abstract class as part of main application. Theoretically, AOI should be able to load and run 3rd-party image filters even if the core filters are removed or disabled.

peteihis:
> I'm wondering if we should keep the abstract class as part of main application.

That would sound right. Thinking of 3party plugins. I'm wondering if any of the current plugins extend any of the built-in filters?

makiam:
Only plugin I found which adds some extra filters is HDRFilters plugin and it extends filters from  abstract ImageFilter so its OK.
Extra plugins which may be designed in future may declare reference to standard filters plugin via <import .../> clause

BTW... I found that HDRFilters plugin contains not only compiled classes in jar but also source codes for this classes

Sailsman63:
:facepalm I thought that you had moved that abstract class along with the others. I think we're good! I'll do one last test build, and merge a little later.

PR #84, opened 2018-08-05
Fix to uninitialized SoftReferences

peteihis:
Was not initializing the SoftReferences, when a MIPMappedImage was created form an image in the memory (the LotteryBall script does that).

The effect was, that preview images could not be created and therefore the images dialog was unusable before save/open, when a MIPMap was created "online". Nothing was damaged though.


peastman:
Looks good.  Thanks!

PR #85, opened 2018-09-13
Make compatible with Java.Lang.Module Resolves #83

### LINKED ISSUE #83: Java Language pathname conflict
[`artofillusion.procedural.Module`](https://github.com/ArtOfIllusion/ArtOfIllusion/blob/master/ArtOfIllusion/src/artofillusion/procedural/Module.java) conflicts with the new `Java.lang.Module` class.

Since Fully Qualified pathnames are resolved at _compile time,_ existing binaries compiled with Java 1.8 or earlier will still run, but AOI currently refuses to compile under Java 9.

Since the `Java.lang` package is always imported, we have two choices:

- Use the Fully Qualified Name in every call to `artofillusion.procedural.Module` (messy and verbose, will tend to break as other code is added in the future.)
- Change the Classname for `artofillusion.procedural.Module`. This would be a fairly significant refactor, and an API breaker that affects existing plugins.

Sailsman63:
This is just the first step - we need to actually remove the old code at version 4.0

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
+++ b/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
@@ -169,7 +169,7 @@ public class ArtOfIllusion
     PluginRegistry.addCategory(TextureMapping.class);
     PluginRegistry.addCategory(MaterialMapping.class);
     PluginRegistry.addCategory(ImageFilter.class);
-    PluginRegistry.addCategory(Module.class);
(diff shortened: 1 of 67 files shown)


makiam:
why not all built-in modules updated to extend from ProceduralModule?

Sailsman63:
Just checking to see if you were paying attention...

Really, I was tired and thought that I had gotten them all. Good catch, should be fixed now.

PR #86, opened 2018-09-16
Fix javadoc to comply with JDK8 restrictions

makiam:


Sailsman63:
Hey, can I ask where you based this commit? The PR history looks looks like @peteihis has a hand in it, and I find that a little confusing.

makiam:
Looks like I occasionally create this branch not from master but from one of my work branches... I may create a new PR with this fix from master if needed.

peteihis:
To me the network graph looks like this branch was created from the master but from the same spot where Peter merged a small fix I sent some time ago. -- Not _after_ the merge as it probably should have been... A kind of strange.

Anyway, I was a little bit wondering, if fixes like changing `&` to `&amp` should be done? I thought that UTF-8 would be taking care of those things?

Related to that, "Fran�ois" is not displayed correctly on `DefaultToolButton.java`. Sice you are at it. why not try deleting the odd part and just type just "François" in there. (Or copy/paste it from this comment. I _think_ it should work now. :) )

Anyway, you seem to have clarified a couple of comments, that look like my writing. Thanks for those. They seem to make more sense now. :)

makiam:
Replacing & with &amp is required by more strict JDK8 html validation processing not due UTF encoding limitations.

Sailsman63:
@peteihis 

>Peter merged a small fix I sent some time ago. -- Not after the merge as it probably should have been... A kind of strange.

Not what happened. The parent of the first commit in the PR is the merge commit that Peter created. This is normal. I think that the commit was originally based on a commit that you had worked on, maybe in a branch that was only ever on your fork, and then abandoned. He then re-based on master. I suspect that it was just the javadoc update on line 638 of ViewerCanvas.

Sailsman63:
Looks good. Thanks! I'm not sure how that non-utf ç made it through two different cleanup passes.

PR #89, opened 2018-11-29
Fix modifier keys Resolves #87

### LINKED ISSUE #87: Fix View Control Modifiers on Windows & Linux
Somewhere around java 8/9, a change was made in how OpenJDK/Oracle Java handles mouse modifiers.

Middle- and Right-Buttons no longer map directly into `isMetaDown`/`isAltDown` and similar. Also, __Ctrl__ no longer maps to/is read as __Meta__ on Windows or Linux.

This means that all code that handles mouse input, particularly anything that works with view control, needs to be reviewed, and modifier detection needs to be updated.

Sailsman63:
@peastman: Can you check that this works as intended on mac? Controls are mapped as follows:

- Meta == Control == Right Click: Strafe/slide moves
    - +Shift: Zoom(parallel mode) or move along view-Z axis(perspective mode)
- Alt == Middle Click: Rotate. (in Tray/space modes, rotate around a selected center. In fly/drive modes, around camera origin)
    - +Shift: Roll on view-Z axis. (Fly/space modes only)

There are several bits of dead code that don't map at all right now. (The original version relied on combinations of Meta _Plus_ Right click, etc. Need to rethink.) Things _not_ available:

- Modifier to speed/slow movements (adjust sensitivity)
- Modifier to constrain to _only_ X or _only_ Z strafe/rotation. Which one picked by which component of mouse move is greater...

I think that these were added without as much discussion as I would have liked. Do you see any prictical use for either of them?


Sailsman63:
@peastman: while doing the recent window-parenting fixes, did you get a chance to test/look at this?

peastman:
Sorry for not getting to this sooner.  I tested your code, and all modifiers work as described above.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
```diff
@@ -414,6 +417,7 @@ private void moveChildren(ObjectInfo parent, Mat4 transform, UndoRecord undo)
   
   private void tilt(WidgetMouseEvent e, ViewerCanvas view, Point clickPoint)
   {
+	System.out.println("Tilting!!!!");
```
peastman: You probably meant to remove this line.

Sailsman63: I probably did...

It was a leftover from the original view re-do

peteihis:
> 
   -  Modifier to speed/slow movements (adjust sensitivity)
    - Modifier to constrain to only X or only Z strafe/rotation. Which one picked by which component of mouse move is greater...

> I think that these were added without as much discussion as I would have liked.

I'm not sure what you mean by that, but those both are existing 3.0.3:

-The speed control (Alt) affects the scroll wheel. I found it helpful with camera but for some reason pressing and releasing Alt seems to block mouse buttons until you press an release a one of them once. 
- The vertical/horizontal selection by shift key is coded in a different way from the other modifiers. You have to press the other ones first and then use a mouse button to change the action but you could press the shift any time _during_ the movement to lock to mouse-x or mouse-y and releasing shift would immediately return the function back to what it was. I probably never used those but I guess there are users who find them handy.

So those are the original 3.0.3 functions. I did not change those but if you have to map Meta = Control... I'm not sure if we're running out of keys in some cases. _(Just an example of, what definitely not to do, from a different context: I have used a 3D-software where Ctrl+Shift+F3+MouseDrag allow to modify perspective... I can do that, but most users would probably need a third hand....)_

> Meta Plus Right click

Where is that? As I recall for the most part the code did not make any difference between mouse buttons at all.

peteihis:
> `System.out.println("Tilting!!!!");`

Sorry. -- I'll try to remember to clean those off in future.

peteihis:
> I don't have a strong preference about whether the key should need to be held down from the start, or whether you should be able to change it in the middle of drag.

To me it a kind of makes sense the way it was but either way, I don't think,  it would not conflict with the Ctrl+Shift+rotate becoming pan. _(The problem with that mode is now the camera position which should recalculated ... It is on my list-of-things-to-do and I feel guilt for not having the last modifications finished yet. The past year has not been all bad but, when it comes to dividing my focus, it has been a disaster....  )_

Sailsman63:
Okay, I missed that - a lot of lines got re-arranged, and I didn't realize that they were from before.

Any thoughts on what might make a good 'extra' modifier key?


>pressing and releasing Alt seems to block mouse buttons until you press an release one of them once.

This is a windows-only thing. [I've known about it for a while.](https://sourceforge.net/p/aoi/bugs/493/) Fixing it seems like it would take some delving into `KeyEvent` handling. Should deal with it separately.



>    you could press the shift any time during the movement to lock to mouse-x or mouse-y and releasing shift would immediately return the function back to what it was.

> That's probably just a quirk of how I coded it. I don't have a strong preference about whether the key should need to be held down from the start, or whether you should be able to change it in the middle of drag.

That was because tin the old system, `Meta` and `Alt` had special handling built into the `MouseEvent` system, while `Shift` was a separate check for `KeyIsDown` state.

peteihis:
I was re reading the above and something here bothers me now:

>Meta == Control == Right Click: Strafe/slide moves
>  +Shift: Zoom(parallel mode) or move along view-Z axis(perspective mode)

Why would you wire the Control and Meta keys to do the same thing? If I'm not mistaken Ctrl and Meta (or "Command"?) are separate keys on Macs. On Window you would not have a meta but you don't need it either as you always have at least two buttons. If you combine them, you'll lose some of the original functionality.

The original set up should be:

Meta+Mouse button = Right button == horizontal and vertical moves.
Ctrl+Meta+Mouse button == Ctrl+Right button == zoom or depth move.
Meta+Mouse button+Shift = Right button + Shift == horizontal OR vertical move.

I don't think there were any conflicts with that setup.

Sailsman63:
>Why would you wire the Control and Meta keys to do the same thing? If I'm not mistaken Ctrl and Meta (or "Command"?) are separate keys on Macs. On Windows you would not have a Meta but you don't need it either as you always have at least two buttons.

Maybe. Worth a thought. I guess the question is whether we want to depend on that for newer systems. (touchscreen w/stylus, anyone?) That my be over-worrying, though.

peteihis:
> That may be over-worrying, though.

Absolutely.

Keep the Java-9 fix as just that. Leave the Ctrl- and Shift- keys as they were and keep alt and meta options as the SW -emulation to middle and right mouse buttons. The modeller's primary tool is and will be in the foreseeable future the 3-button mouse with a scroll wheel and my professional opinion is, that the AoI-mouse controls for view manipulation are the best and most logical to use, that I have ever found anywhere, so let's not start ruining that asset.

The touch screen usability should be an entirely different projects. With finger operated touch-screens it is not possible or at least reasonable to assume the same usability you can achieve with a mouse. I don't know if there are multi function styluses, that would have something to replace mouse buttons. AFAIK touch-sensitivity is a feature in them but AoI does not have the readiness to receive that information.

For touch screen devices I'd simply add two more icons into the view controls,  ("tilt" and "zoom" I recall thinking) so the set would be complete for finger use. Touch screens already produce clicks and double clicks, so you wold not be missing much. Some modeling tool functions require pressing shift or control. Those would require an addition for finger users, to set the view to "selection mode" for instance but that can not be solved in the Java 9-upgrade.

And (a bit off topic now but) I'd rearrange the tools palette so, that the view controls would be on the top, so they'd be always in the same location and the order would be (IMO) more logical. I'd suggest: view controls, scene editing tools, lights and cameras, primitive objects, other core-AoI objects, plugin objects...  _(Now some plugins push the core-AoI and view icons forward and some not, and as the number of plugin tools is not always even it a kind of stirs up the palette)_

vc

Sailsman63:
>The touch screen usability should be an entirely different project.

Probably true. I was thinking more in the range of notebooks that have such atrocious touchpads that the user would prefer the touchscreen/stylus. Usually, someone with that type of hardware does not have a proper mouse, but the keyboard is built in.

True Touch-screen support probably needs an entirely different approach to the UI, especially if that means trying to support android. (Waaaaay out of scope for the forseeable future)

We can bike-shed the tool layout, &tc,. later. I do want to get the compatibility fixes out soon.

Sailsman63:
I think @peteihis convinced me for now.

Try this out.

- Alt combos as above
- meta == RightClick == strafe x/y
    - + Control == zoom/z-slide
    - + Shift == Constrained slide
- Alt + scroll-wheel == slow zoom/z-slide

I've also got a couple of bugs:

- Ctrl + Middle click gets slow strafe, Alt + Right Click does not (Backwards)
- Shift + Middle Click (Z-roll) is reading as "slow, constrained strafe" if the mode does not allow rolls...

Basically, needs a refactor. 

Anything either of you notice?

peteihis:
> I was thinking more in the range of notebooks

I see your point... And there is nothing to replace Meta. Hmm.... Nothing obvious coming to my mind now, but the tool icons... The keyboard is pretty crowded with shortcuts.

> True Touch-screen support probably needs an entirely different approach

Yeap. Two finger zoom & rotate etc .. I wonder if Java has the capability to begin with?

Sailsman63:
> I just pointed out that Ctrl has been the modifier and Shift the constraint because what you described was not matching that and I have undertood that you did not mean to change the old behavior. Souds like somethig unexpected happening.

Natural result of me being _very_ tired when I wrote that. You are correct, and I'll be changing that bit back shortly....

>>passing the various combos to two separate tools
>I thought it was quite logical after all, the mouse button selecting the tool and the main modifier, what to do with it. The latter happened in the tool, but the modifier key (Ctrl) had to be down before to tool was activated or it was considered up.

I think we are talking past each other here. Let's save it for later.

ArtOfIllusion/src/artofillusion/ui/UIUtilities.java
```diff
@@ -251,4 +253,37 @@ private static void addChildrenToList(Widget w, List<Widget> list)
...
+   * Test event for specified mouse/modifier key combination.
+   * This regularizes 3-button and 1-button + modifier support across
+   * multiple platforms.
+   */
+
+  public static boolean mouseButtonOne(WidgetMouseEvent e)
+  {
+    int mods = e.getModifiersEx();
+    return ((mods & BUTTON1_DOWN_MASK) != 0) &&
+           ((mods & (CTRL_DOWN_MASK | META_DOWN_MASK | ALT_DOWN_MASK)) == 0);
```
peteihis: The this would prevent usin Ctrl as modifier (as for example picking individual objects on screen or removing them from selection).

```suggestion
            ((mods & (META_DOWN_MASK | ALT_DOWN_MASK)) == 0);
```

ArtOfIllusion/src/artofillusion/ui/TreeList.java
```diff
@@ -479,7 +479,7 @@ private void mousePressed(MousePressedEvent ev)
       showPopupIfNeeded(ev);
       return;
     }
-    if (ev.isShiftDown() || ev.isControlDown() || ev.isMetaDown())
+    if (ev.isShiftDown() || mouseButtonTwo(ev) || mouseButtonThree(ev))
```
peteihis: Two mousebuttons simultaneously.

```suggestion
     if (ev.isShiftDown() ||  ev.isControlDown() || mouseButtonThree(ev))
```
The control key is used to pick single (not consequently listed) objects for an action like texturing or making a boolean operation.

ArtOfIllusion/src/artofillusion/MaterialPreviewer.java
```diff
@@ -347,9 +348,9 @@ private void mouseReleased(MouseReleasedEvent e)
     Point dragPoint = e.getPoint();
     if (!clickPoint.equals(dragPoint))
     {
-      if (e.isMetaDown())
+      if (mouseButtonThree(e))
       {
-        if (e.isControlDown())
+        if (mouseButtonTwo(e))
```
peteihis: Two mouse buttons!

```suggestion
         if e.isControlDown())
```

ArtOfIllusion/src/artofillusion/MaterialPreviewer.java
```diff
@@ -380,9 +381,9 @@ private void mouseDragged(MouseDraggedEvent e)
       return;
     Graphics g = getComponent().getGraphics();
     Point dragPoint = e.getPoint();
-    if (e.isMetaDown())
+    if (mouseButtonThree(e))
     {
-      if (e.isControlDown())
+      if (mouseButtonTwo(e))
```
peteihis: Two mousebuttons.
```suggestion
       if (e.isControlDown())
```

Ctrl should turn the move to zoom. Something does happen with two mouse buttons down, but that was not a very well controlled event.

There also seems to be other clitches with the view manipulation here, but that'd take a different study and different update.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
```diff
@@ -154,18 +157,18 @@ private void dragRotateTravelSpace(WidgetMouseEvent e, ViewerCanvas view)
...
 
-		if (controlDown && !e.isShiftDown())
+		if (mouseButtonTwo(e) && controlDown)
 		{
 			view.tilting = true;
 			tilt(e, view, clickPoint);
 			return;
 		}
-		else if (controlDown && e.isShiftDown())
+		else if (mouseButtonTwo(e))
```
peteihis: This is supposed the temporary switce between normal rotate and pan. usin bpths Shift and Control. MouseButtonTwo alone has a different function.

```suggestion
 		else if (e.isControlDown() && e.isShiftDown())
```

ArtOfIllusion/src/artofillusion/RotateViewTool.java
```diff
@@ -208,13 +211,13 @@ private void dragRotateSpace(WidgetMouseEvent e, ViewerCanvas view)
...
 
-		if (controlDown && !e.isShiftDown())
+		if (mouseButtonTwo(e) && controlDown)
 		{
 			view.tilting = true;
 			tilt(e, view, clickPoint);
 			return;
 		}
-		else if (controlDown && e.isShiftDown()){
+		else if (mouseButtonThree(e) && e.isShiftDown()){
```
peteihis: Here too.
```suggestion
 		else if (mouseButtonThree(e) && e.isShiftDown() && e.isControlDown()){
```
Should probably check if the selection by mouseButtonThree has alredy been done....

peteihis:
Almost there. I downloaded the changes and I have been checking those. Mostly it's working fine, but some combinations seem to have gone missing...

The RotateViewTool seems to require special attention: Some times the temporary swithg between normal rotate and pan (or back) seems to work and some times somehinf else hijacks the key combination.

I'm in the middle of checkng that, so don't mind thode commenst just yet. :)  There seem to be cases where you have managed to get rid of unnecessary checkups too...

ArtOfIllusion/src/artofillusion/MaterialPreviewer.java
peteihis: There seems to be more to this than I (or I suppose anybody) would have expected.

With `mouseButtonThree(e)` in the outer if-line of `mouseReleased` the drag ghost is drawn as you would expect: Right button+ctrl zoom the ghost during drag but at `mouseReleased` the rendered image is not zoomed but rotated!

It with control down the `mouseButtonThree(e)` returns false in the case, that the event is `MouseReleased` but with `MouseDragged`event it seems to work correctly. If I change mb3 back to isMetaDown for `mouseReleased`, then the rendered image appears zoomed as it should.

With control-key up the moving of the view with the right buttpn works as expected, first dragging the ghost and in at mouse-up rendering the image where the ghost was.

This is not the the kind of news anybody likes at this point but, if I'm not badly mistaken, this could be causing similar problems elsewhere as well...

ArtOfIllusion/src/artofillusion/RotateViewTool.java
Sailsman63: fixed.

ArtOfIllusion/src/artofillusion/ui/UIUtilities.java
Sailsman63: Good catch.

ArtOfIllusion/src/artofillusion/MaterialPreviewer.java
Sailsman63: I have no idea - My first two thoughts did not pan out. That's frustrating.

ArtOfIllusion/src/artofillusion/ui/TreeList.java
Sailsman63: Yeah, this one should not have been changed at all. Fixed.

ArtOfIllusion/src/artofillusion/MaterialPreviewer.java
Sailsman63: @peastman I've hit a snag here. I did a test, and bare AWT events do track modifiers correctly in `mouseRelease,` even when run through the `mouseButtonThree()` method. That suggests that either:

- The event is intercepted and changed by something else in AOI
- Buoy handles these a little differently? The wrapper seems thin, though. I don't know what it might be doing differently.

Any thoughts on where to begin looking?

Sailsman63:
Okay... I think that this should be ready to go. Before merging, I want to squash the mess into two tidy commits, but before I do that, can you confirm that it works on your systems? Especially @peastman - I need to be sure that I did not inadvertently break Macs again.

peteihis:
And here's the result of my review on the RotateViewTool and MoveViewTool. With these updates I'd say "Released for production".

[ViewTools.zip](https://github.com/ArtOfIllusion/ArtOfIllusion/files/2730820/ViewTools.zip)

I thought this would be clearer this way rather than making all of the little fixes comments.

Mostly the changes I made are just returning a few if-conditions to as they were. 
- No need for mouse button detection inside the tool except for confirming the case, that the tool is selected on the palette.
- No two mouse button cases.
- Can not remove the `!isShiftDown()`s with this structure -- otherwise there would be two `ctrlDown`-cases that react to mouse simultaneously. This is what blocked some of the functions. I know the file is a bit of confusing reading, sorry for that.
- Reverted the immediate control-key effect in MoveViewTool to what is was. It was actually quite nice to use the way you set it but the change should then be made to both of the tools and I did not want to have to check the entire RVT for the corresponding changes. _(I would not bet on it being quite that simple all the way through.)_

By some miracle your latest `mouseButtonX` works correctly on J8 though `getButton()` returns the wrong button on drag... :) Tested that on the Material Previewer. I thought that one safe way of going would have been to detect the button at `mousePressed` and use that information in the later methods, but it seems to be OK now.

peastman:
As far as I can tell, all the modifiers are working correctly.  I didn't exhaustively go through all possible combinations in all navigation modes, but everything I tried seemed correct.

Sailsman63:
@peteihis Thanks for looking at that. I _did_ keep some of the changes:

- using the shorter form of the view-control constants.
- checking for the use of `buttton1` instead of `!button2 && !button3`

However, you seemed to be trying to revert to checking for control and shift in cases where the mouse button check is correct - not sure what was intended there. The changes completely broke being able to rotate around the z-axis, and a few other things.

For future reference, it's _much_ clearer when reviewing changes if they are entered as review suggestions, rather than a single file. If you prefer to _not_ use the review mechanism, you can make a Pull Request directly against the branch is my personal branch - the `Sailsman63:fix-modifier-keys` thing.


>By some miracle your latest mouseButtonX works correctly on J8 though getButton() returns the "button zero" on drag... :)

That's why these checks use _several_ conditions...

peteihis:
Hold on, I'll be back soon.

peteihis:
OK, so .. I assure, that my build was working correctly and nothing was broken, when I sent those files. Can't imagine, what could have happened but now to the current situation:

> However, you seemed to be trying to revert to checking for control and shift in cases where the mouse button check is correct - not sure what was intended there.

Exactly. 

The mouse button is recognized in the view _(SceneViewer lines 352 - 355 and the corresponding structure in the object viewers)_ and the tool to be used is selected based on the mouse button. So when you are in the tool code you already know the button (or more exactly you ended up in there because of that particular button), so...:

It is not only unnecessary to recognize the mouse button _again_ in the tool but even worse: It is logically wrong and there are cases, when forcing the function to a mouse button in the tool actually blocks something else from happening. Inside the tool all you need to check is, what combination of ctrl and shift are down. Obviously if you change isControlDown into mouseButton the result can not be the same.

I'll review & comment into the github files in more detail.

ArtOfIllusion/src/artofillusion/MoveViewTool.java
```diff
@@ -65,43 +67,47 @@ public void mousePressed(WidgetMouseEvent e, ViewerCanvas view)
   {
     Camera cam = view.getCamera();
 
-	selectedNavigation = view.getNavigationMode();
-    controlDown = e.isControlDown();
+    selectedNavigation = view.getNavigationMode();
```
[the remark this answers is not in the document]
Sailsman63: I'd prefer both to work directly from the event. Why add another local variable, when it gains us nothing?
I'm trying to get something that _works_ at the moment. I still believe that all of this code needs a refactor soon, but right now, the stable version of AOI does not work for whatever percentage of users have java 9 or 10 installed on their computer. Priorities.

ArtOfIllusion/src/artofillusion/MoveViewTool.java
```diff
@@ -126,7 +132,8 @@ private void dragMoveTravel(WidgetMouseEvent e, ViewerCanvas view)
 		dx = dragPoint.x-clickPoint.x;
 		dy = dragPoint.y-clickPoint.y;
 
-		if (controlDown) // forward move!
+
+		if (mouseButtonThree(e) && e.isControlDown()) // forward move!
```
peteihis:
```suggestion
		if (controlDown) // forward move!
```

The forward move should also be accessible by mouse button 1, when the tool is selected by the tool icon. On the other hand you can not be here if you use MB2, so the button detection is obsolete in that sense.

ArtOfIllusion/src/artofillusion/MoveViewTool.java
```diff
@@ -181,7 +188,9 @@ private void dragMoveModel(WidgetMouseEvent e, ViewerCanvas view)
 		dx = dragPoint.x-clickPoint.x;
 		dy = dragPoint.y-clickPoint.y;
 
-		if (controlDown) // zoom!
+		
+
+		if (mouseButtonThree(e) && !mouseButtonTwo(e) && e.isControlDown()) // zoom!
```
peteihis:
```suggestion
		if (controlDown) // zoom!
```

As above. If the `!mouseButtonTwo(e)`is actually needed, then something else must be broken: MB2 should have taken you to RotateViewTool.  If the user chooses to press two mouse buttons, the selection still happens in view level.

ArtOfIllusion/src/artofillusion/MoveViewTool.java
Sailsman63: True on both of these.

ArtOfIllusion/src/artofillusion/MoveViewTool.java
Sailsman63: Ditto...

ArtOfIllusion/src/artofillusion/RotateViewTool.java
```diff
@@ -154,18 +160,18 @@ private void dragRotateTravelSpace(WidgetMouseEvent e, ViewerCanvas view)
 		dx = dragPoint.x-clickPoint.x;
 		dy = dragPoint.y-clickPoint.y;
 
-		if (controlDown && !e.isShiftDown())
+		if (mouseButtonTwo(e) && controlDown)
```
peteihis:
```suggestion
		if (controlDown && !e.isShiftDown())
```
It should be possible to use this block by MB1 too, when the tool is selected by the icon. Forcing MB2 here prevents z-rotate in the case of icon-selected tool.

Also you can not be in this part of the code if you pressed MB3, so the detection is obsolete.

The `!e.isShiftDown()` is necessary to separate this case from the next, when the shift-key actually is down.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
```diff
@@ -154,18 +160,18 @@ private void dragRotateTravelSpace(WidgetMouseEvent e, ViewerCanvas view)
...
 		{
 			view.tilting = true;
 			tilt(e, view, clickPoint);
 			return;
 		}
-		else if (controlDown && e.isShiftDown())
+		else if (mouseButtonTwo(e))
```
peteihis:
```suggestion
		else if (controlDown && e.isShiftDown())
```
This is the temporary switch from spatial pan to rotate by pressing the control an shift keys simultaneously. 

The mouse button alone that took us to this part of the code is not the sufficient condition to activate the function.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
```diff
@@ -154,18 +160,18 @@ private void dragRotateTravelSpace(WidgetMouseEvent e, ViewerCanvas view)
...
 		{
 			rotateSpace(e, view, clickPoint);
 			return;
 		}
-		else if (!controlDown && e.isShiftDown())
+		else if (mouseButtonThree(e) && e.isShiftDown())
```
peteihis:
```suggestion
		else if (!controlDown && e.isShiftDown())
```
No MB3 available here!

Sorry for the `!controlDown`, see my other comment below or below that. The code could be restructured so, the negative condition would not be needed, but now I don't trust it to work without it.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
```diff
@@ -208,13 +214,13 @@ private void dragRotateSpace(WidgetMouseEvent e, ViewerCanvas view)
 		dx = dragPoint.x-clickPoint.x;
 		dy = dragPoint.y-clickPoint.y;
 
-		if (controlDown && !e.isShiftDown())
+		if (mouseButtonTwo(e) && controlDown)
```
peteihis:
```suggestion
		if (controlDown && !e.isShiftDown())
```
MB2 is already down, no need to redetect.

The !e.isShiftDown() is necessary to to separate this from the next one: If shift is down then both of the blocks with controlDown will try to respond and one of the loses.

What happens is, that in Space and Fly modes you can not temporarily switch between rotate and pan if you remove the !isShift...

This is a bit heretic structure from coding point of view: It is not the common `if(condition){something}, else{something else}` but a `if(condition){something}, if(not condition){something else}` which does not automatically exclude the other when one happens. At the time I ended up with this structure, because I started to have hard time following the logic inside all the nested if-else blocks, so I made each combination their own case. The structure could now be turned more compact.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
```diff
@@ -208,13 +214,13 @@ private void dragRotateSpace(WidgetMouseEvent e, ViewerCanvas view)
...
 		{
 			view.tilting = true;
 			tilt(e, view, clickPoint);
 			return;
 		}
-		else if (controlDown && e.isShiftDown()){
+		else if (mouseButtonThree(e) && e.isShiftDown()){
```
peteihis: No MB3 available here!

Instead you need to know that control is down. This is the temporary switch from rotate to pan, by pressing both, the control and shift keys.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
Sailsman63: Got it... Perhaps the cases should be reversed?

Not going to happen today, though...

ArtOfIllusion/src/artofillusion/MoveViewTool.java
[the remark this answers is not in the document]
Sailsman63: Okay, agreed there. One more point on the list of things that need to be a little more robust. Again, I hope to refactor to simpler state soon, but we need a working version _ASAP_. I'm hoping for a beta this week.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
Sailsman63: Yep...

ArtOfIllusion/src/artofillusion/MoveViewTool.java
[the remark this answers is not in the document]
Sailsman63: Done... push in a few.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
Sailsman63: got it

ArtOfIllusion/src/artofillusion/RotateViewTool.java
Sailsman63: This one works as intended.

ArtOfIllusion/src/artofillusion/RotateViewTool.java
Sailsman63: got it.

Sailsman63:
I think this is done. (finally) The only thing that does not work quite as I expect is 'driving' mode when using the tool-palette RotateView Tool.

No more refinements on what it "should" do. only stuff that's truly broken.

peteihis:
Well, the z-roll has gone missing from the tool palette rotate, and I consider this to be "truly broken".

The culprit seems to be the `mouseButtonTwo(e)` which is still left on lines 196 ans 217. One of them prevents z-roll in the "icon-mode"+MB1 and the other one is probably just unnecessary.

Also both of the cases need the `!e.isShiftDown()`. Without that the first `controlDown` hijacks also the case, when shift _is_ down, preventing the temporary rotate/pan switching. _(I might have explained this a bit differently the last time. Looking at the file through the changes filter was a kind of confusing to me...)_

While I was just testing it, I thought, I saw something else a little bit off with something probably pan related, but .. ehhh .. it's past midnight here again. 

I think with those fixes it should be good enough and what ever is left, has to be dealt separately. 

I also have some things pending, that I was working on over the holidays. Won't take long any more to get those into PRs. _(I'll try to split things into smaller shipments than in the past though things tend to have dependencies....)_

Sailsman63:
> the z-roll has gone missing from the tool palette rotate, and I consider this to be "truly broken".

It's there. <ALT> + <CTRL> + MB1. Same as using the shortcuts on a mac. Trying to use a shorter shortcut in that case broke other things.

I can't recall right now, but when I tried to add the `!isShiftDown()`, something else more important broke.

>I think with those fixes it should be good enough and what ever is left, has to be dealt separately.

Whatever is left needs to have a refactor/restructure for better readability, first.

>I also have some things pending, that I was working on over the holidays

What sort of things? If they're related to view control, it would be nice to talk about _what_ you want to accomplish before we work with code.

PR #90, opened 2018-12-01
Changed some references from Module to ProceduralModule

peastman:
Without this change, I get an exception loading any file that contains a procedural texture.  I don't really understand why, or why this fixes it.  I suspect it may be a JVM bug.  Anyway, this seems to fix it.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/procedural/IOPort.java
+++ b/ArtOfIllusion/src/artofillusion/procedural/IOPort.java
@@ -19,7 +19,7 @@ public class IOPort
   int x, y, valueType, type, location;
   String description[];
   Rectangle bounds;
-  Module module;
+  ProceduralModule module;
   
   public static final int INPUT = 0;
   public static final int OUTPUT = 1;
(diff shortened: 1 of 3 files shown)


Sailsman63:
This does fix the issue with loading and using procedural textures.

It also breaks compatibility with existing plugins that _provide_ procedural modules. [See Here](https://github.com/ArtOfIllusion/ArtOfIllusion/issues/83#issuecomment-405005227) for the known affected plugins.

If we're okay with breaking these plugins, we should consider going all the way, and remove the legacy `artofillusion.procedural.Module` code entirely.

peastman:
> It also breaks compatibility with existing plugins that provide procedural modules.

I know. :(  If you can come up with a better solution, that would be great.

We could make `Module` extend `ProceduralModule` instead of the other way around.  Then plugins that define `Module` subclasses would continue to work, but any code that creates or accesses modules would break instead.

Sailsman63:
Got it. This one was my fault, and so obvious that I don't know how I missed it.

I had simply copied the Implementation of `Module` wholesale into `ProceduralModule`. This is ultimately what we want, but the compiler sees that as overridden methods. Data was not being passed through to the correct level of the inheritance tree.

I'm sorry that this one caused so much frustration, when the problem was definitely between my chair and keyboard. See my #91 to double check on your end.

peastman:
Of course!  That makes sense.

PR #91, opened 2018-12-02
Temporarily empty ProceduralModule implementation

Sailsman63:
Full implementaion was overriding base class, causing
compatibility problems.

To be reverted when artofillusion.procedural.Module is removed.


peastman:
Shall we go ahead and merge this?

Sailsman63:
Unless you can think of anything else I might have missed. I've tested w/file with a procedural texture from an old release, as well as plugins that provide modules...

peastman:
No, it looks good to me!

PR #93, opened 2018-12-02
Use correct parent when creating dialogs

peastman:
In lots of places we were passing the top level window as the parent for dialogs, instead of the dialog they were opened from.  This caused lots of problems when using Oracle Java on Mac, including dialogs appearing in back of other windows and UI events being missed.

This goes a long way toward making it usable with Java 8 on Mac.  There probably are other cases of this I've missed.  We'll need to keep an eye out for them and fix them as they come up.


Sailsman63:
Looks good from here! Windows is unaffected, and the small glitch that I had noticed on Linux is gone.

The code is also easier to read. I'd say go for it.

peastman:
Great, thanks!

PR #94, opened 2018-12-21
Undo class threading and structure update

makiam:
Simplify class internal structure. 
Avoid awt specific threading code

### DIFF
--- a/ArtOfIllusion/src/artofillusion/Camera.java
+++ b/ArtOfIllusion/src/artofillusion/Camera.java
@@ -1,5 +1,6 @@
 /* Copyright (C) 1999-2009 by Peter Eastman
    Changes copyright (C) 2016 by Petri Ihalainen
+   Changes copyright (C) 2018 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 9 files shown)


makiam:
I replace AWT EventQueue.invokeLater call with ExecutorService because EventQueue is intended for updating UI controls not for writing cache data to disk.
Replacing List with Linked list reduces overhead on insrting data at the head of the list
Introducting private final Runnable cacheWriteRunner gives us less overhead for every write operation
Pack command and commandData fields into internal class gives us more clear to understand and simpler code.
introducing isCommandChangesSelectionOnly function gives us class data internals isolation from other classes.

peastman:
> I replace AWT EventQueue.invokeLater call with ExecutorService because EventQueue is intended for updating UI controls not for writing cache data to disk.

Why do you say that?  EventQueue is a completely generic class for asynchronously executing code on the event thread.  And this does need to be run on the event thread, because we're also modifying the undo stack on the event thread.

>  Replacing List with Linked list reduces overhead on insrting data at the head of the list

You're optimizing something that's totally negligible.  Also, *accessing* elements of a linked list is much more expensive.  The effect is still negligible in this case, but you're just trading off one negligible expense for another negligible expense.

>  introducing isCommandChangesSelectionOnly function gives us class data internals isolation from other classes.

I see it the other way around.  There's nothing special about an UndoRecord that only changes the selection.  The only reason we care about that is that a particular piece of code (in LayoutWindow) cares about it.  You've moved the logic from a place where it's relevant to a place where it isn't relevant.

makiam:
>> The event dispatching thread (EDT) is a background thread used in Java to process events from the Abstract Window Toolkit (AWT) graphical user interface event queue. 
The Undo data writing is not looks like GUI event for me. 

>> Also, accessing elements of a linked list is much more expensive.
thats right but we never use random access in UndoRecord. Only used operations is add/addFirst which is O(1) and sequential access

PR #95, opened 2018-12-21
Stream read test

makiam:
Tests to read and create AOI objects from data streams


Sailsman63:
I'm sorry for the *very* long delay in reviewing this one. I appreciate the effort to improve our test suite.

I've made a couple of false starts on reviewing this one, and have only recently realized why I was having such a hard time with it. I have some general feedback overall that I'd like to get discussed before looking at details.

1. There are whitespace, stylistic, and Generics clarification edits on several AOI core classes. These do not belong in this pull request:
    - First, they further inflate a fairly large changeset, making the whole thing even harder to review.
    - Second, they make tracking back through the commit history more difficult, since they are labeled as changed for a completely unrelated reason.
2. Many of the test names don't describe what they're testing. `testConstructor()` tells me *where* the code under test is, but not what expectation is being confirmed.
3. A lot of the tests that I was able to decipher are just checking that a particular `read()` throws an error if faced with malformed data. Is this a place that we really want to devote a bunch of test space to? How do you expect this to come up in real life?
4. I have some concerns about the whole approach of "Manually generate a chunk of an AOI file for testing." Might it not be better to create a <Thing>, write it out to a stream, and then read the stream back in and make sure that the newly created thing is equivalent?

makiam:
4. From my POV write to the stream and read them back gives us less isolated case to check.
3. I assume this error-checking tests as cases to catch accidental and unintended changes in code with tests run (We run tests indeed?)
1. I got carried away as always )))

Sailsman63:
1. (Got carried away?) That happens... :) Those changes really do need to be kept separate, though.
2. See below.
3. The test, though, should also point out what the tested thing means. This kind of folds back into 2, as well.
4. Perhaps. I'm just nervous about tests that check for changes without pointing out why they'd be wrong.

PR #100, opened 2018-12-29
Load/save app keystrokes with jaxb marshalling

makiam:


makiam:
My main idea is to update code to make it simpler and easy to understand and update in the future. The single line my code to read and save xml replaces many lines of manual read and parse code.

peastman:
> The single line my code to read and save xml replaces many lines of manual read and parse code.

Github reports that this PR adds 91 lines and removes 96, so the overall amount of code is pretty much the same either way.

Sailsman63:
Also, it looks like the `javaxml.bind module` and all packages that it contains are deprecated for removal.

(Possibly as soon as JDK 11 - These classes are part of Java EE)

This would mean that we would have to start vendoring _that_ library as well. Not worth the trouble for this particular instance.

PR #104, opened 2019-02-03
View manipulation & graphics update and bug fix

peteihis:
Hi everybody! 

I call this one done. This \<edit\>is a response to\</edit\> #64 at least partially. And in spite of the branch name it is a bit more than just a graphics update.

Splitting this into several PRs would not have worked as all the commits rely on the earlier ones and I really had to start from the basics (befc7a7, which I should have done in the first place). After attempting a few work-arounds this produced the cleanest out come without methods just to fight the core logic.

- View cameras behave better now and you can continue where you left off after save and reopen -- they remember their last working depth in the view.
- SceneCamera perspective setting overrides the view perspective setting.
- Children of  camera objects should now follow their parent in animated moves better than before
- Scroll cues for fly and drive mode are now less "jumpy". You still need to know, what they mean though.
-\<edit\>Should probably add an off-switch to those too, but IMO those modes need a visible indicator that the behavior is different from the usual.\</edit\>
- View frustum display can be turned off entirely of left on for camera views. I believe that option to be helpful.

SceneCamera still has it's own logic (from 3.0) in parallel viewing mode. It can and should be brought closer to the other views and I think It could have a control to set the actual plot y-size in the scene instead of just the fov-setting but that is a different project. Tilting for tray/drive is not in either. That will also have to be another project.

Please have a look, and let me know if something needs fixing or explaining.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/SceneViewer.java
+++ b/ArtOfIllusion/src/artofillusion/SceneViewer.java
@@ -309,12 +309,13 @@ public class SceneViewer extends ViewerCanvas
 
     // Finish up.
 
-	drawOverlay();
-	currentTool.drawOverlay(this);
-	if (activeTool != null)
-		activeTool.drawOverlay(this);
+    currentTool.drawOverlay(this);
+    if (activeTool != null)
+        activeTool.drawOverlay(this);
+    for(ViewerCanvas v : parentFrame.getAllViews())
+        v.drawOverlay(this);
     if (showAxes)
-      drawCoordinateAxes();
+        drawCoordinateAxes();
     drawBorder();
 }
 
@@ -552,7 +553,6 @@ public class SceneViewer extends ViewerCanvas
   protected void mouseDragged(WidgetMouseEvent e)
   {
 	mousePoint = e.getPoint();
-	drawOverlay();
     moveToGrid(e);
     if (!dragging)
     {
--- a/ArtOfIllusion/src/artofillusion/SceneViewer.java
+++ b/ArtOfIllusion/src/artofillusion/SceneViewer.java
@@ -16,6 +16,7 @@ import artofillusion.image.*;
 import artofillusion.math.*;
 import artofillusion.object.*;
 import artofillusion.ui.*;
+import static artofillusion.ui.UIUtilities.*;
 import artofillusion.view.*;
 import buoy.event.*;
 import buoy.widget.*;
@@ -348,9 +349,9 @@ public class SceneViewer extends ViewerCanvas
 
     // Determine which tool is active.
 
-    if (metaTool != null && e.isMetaDown())
+    if (metaTool != null && mouseButtonThree(e))
       activeTool = metaTool;
-    else if (altTool != null && e.isAltDown())
+    else if (altTool != null && mouseButtonTwo(e))
       activeTool = altTool;
     else
       activeTool = currentTool;
--- a/ArtOfIllusion/src/artofillusion/SceneViewer.java
+++ b/ArtOfIllusion/src/artofillusion/SceneViewer.java
@@ -1,6 +1,6 @@
 /* Copyright (C) 1999-2011 by Peter Eastman
    Changes copyright (C) 2016-2017 by Maksim Khramov
-   Changes copyright (C) 2017 by Petri Ihalainen
+   Changes copyright (C) 2017-2019 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -34,7 +34,6 @@ public class SceneViewer extends ViewerCanvas
   Point clickPoint, dragPoint;
   ObjectInfo clickedObject;
   int deselect;
-  Vec3 nextCenter = new Vec3();
 
   public SceneViewer(Scene s, RowContainer p, EditingWindow fr)
   {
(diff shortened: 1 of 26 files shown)


peteihis:
I was considering separating the view management stuff on their dedicated tab, but chose not to at this point. Still I think, it should be done, when more setting become available.

Sailsman63:
Feels a bit better - I need to consider some other issues first, though...

peteihis:
Found a classic case of creating a method to handle a case and forgetting to use it.

ArtOfIllusion/src/artofillusion/ViewerCanvas.java
```diff
@@ -692,14 +692,17 @@ public void setNavigationMode(int nextNavigation, boolean nextPerspective)
...
-    else
-	  perspective = true;
-	  
-    // Turn y up for landscape modes. Animated
+    // Change perspective without animation. This is a bit blunt but softer  
+    // options would need a more complex set of conditions.
+
+    if (nextNavigation < 2)
+        flipPerspectiveSwitch(nextPerspective);
+    else{
```
Sailsman63: Hey, missed the closing brace here...

ArtOfIllusion/src/artofillusion/SceneViewer.java
```diff
@@ -107,26 +106,46 @@ public void setOrientation(int which)
...
-			nextCenter = new Vec3(coords.getOrigin().plus(coords.getZDirection().times(((SpotLight)nextCamera.getObject()).getDistToPlane())));
-		else if (nextCamera.getObject() instanceof DirectionalLight)
-			nextCenter = new Vec3(coords.getOrigin().plus(coords.getZDirection().times(((DirectionalLight)nextCamera.getObject()).getDistToPlane())));
-		else
-			return;
-
-		animation.start(coords, rotationCenter, 100.0, which, navigation);
+        ObjectInfo nextCamera = cameras.elementAt(which-6);
+        CoordinateSystem nextCoords = nextCamera.coords.duplicate();
+        double dtp, nextScale;
```
Sailsman63: What does dtp mean here? Please expand the abbreviations throughout this section, for clarity...

ArtOfIllusion/src/artofillusion/ScrollViewTool.java
```diff
@@ -227,17 +212,17 @@ else if (navigationMode == ViewerCanvas.NAVIGATE_TRAVEL_LANDSCAPE)
 
 	public void mouseStoppedScrolling()
 	{
-	    if (window != null && boundCamera != null)
+        if (window != null && boundCamera != null)
         {
-            boundCamera.getCoords().copyCoords(camera.getCameraCoordinates());
-            if (boundCamera.getObject() instanceof SceneCamera) ((SceneCamera)boundCamera.getObject()).setDistToPlane(distToPlane);
+            //boundCamera.getCoords().copyCoords(camera.getCameraCoordinates());
```
Sailsman63: Unused code should be removed. Were you keeping it for a reason?

Sailsman63:
Could you take care of the above in separate pull requests? Makes it easier when I do final cleanup...

I do like how this feels, though.

ArtOfIllusion/src/artofillusion/ViewerCanvas.java
peteihis: Sorry. That '{' should have been removed, I had a debugging tag there... Must have deviated from my usual work flow, this would have been caught at compile.

ArtOfIllusion/src/artofillusion/SceneViewer.java
peteihis: OK. Did just this bit. 

There are a couple of almost similar ones elsewhere, but I did not feel like rocking the boat... The equations are the same but the content used varies just so slightly, that getting it right takes focus, that I may not have at this hour...

ArtOfIllusion/src/artofillusion/ScrollViewTool.java
peteihis: Yes, it was meant to be removes. Residue from the earlier life.

ArtOfIllusion/src/artofillusion/SceneViewer.java
peteihis: Situations like this are, why I think, all camera-capable objects should be put under a common interface. Could just ask the bound object the parameter, that is needed without having to check several object types and in some cases calculate the needed number from other information. Most of this code would be compressed to a couple of lines.

peteihis:
The `refix to a previous fix`: I had used a private method, that should have been removed already. The needed action is far more simple now.

Sailsman63:
@peteihis I am aware that it is common practice in some projects to merge "master" onto topic branches that have merge conflicts. Please _do not_ do this for AOI PRs, I'll handle merge issues when merging back to the main branch.

I'm also currently using an interactive rebase to squash some of the clean-up commits into the commits that they are cleaning up, so will force-push in a bit and merge from there.

peteihis:
Ookay... sorry.

I was not aware that there was going to a merge back to the branch, when I edited the conflict. I thought I'd view the conflict and, when I realized I could solve it, I thought I'd give it a try.... 

Happy accident, that it was not anything that changed anything, I guess. I promise,  in future I'll leave conflicts as they are. :)

peteihis:
Thanks, but ehh...

This seems to have, at least for the most part, reverted the already merged #106.

I don't know but could my illegal merging of the master to the branch have caused that? Also before the merge notification I got an other one about something changing, that should have been history already.

I have got all the branches in my local repository and I can join the relevant ones into an one-commit fix if that would help?

Sailsman63:
In terms of functionality? I tested, and thought it worked...

No, your merge commit was completely ignored... I may have made mistakes when doing the merge. Oddly, if I _rebased_ the branch on `master`, the auto-merge was able to figure it out...

peteihis:
It is in the 6da349b that I got notified of and that is where the #106 changes are in the red section. I don't see why d3f93bb and 6da349b appear to be added into #104. I those were already in the HEAD that all of #104 thru #107 are based on.

peteihis:
I had another look at this.

> In terms of functionality?

The rotate/pan switch is broken again. The #106 was the fix to that.

What seems to have happened is, that three commits whose content belonged to earlier history were somehow added to the beginning of #104. Because the commits appeared with new IDs, auto-merge never suspected anything being wrong. The other changes that the PR was based on were included in the rest of the commits/files so those were the reinstated correctly but the poor #106 got written over.

See the screen capture: 
pr 104 edited

I don't know how easy or difficult or even impossible it would be to rewrite the history of that merge, but in case it would be either of the latter I'll push a fix, that contains the changes that belonged to #106 soon.

Also you mentioned, that something still seemed to be off with the navigation modes but did not specify what. -- Well I did stumble into something: After switching modes, using tool-icon manipulation and the pan/rotate ... the drive mode started to drive into a strange direction for a moment. I did not study that any further nor do I know, what exactly triggered it. After switching modes the problem corrected itself.

However, I have a strong hunch where to look: Currently there are duplicate modes for pan and rotate. The error must be in one of those. The arrangement, as heavy and confusing as it is, had to be made because at the time it turned out to be very difficult to have the camera turn around the correct point in a correct way in every situation. So the main pan and rotate operations and the temporarily invoked ones ended up having methods of their own each. This of course should not be the case and I believe it does not have to any more. (I did something in 104 to simplify things and it should help in this case too.)

Also what I think, is that since I wrote those bits, I should be the one to remove them too. Now that I have been able to spend time on the code (that I wrote) again, I have a pretty good idea how it works and I'm sure I can make it more readable for further editing. -- But if I read the current state of things right, that should happen after the 3.1 release.

Sailsman63:
Yeah, when I rebased to clean up some of the fix-of-fix commits, I messed some things up. Working on a fix.

PR #105, opened 2019-02-10
Coordinate axes drawing

peteihis:
Hi.

Renewed drawing of the coordinate axes on the views. The problems were the flickering and the misshaped labels, due to differences in the CanvasDrawers. I also gave up the idea of having the axes react to perspective setting. It probably was not intentional to begin with though a kind of cool, and definitely different from anywhere else. Now it is purely an orientation indicator.

The axes are now handled as an inner class. It did not seem reasonable to have the elements available to the rest of the word.

There is a public static variable to control the size. You can try

~~~~groovy
window.getView().coordinateAxisSize = 45;
~~~~

to test.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -1,5 +1,5 @@
 /* Copyright (C) 1999-2011 by Peter Eastman
-   Changes Copyrignt (C) 2016-2019 Petri Ihalainen
+   Changes Copyrignt (C) 2016-2020 Petri Ihalainen
    Changes copyright (C) 2016-2018 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
@@ -21,6 +21,7 @@ import buoy.widget.*;
 
 import java.awt.*;
 import java.awt.event.*;
+import java.awt.image.*;
 import java.io.*;
 import java.text.*;
 import java.util.*;
@@ -63,6 +64,8 @@ public abstract class ViewerCanvas extends CustomWidget
   
   private static boolean openGLAvailable;
   private static List<ViewerControl> controls = new ArrayList<ViewerControl>();
+  private CoordinateAxes coordinateAxes;
+  
   static
   {
     try
@@ -127,6 +130,7 @@ public abstract class ViewerCanvas extends CustomWidget
     theCamera.setCameraCoordinates(coords);
 	finder = new ClickedPointFinder();
     frustumShape = new FrustumShape();
+    coordinateAxes = new CoordinateAxes();
     setBackground(backgroundColor);
 	setRotationCenter(new Vec3());
 	setDistToPlane(Camera.DEFAULT_DISTANCE_TO_SCREEN);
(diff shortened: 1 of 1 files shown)


Sailsman63:
I've only now just gotten to looking at this one, and I like the result.

I'm not entirely sure that the InnerClass solution is necessary, as all of the variables that you have there _could_ have been local variables within the scope of a single method. I'm not sure that there is any *disadvantage* to doing it that way, though.

ArtOfIllusion/src/artofillusion/ViewerCanvas.java
```diff
@@ -77,6 +80,7 @@ Changes copyright (C) 2016-2018 by Maksim Khramov
     }
   }
 
+  public static int coordinateAxisSize = 30;
```
Sailsman63: Lets make sure we get this right, as it's part of the API of what we can consider an AOI "foundational" class:

Making this a `static` means that any time you change this value through a plugin or script, it automatically applies to *all* `ViewerCanvas`es. Is this the intent?

Also, do we want to hard-code a default here, or should it be an added preference option?

peteihis: > static ..Is this the intent?

Yes it is.

> hard-code a default here, or should it be an added preference option?

I made it so that it's easy to add to preferences but so far we have been living with a hard coded thing. I'd say that the preferences dialog is too crowded already and needs redesign first.

Short answers above, but I can tell more if you wish. :)

peteihis:
> I'm not sure that there is any disadvantage to doing it that way, though.

I saw the definite advantage, that all that is specific to the entity is kept inside it .. And there is quite a lot happening there. Also thinking further a head, I'd favor this model for any other onscreen ... 'widgets' (?)  .. objects. I can think of at least one more but with the current graphics processing that is not feasible.

ArtOfIllusion/src/artofillusion/ViewerCanvas.java
Sailsman63: Okay, you've got me convinced about `static`. 

Also a good point about about the Preferences Dialog. I suspect that it could afford to grow some tabs!

However, if we do, eventually, move this over to preferences, we would have a dilema: a `public` field cannot be made private without marking the API as a breaking change, so it would be tempting to leave the field as-is. But that leaves us open to situations where a script, etc. changes the value, taking it out of sync with the preferences.

So: Other than for testing purposes, do we really have cause to make this value available to every script or plugin that comes along?

Sailsman63:
Okay, I think I see where you are going with the 'mini-widgets' thing, but can you be a little more specific? How do you see this code being called if/when we refactor how the graphics interface is managed?

(I'm not trying to be difficult here, Just the more we specify what we expect, the easier it will be to do it right.)

ArtOfIllusion/src/artofillusion/ViewerCanvas.java
peteihis: So let's make it private.

As a matter of fact I don't think I know anybody who would have ever changed the factory setting of the orientation indicator even if available. Some of the usual things I have seen in preferences would be show/don't, in which corner and size. Everybody just switches it shown (if originally not) and lives happily ever after.

peteihis:
> How do you see this code being called if/when we refactor how the graphics interface is managed?

As before, but the image should preferably be drawn on the overlay layer.

> 'mini-widgets' thing, but can you be a little more specific?

For one thing, the canvas that those are drawn on needs to be able to handle alpha transparency, so it can display texts and animated images without pixelated edges. Currenty SWCDrawer has no problem with that but GLCD chops all partial transparency just off which is just ugly.

One of the tools I'm thinking would be a an on-screen view orientation tool. Those have 6+12+8 predefined orientations available just by clicking it. Very fast and effortless to use. This one also has roll arrows and rotation wheels appear on mouse over.

Orientation tool

Also sometimes a pop-up window as a warning might be overdo, but a "soft warning" would be helpful (when trying to do something impossible). In those cases just a flashing transparent sign on the layer would do the job. You would not have to close the warning to get forward if it were automated.

One possible use could be to keep some information or even controls of the tools you are currently using shown on the screen but subtle enough not to disturb working. 

Those kind of things coming to my mind now. Was I answering your question? :)

Sailsman63:
That does give me some context (and a bit more than that...) We should perhaps set up a different issue/thread to track and discuss some of that. Please understand that *that* is likely to be an API-breaker (Version 4.0 fodder)

The fact that it is a `private class` means that:

 - Plugins can't inject an alternate implementation or graphic (This is fine, as they can't do that with current master anyway)
 - Sub-classes of `ViewerCanvas` can't override the behavior by extending the `CoordinateAxes` class. (They  can't see it!) (They would have to override `drawCoordinateAxes()` instead - again, much as it currently is.)

I think that I'm okay with both of those, as they also mean that we can change anything we want about the implementation, and know that it won't break current code. Just something to be aware of: Any time we do anything with the 'core` classes, we're designing an API, and need to be aware of the implications for plugins and maintainability.

ArtOfIllusion/src/artofillusion/ViewerCanvas.java
Sailsman63: >So let's make it private.

Let's do it. However, I think that it actually belongs down in the `CoordinateAxes` class, (Will need to be marked `static final` to compile) as it is only used in that code.

If we do do eventually move it into preferences, the inner class can check the preference just as easily as the outer can. However, you are right that many options don't really get used, which is another way of saying that they may not be necessary.

peteihis:
Ok. So it might be simplest to just make it a private "black box" that works on it's own and thik about he rest later.

How about making it a class file? There are viewers that don't extend ViewerCanvas but CustomWidget instead. Then it could be used with those too. Though, that kind of an updates to the little viewers are not happening in 3.1.1.  

> different issue/thread to track and discuss some of that

Yeap.

peteihis:
Oookay, did not go as planned.... "All up to date", "No conflicts" but the remote and local had a mis-sync ....  Ended up recreating the branch as new from current master, copy changes there & force push....

Anyway `private static final` seemed to lose it's meaning, it's just `private` now and no global static variable. Forgetting the class file idea for the time being.

Sailsman63:
Gahh! Not sure what happened There!

You mean your remote had gotten out-of-sync from your local? Somehow it's now trying to re-apply changes that are already in master.

I'll reset it to where it was, and add the change (My local has the original version)

Thanks for taking the time to look at it, though.

peteihis:
> You mean your remote had gotten out-of-sync from your local? 

Yes. Ahead to be exact. Attempts to catch up ether did nothing or lost the last commit entirely. I would have thought force push to reset the branch to match the new local, that was now branched from the current head (I have done something like that before), but I see something else going on.... "Files changes 1"  still but I probably _should_ have abandoned this PR and created a new one.

Sailsman63:
No, should have `git reset --hard <commit>` to your local to match local to the commit on the PR, then added the change on top of it as a separate commit. I'd squash-merge from this end.

Sailsman63:
This look good?

By convention, usually inner classes are defined at the bottom of the class, after all the outerclass methods.

Sailsman63:
What happened?  We're back to all the duplicate history, which we don't want!

peteihis:
Thanks. I think we're done.

> Once you've got a mixed up history, pulling and rebasing often don't work as expected.

Git fetch might have had something to do with it... Devil's invention. My master branch looks a bit weird too now. (Hard reset to remote head?)

>  inner classes are defined at the bottom

Yes, it was a comfort thing to write it, where the old code was. Then eventually I just left it where it was.

Sailsman63:
>How about making it a class file? There are viewers that don't extend ViewerCanvas but CustomWidget instead. Then it could be used with those too.

You're speaking of the Texture/Material previewers? They're already a little special... Though, breaking things like this out into a class (Or at least adding an interface) would probably be part of a larger view-drawing refactor...

I'm going to merge as is, we'll leave the rest for later.

>Git fetch might have had something to do with it... Devil's invention.

I actually find `git fetch` *good*. It allows me the chance to get the latest changes, and then decide what to do with them. OTOH, `git pull` used without some care can cause issues - it does `git fetch && git merge` under the hood, and if the local and remote have diverged, you get a mess.

>master branch looks a bit weird too now. (Hard reset to remote head?)

Yes. Make sure that your remote head is not messed up either. If they are both messy, `git reset --hard <head of the core repository>` then `git push -f` to your remote master.

peteihis:
> If they are both messy

Somehow the problem traveled from one place to other... Hard to say what got it started. The reset to upstream/master. I remember now doing it before but not recently. Thanks.

Sailsman63:
Probably, you started doing something in a different branch then your thought you we're in.

Happens. Been there.

PR #106, opened 2019-02-13
Restoring pan/rotate switch

peteihis:
This basically resumes two methods of the state they were in before the Java-9 adaptation. Nothing is new. The latter commit presents the exact same logic in a more conventional form.

Both commits have been tested both on Java 8 and Java 11.


Sailsman63:
Okay, Better.

I'm thinking I'm missing something...
In most modes, and in "free" camera views, this is working as I expected.

However, in "drive" mode, the tool-pallete set to "Rotate Tool," and the view itself locked to a camera or light, `Left Click` does nothing, either with or without the `Ctrl + Shift` combination.

peteihis:
Thanks. Got it.

Sailsman63:
Working better. I'm still feeling like there is some inconsistency with the patterns here between the various modes, but I think this will work well for now.

PR #107, opened 2019-02-17
The strange perspective matrix call

peteihis:
This is something I was wondering about, some couple of years back: There were a couple of variables that were (supposedly) used to set the camera to perspective mode but something there seemed to be a little bit off.

It turns out that the `Mat4.perspective(...)` is always used the same way as the `Mat4.identity()`. The input parameter is always set to 0.0 and scaling and the projection distance are then set in further calculations. The use of the variable seem to suggest that may not have always been the case in the history.

This hopefully clarifies the case a bit.

The `Mat4.perspective(double d)` with the parameter could probably be used for example in the perspective change animations. In the current function the camera position is changed in conjunction with the other parameters, though, I doubt if changing the approach would reduce the amount of calculations at all.


Sailsman63:
Are we sure that there was not a reason for this as a parameter to exist? (it may be debris from an older implementation, but we would need to run through the version history to see.) I'd hate to break code that relied on the current implementation. Since it does not seem that the current _behavior_ is wrong, I want to leave this refactor until after the 3.1 release.

peteihis:
I hear, it has been like that for a very long time, so it can probably wait a bit more, but...

There are not that many calls to `Mat4.perspective` and the d-parameter is always 0. In this case the 0 is first put into a variable, but it is still 0.  And to add confusion, the variable, that would go to the deprecated version of the method, is given a value in the method, that is in use, but naturally the deprecated one is never called. That bit does not look intentional. This is bit of code that I found confusing, when I first stumbled into it. Knowing now what happens there, I thought it'd be easy to clean and would be less confusing to who ever is reading it next.

At the time I asked, Peter did not remember, why it was left the way it was, but scanned the calls and confirmed that the parameter never gets any other value than 0. Also if I read the description of the  `perspective(double d)` is correctly, it would be very hard to predict how it would project the image, with `d <> 0` as the scaling and projection distance are added to the equation later and then there would still be the separate perspective origin _somewhere_...  -- Basically, if you see the analogy, now the perspective matrix only provides the pinhole and the rest of the camera (or better said "projector") is built later.

And I'm not suggesting to remove the more complex form of the perspective matrix. It still might have it's uses but I find it a bit dangerous in the current use cases.

Sailsman63:
>There are not that many calls to `Mat4.perspective` and the d-parameter is always 0

Sounds like it _Might_ be safe to change. I'd rather focus on stuff that makes a difference for users at this moment, though.

I wonder if any _Plugins_ call this code at all?

peteihis:
> I'd rather focus on stuff that makes a difference for users at this moment, though.

Yes, I see your point.

> I wonder if any _Plugins_ call this code at all?

Probably not and if they did this would not change anything to them. The Camera or the SceneCamera classes don't change on the outside and if a plugin uses `Mat4.perspective(double d)`, it is still there. 

Of course plugins that do rendering or create their own visualization could come with cameras of their own (an idea that I a kind of like) and create perspectives starting from the from the start.

peteihis:
Rewrote the history. I hope it makes more sense now. The old methods are still there working the old way if called.

I checked how it was in 3.0. The private 'viewDist' was there in use, but it was being sent to a method that reset it into it's own value, which was 0.0 since the first call to the method... 

Later I had already removed it from the only one actually using it, as a part of separating camera-internal space from the camera-to-scene interaction. As I now recall, I did not remove it entirely because a deprecated method was also still using it and causing an error in the compiler if the parameter was not there with a value.  At the time I did not go tracking where it came from.

PR #108, opened 2019-03-02
The fix to reapply #106

peteihis:
As commented on #104 discussion.


Sailsman63:
Thanks for working on this. I'm actually hoping to revert that merge, and redo the rebase/merge correctly. I'll definitely use this for reference, if nothing else.

Sailsman63:
Just reverted out the original merge, and re-did it with a cleaned up copy. This PR did help as a sanity check! Should be good now.

PR #110, opened 2019-03-08
Fix Illegal reflection - ToolTips

Sailsman63:
@peastman It's been a while since I saw any issues with tooltip z-order on Windows or Linux.

I suspect that Newer JVMs may have improved their handling in this respect. Could you check with a Mac? Might this also have been related to the window parenting issue?

I'd leave it in place, but the more recent, more aggressive deprecation/removal policies make it likely that we will need to deal with this sooner or later. Thoughts/feedback appreciated.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
+++ b/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
@@ -141,24 +141,10 @@ public class ArtOfIllusion
     catch (Exception ex)
     {
     }
+
     JPopupMenu.setDefaultLightWeightPopupEnabled(false);
     ToolTipManager.sharedInstance().setLightWeightPopupEnabled(false);
-    try
-    {
-      // Due to the strange way PopupFactory is implemented, we need to use reflection to make sure
-      // we *really* get heavyweight popups from the very start.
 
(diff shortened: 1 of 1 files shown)


Sailsman63:
@peastman Any thoughts here?

Sailsman63:
@peastman Could you run this one on a mac, just to see if it creates any issues?

peastman:
Sorry for not replying here.  I've been away on vacation.  I'll try to test it once I'm back home.

Sailsman63:
This may have just taken on a little more urgency. I think that some newer builds of OpenJDK 11, etc (particularly on Ubuntu-based systems), have actually broken this code. This seems to be the root cause of [Klaus' issue here](https://sourceforge.net/p/aoi/discussion/47782/thread/b472a1cb10/).

If removing this code still causes issues with tooltips, we might actually need to find an alternative.

peastman:
It should be safe to remove this code.  It was needed to make popups appear over the heavyweight GLCanvas.  But we switched it to a lightweight GLJPanel in #57, so that isn't an issue anymore.

Sailsman63:
Okay, I'm going for it. There may be more similar hiding. Apparently, getting java 8 to work correctly on mint is not easy.

PR #111, opened 2019-03-09
Whitespace cleanup

### LINKED ISSUE #109: White space editing
Hi!

As anyone who has had a look at, what I have done so far can see, I have made quite a mess with the formatting of the code, primarily using tabs for indentation and occasionally panicking with the format and dropping to 2- or 4-space indents and the coming back.... I have been thinking, that before anyone does anything else, I should perform a "grand white wash" to the files, that I have created or edited but before that I'd like to make a suggestion about the formatting.

Now the guideline says, that the indentation style is 2-spaces, but I'm wondering, what the reasoning behind this instruction is or if the reasons are really valid nowadays?

I've got a feeling, that this practice pretty much comes from the era of VGA displays or before, where you could realistically fit about 20 lines on an editor screen and max 80 characters on a line. Nowadays displays are a lot larger. You can easily fit 60 x 200+ characters on the screen and with the latest toys even more.

While the 2-space indent saves space on small resolution displays, on large ones the presentation begins lose clarity of structure. At least that is very strongly my own perception of it. To me writing in 4-space indents makes the code much easier to browse and to keep track of the 'big picture'. Also (though this is really no argument here) Java examples on Oracle are written with 4-space indents and as I might guess, for very similar reasons.

Sure, there would be no point in starting to modify all of the code base into a new style but I'd suggest, that I'd clean/edit the new files (of my writing) and the ones, that I have heavily edited to 4-space indents. This would be rather easy to do even with files of currently mixed styles, while converting tabs to spaces.

While I'm at it, I might _(where I find it bothering me)_ do some other minor styling as well: In my experience for example this

~~~~java
    if(something)
      {
        doThis();
      }
    else
      {
        doThat();
      }
~~~~

is somewhat confusing to read especially, when the blocks get longer or when there are more similar indents inside the blocks. To me this would be more logical to follow:

~~~~java
    if(something)
    {
      doThis();
    }
    else
    {
      doThat();
    }
~~~~

Then especially in cases of if-without-else or a short(ish) for, I might go:

~~~~java
    for(something){
      doThis();
      doThat();
    }
~~~~

What works better seems to depend on, what is inside the block. At least, if there are more blocks, I'd probably want to drop the first brace on it's own line. The examples above are in 2-space style and currently you can find all of the three in different parts of the code, depending on when each piece was written.

So, what do you think? Would you find modernization to 4-space intents, with some other minor flavors reasonable in the mentioned cases or should I strictly stick to the old rules?

peteihis:
Resolves #109. What has been done:

- Mostly just removing tabs, where those from me
- Some comment typos fixed
- Some indent unifying, but not nearly everywhere.
- Removed some copyright notes, that were no longer valid _(on some files I had probably initially made changes during development, but eventually ended up making the change to a super class / somewhere else instead....)_ 
- ..,And added one, I had forgotten :)

Tested, that the compiler has no complaints, the software starts, all the editors, I could think of start, objects can be created... 

I also discovered some really messy historical bits in the code base, but decided, that those are not in the scope of this cleanup. There are quite a few files, where different people have been working in a different style (or in lack of one) and there is code, that has been imported from an external source as it is etc... -- Again: A different project. :)

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
+++ b/ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
@@ -29,7 +29,7 @@ public class ApplicationPreferences
   private boolean keepBackupFiles, useOpenGL, useCompoundMeshTool, reverseZooming, useViewAnimations;
   private boolean drawActiveFrustum, drawCameraFrustum, showTravelCuesOnIdle, showTravelCuesScrolling, showTiltDial;
   private Renderer objectPreviewRenderer, texturePreviewRenderer, defaultRenderer;
-  
+
   /**
    * Create a new ApplicationPreferences object, loading the preferences from a
    * file in the default location.
@@ -48,21 +48,21 @@ public class ApplicationPreferences
     }
     initDefaultPreferences();
     if (!f.exists())
-      {
-        properties = new Properties();
-        Translate.setLocale(Locale.getDefault());
-        return;
-      }
+    {
+      properties = new Properties();
+      Translate.setLocale(Locale.getDefault());
+      return;
+    }
     try
-      {
-        InputStream in = new BufferedInputStream(new FileInputStream(f));
-        loadPreferences(in);
-        in.close();
-      }
+    {
+      InputStream in = new BufferedInputStream(new FileInputStream(f));
+      loadPreferences(in);
+      in.close();
+    }
     catch (IOException ex)
-      {
-        ex.printStackTrace();
-      }
+    {
+      ex.printStackTrace();
+    }
   }
 
   /**
@@ -108,15 +108,15 @@ public class ApplicationPreferences
 
     File f = new File(getPreferencesDirectory(), "aoiprefs");
     try
-      {
-        OutputStream out = new BufferedOutputStream(new FileOutputStream(f));
-        properties.store(out, "Art of Illusion Preferences File");
-        out.close();
-      }
+    {
+      OutputStream out = new BufferedOutputStream(new FileOutputStream(f));
+      properties.store(out, "Art of Illusion Preferences File");
+      out.close();
+    }
     catch (IOException ex)
-      {
-        ex.printStackTrace();
-      }
+    {
+      ex.printStackTrace();
+    }
   }
 
   /** Get the directory in which preferences files are saved. */
@@ -143,9 +143,9 @@ public class ApplicationPreferences
     keepBackupFiles = false;
     useCompoundMeshTool = false;
     reverseZooming = false;
-	useViewAnimations = true;
-	maxAnimationDuration = 1.0;
-	animationFrameRate = 60.0;
+    useViewAnimations = true;
+    maxAnimationDuration = 1.0;
+    animationFrameRate = 60.0;
     drawActiveFrustum = false;
     drawCameraFrustum = true;
     showTravelCuesOnIdle = false;
@@ -160,7 +160,7 @@ public class ApplicationPreferences
     objectPreviewRenderer = getNamedRenderer(properties.getProperty("objectPreviewRenderer"));
     texturePreviewRenderer = getNamedRenderer(properties.getProperty("texturePreviewRenderer"));
     defaultRenderer = getNamedRenderer(properties.getProperty("defaultRenderer"));
-    
+
     defaultDisplayMode = parseIntProperty("defaultDisplayMode", defaultDisplayMode);
     interactiveTol = parseDoubleProperty("interactiveSurfaceError", interactiveTol);
     undoLevels = parseIntProperty("undoLevels", undoLevels);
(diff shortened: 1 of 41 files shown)


peteihis:
Rebased to master

Sailsman63:
I'm not sure that the noise in the commit history is worth it at this point - I'd prefer to keep this open, and add to it when/if we are ready to do a _major_ cleanup of the whole codebase.

peteihis:
Well, now I honestly don't understand your thinking at all.

> I'm not sure that the noise in the commit history is worth it at this point

So essentially, what you are saying that cleanliness of the commit history is more important than the readability of the code itself??

Look at this side: In the 3.0.3-release the is not yet much that I wrote. At the time I did not pay attention to coding style -- did not realize, that is was an issue in the first place -- but since then I got noted, that I should change my writing to match the agreed conventions. Also I keep seeing, that other contributors are instructed to correct even the smallest deviations from the intended style. Well over 90% of the changes here are just changing start of line tabs into two spaces, so basically I'm just cleaning up the mess I made all by myself, as I was asked to do.

I understand, that you don't want white space editing into commits, that do functional changes but now, since this is just the next actual release after 3.0.3 and other development is _a kind of_ on hold for a while, it would be the perfect and most reasonable time to get my coding back on track. That way the damage done stays in the time between the two releases and, in a way, between the current developers. 

> I'd prefer to keep this open, ...

No. I am not looking forward to keeping this branch alive and up to date maybe for years _aside_ of the main development line while the content would be benefiting everybody, if it were _merged_ as early as possible (= now IMO). If things, that we have discussed will start to happen, it will become too difficult to keep the branch up and I'll probably just take it down and not be back into cleaning business again.

> ... and add to it when/if we are ready to do a _major_ cleanup of the whole codebase.

Can you honestly see that time coming one day? I don't.

Art of Illusion is already 20 years old. There is a lot of catching up to do and the world is moving forward all the time. There will always be something else happening and other things, that need improving. Also I can already see, that some things, that I did there could be done better (I'm learning) or are otherwise unfinished. I don't think, that the will ever be a day, when you can say _"Hey, there's nothing to do, let's sit down at the white spaces now!"_ -- Or, the other way around, looking back from (hopefully distant) future: On the day that it happens, white spaces don't matter any more because nobody is reading the code any more anyway.

No. The way I see it, the procedure for doing code standardizing should be that, when ever we need to work on an old piece of code, that is _too_ far off the intended style, the _first_ thing to do is a standardizing commit/PR/merge. Then it should be the time for any the functional changes. That way commenting and correcting would be more fluent as the code at least looks what it expected.

Any other votes?

Sailsman63:
You've made some good points... I seem to have struck a nerve here.

>I am not looking forward to keeping this branch alive and up to date maybe for years aside of the main development line.

I particularly do _not_ want to ask you to do that.

You're probably correct that will never have time to do everything at once. I would like the cleanups to be coherent, though. Rather than 'normalise 5 files throughout the repo that I need to work on,' I'd prefer something like 'I'm going to be working on some of the UI base classes, so normalize everything in `artofillusion/ui`

I'd be glad to hear further thoughts.

I'll take a closer look at this soon

peteihis:
> I seem to have struck a nerve here.

You did. This has been bothering me for a long time. :)

In principle I could keep dropping this in as smaller bits but it may be difficult to decide where to limit the cleanups as for example UI-modifications tend affect files everywhere. 

> ... so normalize everything in `artofillusion/ui`

Above I do have collected the changes to commits per area. Including `artofillusion/ui`.

Sailsman63:
For clarity: of the files that you have worked on here, have you covered the entire file from end to end, or did you only work on the areas where you had previously made changes?

peteihis:
> For clarity: of the files that you have worked on here, have you covered the entire file from end to end, or did you only work on the areas where you had previously made changes?

Well, there are those two main types and then probably something in between. I gave a full treatment for example to `ViewerCanvas ` but on `GLCanvasDrawers` I only replaced the few tabs by spaces (some editing had happened already before). Some others I browsed through and in addition the tab fixes might have changed some of the parts that looked messiest to me.

At the time it did not seem worth the effort to spend time on files, that may have a small difference in style but are quite close, as they are still readable and have been written already 10 or more years ago.

But on the other hand.... I could group the commits into smaller PRs like `viewers` and `viewertools` together and dive a bit deeper into some of the parts, where I, so far, only fixed the most obvious flaws.

Sailsman63:
I'm sort of coming around to doing this sort of thing a bit at a time.

I've recently discovered some extra options to use with the `git blame` command, which will bypass whitespace and line-reordering changes.

I would, however, like to make sure that in most cases, once we've gone over a file, it's _done_.

Sailsman63:
I finally got around to reviewing this,and re-basing to master again.

Thank you for the effort you put in! These parts will be much easier to read now.

I'm actually going to squash, as some of the command line options will ignore these types of changes when trying to determine where a line came from.

PR #112, opened 2019-03-09
ObjectList hide/show fix

peteihis:
First part of  #31.

Two things were wrong:
- The command was in the wrong block
- Visibility was recognized by the size of `itemTreeScroller` _(width or height == 0)_. However the size of the element itself was not changing, when the split-bar was forced to the edge. --> It was thought always to be shown. Now there is a boolean to remember the user's choice. _(Maybe not the ultimate clever solution, but works for now.)_

The command only applies to the panel, that the object list is part of and if the list is afloat, the command does nothing.

EDIT:

> the size of the element was not changing

The visibility of the score is detected the same way and that one works.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -1,6 +1,6 @@
 /* Copyright (C) 1999-2015 by Peter Eastman
    Changes copyright (C) 2016-2017 by Maksim Khramov
-   Changes copyright (C) 2017 by Petri Ihalainen
+   Changes copyright (C) 2017-2019 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -65,7 +65,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
   UndoStack undoStack;
   int numViewsShown, currentView;
   private ActionProcessor uiEventProcessor;
-  private boolean modified, sceneChangePending;
+  private boolean modified, sceneChangePending, objectListShown;
   private KeyEventPostProcessor keyEventHandler;
   private SceneChangedEvent sceneChangedEvent;
   private List<ModellingTool> modellingTools;
@@ -84,6 +84,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
     sceneChangedEvent = new SceneChangedEvent(this);
     uiEventProcessor = new ActionProcessor();
     createItemList();
+    objectListShown = true;
 
     // Create the four SceneViewer panels.
 
@@ -1081,7 +1082,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
     addTrackMenu.setEnabled(numSelObjects > 0);
     distortionMenu.setEnabled(sel.length > 0);
     
-    viewMenuItem[1].setText(Translate.text(itemTreeScroller.getBounds().width == 0 || itemTreeScroller.getBounds().height == 0 ? "menu.showObjectList" : "menu.hideObjectList"));
+    viewMenuItem[1].setText(Translate.text(!objectListShown ? "menu.showObjectList" : "menu.hideObjectList"));
     viewMenuItem[2].setText(Translate.text(view.getShowAxes() ? "menu.hideCoordinateAxes" : "menu.showCoordinateAxes"));
     viewMenuItem[3].setEnabled(view.getTemplateImage() != null); // Show template
     viewMenuItem[3].setText(Translate.text(view.getTemplateShown() ? "menu.hideTemplate" : "menu.showTemplate"));
@@ -1339,6 +1340,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
 
   public void setObjectListVisible(boolean visible)
   {
+    objectListShown = visible; // in case this method is called from outside
     setDockableWidgetVisible((DockableWidget) itemTreeScroller.getParent(), visible);
   }
 
@@ -1642,8 +1644,6 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
           new RenderSetupDialog(this, theScene);
         else if (command.equals("renderImmediately"))
           RenderSetupDialog.renderImmediately(this, theScene);
-        else if (command.equals("hideObjectList"))
-          setObjectListVisible(itemTreeScroller.getBounds().width == 0 || itemTreeScroller.getBounds().height == 0);
         else if (command.equals("images"))
           new ImagesDialog(this, theScene, null);
       }
@@ -1666,6 +1666,8 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
         updateImage();
         updateMenus();
       }
+      else if (command.equals("hideObjectList"))
+        setObjectListVisible(objectListShown = !objectListShown);
       else if (command.equals("fitToSelection"))
         getView().fitToObjects(getSelectedObjects());
       else if (command.equals("fitToAll"))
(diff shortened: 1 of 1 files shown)


ArtOfIllusion/src/artofillusion/LayoutWindow.java
```diff
@@ -1325,6 +1326,7 @@ public ToolPalette getToolPalette()
 
   private void setDockableWidgetVisible(DockableWidget widget, boolean visible)
   {
+    System.out.println(visible);
```
peastman: I think this got left in accidentally.

peteihis: Thanks. -- There's always one that get's away...

Sailsman63:
That simple a workaround? I remember banging my head on this one for hours, the last time I looked at it!

Does not explain why the show/hide score works as expected, and this is broken...

This is still fragile. A script or plugin can call `setObjectListVisible(bool visible)` which causes the actual state of the sidebar to de-sync from the variable. Need to click the menu item twice to re-sync. Do we want to go with this as is, or dig deeper for a real fix?

I still consider removing the option as a viable alternative...

peteihis:
> A script or plugin can call setObjectListVisible(bool visible) which causes the actual state of the sidebar to de-sync from the variable. 

Taken care of.

I'm thinking of this as a temporary kid of a fix, to get the 3.1 release out in tact.  In my opinion the goal should be, that the command hides or shows the panels, not a single item in there, but that'd probably be a bigger makeover to the code.

> I still consider removing the option as a viable alternative...

Well, you know my vote on that. :) -- You are probably right, that this is not used much by "the coincidental modeller", but when you spend a little more time with AoI and learn to use it, you'll probably start to miss it.

If you work on a smaller display, being able to free more viewing space also clears obstacles from thinking. This is very much the experience of every (3d-modelling) colleague of mine with whom the matter has come up. Now, thinking of the usage, I feel that we should add a keyboard shortcut for it. That way it'd be much more usable and useful -- and eventually possibly more used. :)

To me the view menu does not look excessively long yet, but where exactly this particular item should be is a good question. Right now there does not seem to be an obvious place for it, but the next best thing probably is the view menu. Function wise it forms its own group, so it might be logical to move it either between the view-state section and the the fit-commands and fence it with separators on both sides, or to the bottom of the menu with a separator.

peteihis:
There's, what I'd suggest. -- The `isObjectListVisible()` added for a keyboard shortcut.

Now, the hide/show for the score looks like it should be moved last in the animation menu. Then they would be equally easy to access.

peteihis:
A bit outside the topic but fits the bigger picture.

Actually it was not that long ago, when I was wishing, that pulling out and hiding the score would be easier. I don't necessarily build animations, but the animation tracks are handy way of remembering good camera angles.

peastman:
> If you work on a smaller display, being able to free more viewing space also clears obstacles from thinking.

Remember you can do the same thing by just clicking the little arrows on the divider bar.

ArtOfIllusion/src/artofillusion/LayoutWindow.java
```diff
@@ -1344,6 +1346,13 @@ public void setObjectListVisible(boolean visible)
     setDockableWidgetVisible((DockableWidget) itemTreeScroller.getParent(), visible);
   }
 
+  /** Set whether the user has set the object to be displayed. */
+
+  public boolean isObjectListVisible()
```
Sailsman63: Why do we need a public getter here? This state is only relevant for the internals of Layout window.

I'm actually unsure why `setObjectListVisible()` is public either... I don't think that it is used outside of the class.

Sailsman63:
>Remember you can do the same thing by just clicking the little arrows on the divider bar.

I just played with these a bit, and found that they are a _really_ hard target. I use a real mouse, not a touchpad, and I think that I am fairly good at precise click targeting. What I found:

- 1080p display: forget it. It's easier and faster to use the menus, and even faster than that to just drag the divider.
- 720p display:  Sort of reasonable-ish. It's now easier to use than the menu items, about on a par with dragging the divider.

This is because the arrows are at a set pixel size. It is exacerbated by the fact that there is no mouseover indication - If I didn't know about it, I would never guess that they are buttons. This seems to be inherited from the underlying swing code, so I have not yet dug in to see how hard it would be to improve.

peastman:
It all depends on the L&F.  Using the default one on Mac, it's easy to click the arrows even at 1080.

Sailsman63:
Hmm... I'd wondered if that was the case. I'll have to dig in to the L&F for Windows and Linux - They are both fairly small targets.

For context, this is what it currently looks like for me: (Windows 10, 1080 p)

image

Sailsman63:
Other than that public getter, I'm good with these changes as an interim fix. @peastman do you have any other concerns?

peteihis:
> It all depends on the L&F. Using the default one on Mac, it's easy to click the arrows even at 1080.

The users probably can not affect the L&F very much or at least I wouldn't expect them to.

I'd have to agree with @Sailsman63 here. It is virtually impossible to hit those arrows on Windows. I probably went years before I even noticed those arrows. The pointing accuracy, that is required just to pop a panel in or out is just unreasonable. I'd also add, that when the panels are hidden, you are more likely to hit the Windows 10's (IMO idiotic) invisible frame, which is probably 3 or 4 pixels wide and end up stretching the window instead of the panel.

ArtOfIllusion/src/artofillusion/LayoutWindow.java
peteihis: > Why do we need a public getter here? / ... / I'm actually unsure why `setObjectListVisible()` is public either... I don't think that it is used outside of the class. ... 

For a moment I wondered that myself, but all the other similar things are public as well. Now this _almost_ enables external control, but a little something was needed to have that happen properly. As I said: "added for a keyboard shortcut." Went to preferences and added a toggle to the letter `L` (list) -- My other candidate  would have been `I` (item list). 

Looks like this in the xml file.

~~~~xml
<keystroke language="Groovy" modifiers="0" code="76" name="Hide/Show Object List">if (window instanceof LayoutWindow)((LayoutWindow)window).setObjectListVisible(!((LayoutWindow)window).isObjectListVisible());</keystroke>
~~~~

To a _"heavy-user"_ _(a favorite term of my previous employers)_ that would be handy and actually I'd like to add this arrangement for the score too.

The obvious problem of course is, that users probably will not know about the shortcut keys, that are not suggested in the menu item prompts and an unfortunately there is no way of adding single key shortcuts (without modifiers) thru translations so they'd become visible. --> Introduction video needed.

Sailsman63:
>The users probably can not affect the L&F very much or at least I wouldn't expect them to.

No, but we can consider overriding some of the defaults on startup, now that we recognize the issue. (Frankly, I'd forgotten about that functionality!)

ArtOfIllusion/src/artofillusion/LayoutWindow.java
Sailsman63: >unfortunately there is no way of adding single key shortcuts (without modifiers) thru translations so they'd become visible.

There is, however, an overloaded version of the `Translate.menuItem()` method that takes a `buoy.widget.Shortcut`. I think that it should be possible to hard-code a shortcut with no modifiers. We might even be able to localize with a little more code added. This would be my preferred solution.

Sailsman63: To be clear:

Adding a shortcut key/pattern for this operation is a separate matter. I'd rather not add the getter, because then we kind of have to support it, even if we do a better fix later.

peteihis: pushing

peteihis: ok

peteihis:
Somehow I managed to leave two lines of commit message on that earlier one (the hide/show-commit). Now got that removed. Don't know what would have been the correct way of fixing it.

Sailsman63:
You also somehow ended up with duplicate commits. Fixed now.

The correct way to change commit messages is to `git rebase -i <commit before the one you want to re-phrase>` In the rebase file, change `pick` to `reword` for every commit that you want to re-phrase. The system will prompt you for the rest. You will need to do a `git push -f` to get those changes to update to the remote.

Thanks for the work here!

On the subject of shortcut keys, I think that I could do a startup script that would inject one with the code as is. (Including in-menu cues!)

PR #113, opened 2019-03-09
Updated translations

peteihis:
- Mostly sharpening & correcting the Finnish translation
- One translation call bug
- One case fix


Sailsman63:
Thanks!

PR #114, opened 2019-03-16
Fixing a small glitch in fit view to objects

peteihis:
Scale had to be different in perspective and parallel.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -1030,21 +1030,20 @@ public abstract class ViewerCanvas extends CustomWidget
 	newCenter = new Vec3((minx+maxx)*0.5, (miny+maxy)*0.5, (minz+maxz)*0.5);
 	viewToWorld.transform(newCenter);
 
-	double dts;
+    double projectionDist;
 	if (boundCamera != null && boundCamera.getObject() instanceof SceneCamera)
 	{
-		int yp = getBounds().height/2;
-		double fa = Math.PI/2.0 - Math.toRadians(((SceneCamera)boundCamera.getObject()).getFieldOfView()/2.0);///180.0*Math.PI;
-		dts = Math.tan(fa)*yp/100;
+		double compAngle = (Math.PI - Math.toRadians(((SceneCamera)boundCamera.getObject()).getFieldOfView()))/2;
+		projectionDist = Math.tan(compAngle)*getBounds().height/2/100;
 	}
 	else
-		dts = theCamera.getDistToScreen();
-    double newDistToPlane = 100*dts/(double)d/0.9*Math.max(maxx-minx, maxy-miny)+(maxz-minz)*0.5;
+      projectionDist = theCamera.getDistToScreen();
+    double newDistToPlane = 100*projectionDist/(double)d/0.9*Math.max(maxx-minx, maxy-miny)+(maxz-minz)*0.5;
     double newScale;
     if (perspective)
       newScale = 100.0;
     else
-      newScale = 100.0*dts/newDistToPlane;
+      newScale = 100.0*projectionDist/newDistToPlane;
     newCoords = theCamera.getCameraCoordinates().duplicate();
     newCoords.setOrigin(newCenter.plus(newCoords.getZDirection().times(-newDistToPlane)));
 
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -1029,7 +1029,6 @@ public abstract class ViewerCanvas extends CustomWidget
 
 	newCenter = new Vec3((minx+maxx)*0.5, (miny+maxy)*0.5, (minz+maxz)*0.5);
 	viewToWorld.transform(newCenter);
-	newCoords = theCamera.getCameraCoordinates().duplicate();
 
 	double dts;
 	if (boundCamera != null && boundCamera.getObject() instanceof SceneCamera)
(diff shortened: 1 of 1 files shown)


Sailsman63:
I'm not sure what you are trying to fix/change here? What about the current behavior does not work as you expected?

peteihis:
The glitch is that, when you first fit the view to an object and right after that switch between 'space' and 'tray' navigations, there is a small jump in scaling. This only happens in parallel viewing mode and if you do something else between the fit and the navigation mode change, then the problem has taken care of itself. 

Basically ViewAnimation should take care of the correct scale, but I'm sot sure why this is a special case that it misses. -- So this is a kind of a quick fix and the problem is not very serious. Only that it is annoying, when you know it. The down side of fixing is it this way is, that the end scale of fit is now different in parallel and perspective modes.  

I can try to look a bit deeper, if I can find what I have missed, it should not be anything big.

On the longer run, I'd like the fit operations to be based on a real boudung sphere or an 'enclosing ball' as it often seems to be called. I have been _(on and off)_ working on a simple way of generating an englosing ball and it looks like I should work  without stretching or shirinking by iteration. 

Therei is a lot of science written about producing enclosing balls but things like '10-dimensional space' and, using single value decomposition & stuff soud like a bit an overkill in a simple 3D-case. _(+ I really have no idea how they work, but I doubt if you can define for example a set of chemical factors as a ball...)_

peteihis:
Took  fresh look. The problem was closer than I thought. Now the logic is, what I was originally after.

Sailsman63:
Now I see the glitch. Unfortunately, I still see some goofiness with the fix...

I'm a bit unclear on why these two camera managers need different math?

peteihis:
>  why these two camera managers need different math?

I was wondering, when you were going to ask about this.... The reason it that the perspective control of `Camera` and `SceneCamera` are built on an entirely different logic. 

The 'distToScreen' parameter of `Camera` simply sets the projection distance inside the Camera. This being 20 by default means that an object 1 unit high appears 1 unit high on the camera screen and that is of course scaled by the default scaling, 1unit = 100 pixels. The effect of this arrangement is that `Camera` does not have a fixed field of view. When you change the size of a view, the field of view changes, but the projection distance and the scaling don't and your perceived perspective stays the same in relation to you own eye.

`SceneCamera` does not use the a projection distance parameter at all. Instead the thinking there is based on `fieldOfView`, which is fixed to the vertical dimension of the view. Now when the view size changes the scaling has to change accordingly. All you need to change actually is the scale, but to calculate the correct scale you need the information of what the projection distance would be. When we remove the (default) projection distance and scale from the equation, the number that is needed is the tangent of the complementary angle of half of field of view (or cotangent of fov/2 or 1/tan of fov/2). You can find a similar calculation in the `SceneCamera` code for example on line 267.

What makes the case here even more complex is, that to fit the view to something in the scene you have to compare the size of the objects to the size of the screen and add the distance between the viewing point (camera) the object to the calculations.

Now the calculations look like being on the brink on insanity dividing and multiplying the same numbers up and down, but I made that choice to keep at least the parts that are common to all cases simple. Also the similar calculations are done in different parts of the software but in every case the equations are equipped with a different set of information about the environment, depending of what the action is supposed to be.

I hope this clarifies at least some of the background of what is happening there. Unfortunately I don't have any pictures to visualize what is happening. It is true, that there would be different choices for writing the calculations out, but they have to be somewhere. You can not get rid of the different parametrization of the different types of camera.

What I'd like to see in future _(and I'd be happy to make it happen)_ would be a 'ViewCamera' interface that all objects, that can act as cameras for a view, should implement. This would have to include, that all of them are capable of producing the parameters that are need in view handling by them selves, regardless of if they actually use them internally or not. With an interface we'd also get rid of the if-else if stacks like `SceneViewer` 116-123, where every possible type of camera object need to be checked separately. -- And it would be possible to add camera objects by plugins ets. (now a plugin camera would have to be an extension of `SceneCamera` to work.)

Above I have not touched the relation between the scene and the internal parameters of the cameras, but I'll just say that in the current form of things they are now completely separated. _(Reference to real life: You can use the same camera to take a picture of a paper clip or a mountain. When I started, this was not the case in views of AoI ... You were supposed scale the camera to match the size of the mountain or the paper clip. In some cases it was done and in some not.)_

Sailsman63:
Ah... I meant, why does the seen to be a difference between 'tray' and 'space'? Transitioning them should not switch between `Camera` and `SceneCamera`... Is an I missing something?

peteihis:
> Ah... I meant, why does the seen to be a difference between 'tray' and 'space'? Transitioning them should not switch between `Camera` and `SceneCamera... Is an I missing something?

No, the handling is the same in both navigation modes, the navigation mode is not even checked here, and it does not switch between `Camera` and `Scene Camera` ether but this method, of course, _checks_ if there is a `Scene Camera` present and acts accordingly. `SceneCamera` does not comply with view handling in parallel mode well, but that needs some work on the Scene camera itself.

The glitch (that this corrects) was, that in this method I had used a sightly different working depth value to calculate the new camera position and the new scale for the animation engine. The animation engine does not notice the mismatch but switching between tray and space recalculates the scale with the correct values. (Line 705.) Some other parts of the code recalculate the working depth or the camera position based on the scale instead, which too rematches the values but in an order, that does not show on the screen. 

The history of the mistake is probably that in 3.0 and before, in parallel mode, the camera was placed at a fixed distance from the assumed target. That is fine as long as you only use the camera to produce the view-image, but as soon as you start controlling objects with the camera position or have a pan function, the camera is best placed the same way for both in parallel and perspective modes. Making the change from fixed to proportional camera position, I had to check all the places where the difference existed. Here I had a slight case of miswiring in the process.

Sailsman63:
There's a secondary, related-ish bug here. When switching to the "Travel" modes, projection is forced to `Perspective`, but the dropdown is only grayed out, it does not reflect the projection change.

Why must the travel modes have `Perspective`? Could they not function just as well with `Parallel`, if that is what the user wanted?

peteihis:
> When switching to the "Travel" modes, projection is forced to Perspective, but the dropdown is only grayed out, it does not reflect the projection change.

I am aware of that. Currently the `ComboBox` reflects the latest user selection, where the view would be returned after returning on space of travel. The logic is different from the navigation mode selector, which is a bit confusing. I can take a look at that, but as I recall it needs a few additions to the code that I did not want to include into the latest bigger update. It works still? I guess not worth delaying the release?

> Why must the travel modes have Perspective? Could they not function just as well with Parallel, if that is what the user wanted?

Technically it of course would be possible to leave the view to parallel viewing for the travel functions, but that way the mode would not make any sense the user. In parallel mode you don't have any reference to how deep in the scene you are, which really is fundamental idea of the modes. -- When you only have a tube vision to the scene you don't have any reference to how deep in the scene you are. For example 
scrolling forward you would not know, you did, and then turning even a little bit you'd end up seeing an infinitely small section of the horizon without any idea where you are.

If you can not figure the trouble out to yourself, there is very little I can do to explain it well enough and I'll just have to say "take my word for it".  As I recall, we have had bits of this discussion before. Mapping out all the scenarios of different attempts for a parallel viewing implementation, would take a very long post but the result would still be the same: No.

As a matter of fact I was considering to make the perspective effect stronger for the travel modes. It helps with the feel of space there, but I left that option out for the time being.

Sailsman63:
>In parallel mode you don't have any reference to how deep in the scene you are,

Gotcha.

I think that we are about ready to get some feedback from other users.

peteihis:
> I think that we are about ready to get some feedback from other users.

How about merging this fix fisrt? ;)

Sailsman63:
I certainly intended to merge this fix! Thanks for the effort.

Whitespace cleanup does nothing to the functionality, so it should not delay the release. We'll come back to that once the release is final.

peteihis:
Thanks

> We'll come back to that once the release is final.

OK... I didn't think it would delay it, but now there already are some conflict bits. Your choice, of course.

PR #115, opened 2019-04-22
Vector to list clean

makiam:
Replace obsoleted collection classes usages (i.e Vector and Hashtable) with new List and Map implementations


Sailsman63:
Please break this up into manageable chunks.

- The `SceneTest` stuff should _not_ be merged into this PR. I'm holding off on reviewing that until after we release 3.1.0, but it _is_ a separate issue.
- Adding tests to `Actor` should be it's own project.
- Vector-To-List should be separated from Hashtable-to-map.
- `vector` has some very specific multi-threading guarantees, which ArrayList lacks. While many sources state that `vector`'s implementation is sort of useless, have you checked to make sure that the underlying code is not relying on this behavior?

PR #117, opened 2019-04-29
Fix some resource issues

makiam:
Fix some russian locale issues. Create missed errorExportingImage resource (English only). Update Image dialogs resource usages


peteihis:
Ah, that's how it should be done! 

The '-HEAD' and '-TAIL'  lines in the properties files (Finnish translation 1283-1285) that are replaced by this should be removed. They are not needed by anything else.

makiam:
> I'm not familiar with the Object... args. Somebody else might want to comment on that?
just a simpler form to declare method with unlimited arguments. Instead of pass array of values just pass each value one by one

Sailsman63:
The Russian translations were very recently added, See [sourceforge ticket](https://sourceforge.net/p/aoi/feature-requests/389/)

Could you please communicate through the ticket with AlexL regarding the differences in opinion? If it comes from differing understandings of what the English means, I or Peter will be glad to clarify.

makiam:
Add some comments in SF ticket. Do anyone see updates from closed ticket?

Sailsman63:
Yes, updates are still sent to those who are subscribed, including the person who opened the ticket.

PR #120, opened 2019-05-28
Removing duplicate line.

peteihis:
As the subject states.


Sailsman63:
Thanks! Not sure that it was doing any harm, but not needed!

PR #123, opened 2019-06-28
Change plugin notification layout

makiam:
Created PluginRegistry.notifyPlugins as common code to notify all plugins of some type
SCENE_WINDOW_CREATED/SCENE_WINDOW_CLOSING notifications as only it uses such notifications

APPLICATION_STOPPING event handled via Runtime shutdown hook in PluginRegistry class... for GUI applications as no  more active windows opened the shutdown hook automatically calls with notifier.


Sailsman63:
What motivated this change? In other words, what makes this an improvement?

makiam:
PluginRegistry.notifyPlugins avoid us to write for(...) { processMessage(...) } in multiple locations.
Using shutdown hook gives more clean processing for ArtOfIllision,quit(). Move SCENE_WINDOW_CREATED/SCENE_WINDOW_CLOSING do the same for newWindow/closeWindow methods

Sailsman63:
Just got back to being able to review this one further. Overall, I do like the main concept. Due to the time that has passed, it does need a rebase (Sorry!). Some more detailed feedback to follow.

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
```diff
@@ -52,7 +50,7 @@ so no instance of this class ever gets created.  It starts up the application, a
   private static Texture clipboardTexture[];
   private static Material clipboardMaterial[];
   private static ImageMap clipboardImage[];
-  private static ArrayList<EditingWindow> windows = new ArrayList<EditingWindow>();
+  private static LinkedList<EditingWindow> windows = new LinkedList<EditingWindow>();
```
Sailsman63, line 53: Changing list implementations should have a specific reason. What makes you prefer a `LinkedList` here?

ArtOfIllusion/src/artofillusion/LayoutWindow.java
```diff
@@ -811,6 +813,18 @@ public void showPopupMenu(Widget w, int x, int y)
...
 
+
+  @Override
+  public void setVisible(boolean visible)
+  {
+    System.out.println("svc....");
+    if (visible && !hasNotifiedPlugins)
+    {
+      hasNotifiedPlugins = true;
+      PluginRegistry.notifyPlugins(Plugin.class, Plugin.SCENE_WINDOW_CREATED, this);
```
Sailsman63, line 824: I think these are reversed. We probably shouldn't claim that we've notified plugins until we've... notified plugins.

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -35,6 +35,39 @@ Some parts copyright (C) 2006 by Nik Trevallyn-Jones
...
+      public void run()
+      {
+        Object[] pso = new Object [0];
+        PluginRegistry.notifyPlugins(Plugin.class, Plugin.APPLICATION_STOPPING, pso);
+      }
+    }, "Plugin shutdown thread"));
+  }
+  
+  
+  public static <T> List<Throwable> notifyPlugins(Class<T> category, int message, Object... args)
```
Sailsman63, line 50: Why are we returning a list of `Throwable`? AFAICT, it's never used.

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -35,6 +35,39 @@ Some parts copyright (C) 2006 by Nik Trevallyn-Jones
...
+          try 
+          {
+            ((Plugin)plugin).processMessage(message, args);
+          } 
+          catch(Throwable tx)
+          {
+            errors.add(tx);
+          }
+        }
+        return errors;
```
Sailsman63, line 66: And here, swallowing the errors without doing anything... These errors need to either be logged or bubbled up to the user.

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
makiam: This was tied to next changes in code where LinkedList's removeLast() was uses. Seems no more actual.

Sailsman63: That works for that one.

Sailsman63:
Um... not sure how that happened, but that's not what a rebase looks like. Would you like me to fix it, or should I walk you through the process?

makiam:
Can You please fix this...

Once I started rebase I proceed with resolve conflicts twice and then when I see branch diverged warning I merge changes.
Seems that was not right.
What was right steps?

Sailsman63:
Okay, got the rebase fixed.

**Before you do *any* more work on this branch, do the following:**
- `git fetch <the remote name for your fork repository>`
- `git reset --hard 8015406`
- If you don't do the above, you will get more errors and possibly undo what I just did


>when I see branch diverged warning I merge changes.

And that was the mistake. It sounds like you did the rebase correctly. But then you have to use `git push -f` (the "f" is for "force") to "Write over the existing branch, rather than keep existing changes." Typically, you don't want to use `-f`, but it's appropriate when re-basing a feature branch like this one.

ArtOfIllusion/src/artofillusion/LayoutWindow.java
```diff
@@ -881,6 +883,26 @@ public void showPopupMenu(Widget w, int x, int y)
...
+
+  @Override
+  public void setVisible(boolean visible)
+  {
+    List<PluginRegistry.PluginThrowable> errors = null;
+    if (visible && !hasNotifiedPlugins)
+    {
+      hasNotifiedPlugins = true;
+      errors = PluginRegistry.notifyPlugins(Plugin.class, Plugin.SCENE_WINDOW_CREATED, this);
+      errors.forEach(error -> System.out.println("Plugin " + error.getSource().getClass().getSimpleName() + " throws error: " + error.getCause().getMessage()) );
```
Sailsman63, line 895: I don't think we want to do the `println` stuff here...

ArtOfIllusion/src/artofillusion/LayoutWindow.java
```diff
@@ -993,6 +1015,8 @@ public boolean confirmClose()
       if (choice == 2)
         return false;
     }
+    List<PluginRegistry.PluginThrowable> errors = PluginRegistry.notifyPlugins(Plugin.class, Plugin.SCENE_WINDOW_CLOSING, this);
+    errors.forEach(error -> System.out.println("Plugin " + error.getSource().getClass().getSimpleName() + " throws error: " + error.getCause().getMessage()) );
```
Sailsman63, line 1019: Or here...

ArtOfIllusion/src/artofillusion/ObjectEditorWindow.java
```diff
@@ -486,14 +487,21 @@ public void redoCommand()
...
   {
+    List<PluginRegistry.PluginThrowable> errors = null;
     if (visible && !hasNotifiedPlugins)
     {
       hasNotifiedPlugins = true;
-      List<Plugin> plugins = PluginRegistry.getPlugins(Plugin.class);
-      for (int i = 0; i < plugins.size(); i++)
-        plugins.get(i).processMessage(Plugin.OBJECT_WINDOW_CREATED, new Object[] {this});
+      errors = PluginRegistry.notifyPlugins(Plugin.class, Plugin.OBJECT_WINDOW_CREATED, this);
+      errors.forEach(error -> System.out.println("Plugin " + error.getSource().getClass().getSimpleName() + " throws error: " + error.getCause().getMessage()) );
```
Sailsman63, line 495: Or here...

ArtOfIllusion/src/artofillusion/ObjectEditorWindow.java
```diff
@@ -503,9 +511,8 @@ public void setVisible(boolean visible)
   @Override
   public void dispose()
   {
-    super.dispose();
-    List<Plugin> plugins = PluginRegistry.getPlugins(Plugin.class);
-    for (int i = 0; i < plugins.size(); i++)
-      plugins.get(i).processMessage(Plugin.OBJECT_WINDOW_CLOSING, new Object[] {this});
+    super.dispose();    
+    List<PluginRegistry.PluginThrowable> errors = PluginRegistry.notifyPlugins(Plugin.class, Plugin.OBJECT_WINDOW_CLOSING, this);
+    errors.forEach(error -> System.out.println("Plugin " + error.getSource().getClass().getSimpleName() + " throws error: " + error.getCause().getMessage()) );
```
Sailsman63, line 516: Or here either...

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -36,6 +35,35 @@ Changes copyright (C) 2018-2022 by Maksim Khramov
   private static final HashMap<String, ExportInfo> exports = new HashMap<String, ExportInfo>();
   private static final HashMap<String, Object> classMap = new HashMap<String, Object>();
 
+  static {
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+        List<PluginRegistry.PluginThrowable> errors = PluginRegistry.notifyPlugins(Plugin.class, Plugin.APPLICATION_STOPPING);
+        errors.forEach(error -> System.out.println("Plugin " + error.getSource().getClass().getSimpleName() + " throws error: " + error.getCause().getMessage()) );
+    }, "Plugin shutdown thread"));
```
Sailsman63, line 42: Nor here either.

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -36,6 +35,35 @@ Changes copyright (C) 2018-2022 by Maksim Khramov
...
+        List<PluginThrowable> errors = new ArrayList<>();
+        for(Object plugin: categoryClasses.get(category))
+        {
+          try 
+          {
+            ((Plugin)plugin).processMessage(message, args);
+          } 
+          catch(Throwable tx)
+          {
+            errors.add(new PluginThrowable((Plugin)plugin, tx));
```
Sailsman63, line 59: The system println calls should instead be in here. Note: the stacktrace shows the code that caused this to be called, so we can always tell what kind of event triggered the error.

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/PluginRegistry.java
Sailsman63: Eventually, they will be replaced with Logging calls, once we actually start work on #50.

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -36,6 +35,35 @@ Changes copyright (C) 2018-2022 by Maksim Khramov
...
 
+  static {
+    Runtime.getRuntime().addShutdownHook(new Thread(() -> {
+        List<PluginRegistry.PluginThrowable> errors = PluginRegistry.notifyPlugins(Plugin.class, Plugin.APPLICATION_STOPPING);
+        errors.forEach(error -> System.out.println("Plugin " + error.getSource().getClass().getSimpleName() + " throws error: " + error.getCause().getMessage()) );
+    }, "Plugin shutdown thread"));
+  }
+  
+  
+  public static <T> List<PluginThrowable> notifyPlugins(Class<T> category, int message, Object... args)
```
Sailsman63, line 46: I'm still unsure if we should be returning an error list here at all. I'd prefer the calling code to not have to worry about this at all.

If we *do* need this, why not return a `Map <Plugin, Throwable>` instead of this custom type?

Sailsman63 (commented):
I'd really like to know what reasons you have to preferring returning an error list to the caller rather than outputting it in the notification code.

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
```diff
@@ -101,7 +99,7 @@ so no instance of this class ever gets created.  It starts up the application, a
     }
     
   }
-
+
```
Sailsman63, line 102: Blank spaces

ArtOfIllusion/src/artofillusion/ObjectEditorWindow.java
```diff
@@ -486,26 +486,27 @@ public void redoCommand()
...
+      errors = PluginRegistry.notifyPlugins(Plugin.OBJECT_WINDOW_CREATED, this);
     }
     super.setVisible(visible);
+    if(errors != null && !errors.isEmpty())
+    {        
+        ArtOfIllusion.showErrors(errors);
+    }
   }
-
+
```
Sailsman63, line 501: And here

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
Sailsman63, line 102: Added blank spaces

ArtOfIllusion/src/artofillusion/ObjectEditorWindow.java
Sailsman63, line 501: And here.

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -754,4 +786,5 @@ public ClassLoader getClassLoader()
     String type, id, name;
     Locale locale;
   }
+
```
Sailsman63, line 789: And a blank line added.

Sailsman63 (changes requested):
Please be careful of whitespace changes...

Sailsman63:
Okay, this is good to go. (I did rebase again for a cleaner history.) Thanks!

PR #124, opened 2019-07-08
Some test cases for implicit objects

makiam:


Sailsman63:
The changes to `ImplicitSphereTest` are incompatible with the current, refactored set of tests.

PR #127, opened 2019-07-14
Preview respond to mouse

### LINKED ISSUE #125: View tools don't update tool previews
I just noticed, that the mouse view tools (rotate, move, scroll) do not update the previews of Boolean, Extrude, Lathe.... The tools do work but there is a error in the call to view repaint and the effect of the move or rotate will be seen, when the view is updated by something else like a change to the parameters on the dialog.

All the viewtools come with the same basic structure and I copied the same mistake to all of them. Extremely unfortunate, that this got through the betas and all..... Earlier in the process I tried to remember to check all the little views too, but on the last steps I obviously forgot these.

I'll push a fix soon --> 3.1.1 or a "correction package" to 3.1.0?

peteihis:
Resolves #125

While I'm at it, I could add scroll wheel zoom to the previews too, if this is not merged first.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/MoveViewTool.java
+++ b/ArtOfIllusion/src/artofillusion/MoveViewTool.java
@@ -118,8 +118,9 @@ public class MoveViewTool extends EditingTool
 		if (view.getBoundCamera() != null)
 			view.getBoundCamera().getCoords().copyCoords(view.getCamera().getCameraCoordinates());
 		view.frustumShape.update();
-		if (ArtOfIllusion.getPreferences().getDrawActiveFrustum() || 
-		   (ArtOfIllusion.getPreferences().getDrawCameraFrustum() && view.getBoundCamera() != null))
+		if (theWindow != null &&
+		   (ArtOfIllusion.getPreferences().getDrawActiveFrustum() || 
+		   (ArtOfIllusion.getPreferences().getDrawCameraFrustum() && view.getBoundCamera() != null)))
 			theWindow.updateImage();
 		else
 			view.repaint();
--- a/ArtOfIllusion/src/artofillusion/RotateViewTool.java
+++ b/ArtOfIllusion/src/artofillusion/RotateViewTool.java
@@ -147,8 +147,9 @@ public class RotateViewTool extends EditingTool
 		if (view.getBoundCamera() != null)
 			view.getBoundCamera().getCoords().copyCoords(view.getCamera().getCameraCoordinates());
 		view.frustumShape.update();
-		if (ArtOfIllusion.getPreferences().getDrawActiveFrustum() || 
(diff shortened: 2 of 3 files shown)


Sailsman63:
I'll test & probably merge in the morning. (UTC-6)

ArtOfIllusion/src/artofillusion/MoveViewTool.java
```diff
@@ -118,8 +118,9 @@ public void mouseDragged(WidgetMouseEvent e, ViewerCanvas view)
 		if (view.getBoundCamera() != null)
 			view.getBoundCamera().getCoords().copyCoords(view.getCamera().getCameraCoordinates());
 		view.frustumShape.update();
-		if (ArtOfIllusion.getPreferences().getDrawActiveFrustum() || 
-		   (ArtOfIllusion.getPreferences().getDrawCameraFrustum() && view.getBoundCamera() != null))
+		if (theWindow != null &&
+		   (ArtOfIllusion.getPreferences().getDrawActiveFrustum() ||
```
makiam: Querying for preferences value on every mouse move looks not the good solution. IMO implementing property change listener for propertie @ Preferences and listen for these changes @ tool class is more elegant.

peteihis: I'll agree, that this does not look very good. 

The reason to check the preferences is to avoid constant redraw on all views, when there is nothing changing.  Unfortunately if there are any moving graphics on the view, the only way to update them is the redraw the 3D-objects as well. 

Putting a listener here would be a bit of an overkill to me: It is not really possible, that the values of preferences would change during view rotation/move/zoom.  It would probably work best to check all the above conditions at mousePress, put the result into a `boolean updateWindow` and use that during drag. -- Did not come to my mind at the time, though.

I'm hoping to get rid of the problem entirely in the future by making the ViewerCanvas layered: Have the 3D on its dedicated (OGL)layer and do the mere graphics on a transparent layer on top of it. This is actually a necessary development step to modernize the graphic feedback to the user. Thinking of adding pre-hilight of objects / lines / vertices on mouseMove, implementing a real hilight of moving objects instead of just showing the drag handles, displaying warnings and cues directly on the view instead of popping up dialogs (which currently does not happen at all).

makiam: > Putting a listener here would be a bit of an overkill to me: It is not really possible, that the values of preferences would change during view rotation/move/zoom
I mean that the fields in Preferences may be listenable and the XXXTool may listen on these changes
I try this approach in #47

Sailsman63: Either way, you then still need to store the preference value in a local variable.

Discussions like this need to be in their own issues - this is off topic for this PR, and will be harder to find later. (PR is about to be closed)

PR #131, opened 2019-07-23
Confirm imageconvert

### LINKED ISSUE #130: Two part translation not working correctly
At least this translation, that is split into two parts is not working as it is supposed to. The ´{0}´ is supposed to be the name of the image, that is going to be processed.

translation_error

This is in **Scene / Images... / View Details... / Convert to local** that is used on linked images. Can't say what is wrong yet but should probably double check the other similar cases too.

peteihis:
Resolves #130

### DIFF
--- a/ArtOfIllusion/src/artofillusion/image/ImageDetailsDialog.java
+++ b/ArtOfIllusion/src/artofillusion/image/ImageDetailsDialog.java
@@ -318,7 +318,7 @@ public class ImageDetailsDialog extends BDialog
             return;
 
         String name = im.getName();
-        if (name.isEmpty());
+        if (name.isEmpty())
             name = Translate.text("unNamed");
 
         if (confirmConvert(name))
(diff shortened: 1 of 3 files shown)


ArtOfIllusion/src/artofillusion/image/ImageDetailsDialog.java
```diff
@@ -318,7 +318,7 @@ private void convertToLocal()
             return;
 
         String name = im.getName();
-        if (name.isEmpty());
+        if (name.isEmpty())
```
makiam: BTW... This is a good example that no-brace code style is no good

peteihis: I fail to see the connection. It happens just as well with or without braces.

Sailsman63:
Good catch!

PR #132, opened 2019-08-23
Bad method removal

peteihis:
Removing a sick method. It was never called, so it did not cause any issues but it definitely should not be there.

I have no idea how that got there, but the most likely scenario I can think of is, that it was a left-over line from something else and I just patched without thinking to get it through the compiler  ... and then forgot.


Sailsman63:
Thanks!

PR #133, opened 2019-09-07
Boolean modelling tool: Option to hide the original pieces

peteihis:
A little something, that I have had on my mind for a long time.

By default the tool sets the parent objects hidden. If the user unchecks the option, their visibility is left unchanged.

Also, WindowClosing was set to perform the operation. I think the expectation would be to cancel it --> Now red-X = cancel.


ArtOfIllusion/src/artofillusion.properties
```diff
@@ -935,6 +935,7 @@ Union=Union
 Intersection=Intersection
 firstSecond=First - Second
 secondFirst=Second - First
+hideOriginals=Hide the original pieces
```
peastman: "Hide the original objects" would probably be clearer.

Sailsman63:
I like! Thanks.

PR #137, opened 2019-09-15
Language selection menu modernization

### LINKED ISSUE #136: Translation names in their own language?
Isin't it annoying, when you try to pick your own language from a list, that is written in a language, that you couldn't possibly know? That has happened to me. 

Nowadays it seems common to have language selection present each available choice in _that_ language. Definitely a user firendly approach. I don't know, what it would take but I suppose, substituting sometihng, that is now provided by the system with something else? -- And a fall back system, if the language does not exist in that particular computer?

peteihis:
- Available languages are presented in those languages.
- The menu is longer. It can show up to 20 items, but if there are more than 20 then it shows 16. The plan is to avoid the annoying one or two item scrollbars that you every so often see in AoI. IMO 20 items is beginning to get too long, but 16 is still quite comfortable to use.

Resolves #136 
Edit: Target release 3.1.1


peastman:
This look good to me.

peteihis:
One more thing:

Some of the Language selections appear with country and some not. What happens is that those that are coded with field constants that refer to _language_ come without the country note. US-English is marked by a constant that refers to _country_ and that appears with both language and country on the list.

It is obvious that US-English and Brazilian Portuguese need to be pointed out but AFAIK there are no notable variants of Finnish outside Finland or probably Africans outside South Africa. One possibility might be to modify the code so, that the country variant is mentioned only, when necessary but then it would be hard to draw the line... There are local written variants of German, the Finnish variant of Swedish etc...

I'd suggest to unify the coding to show county for all of the choices. With field constants it happens for example by changing `GERMAN` to `GERMANY`.

Any objections? I'll take, silence means acceptance and push a commit soon. :)

Country variants

Sailsman63:
I honestly don't see AOI having a separate english-GB or Portuguese-portugal translation. We have to few maintaining the translations that we have, and most strings are short, so grammer differences won't be prominent.

I'd favor removing the country codes from the display altogether.

I do like having each language displayed _in_ that language. Much more user-friendly.

peteihis:
I guess that makes sense. And I guess the empty _en could be removed?

peteihis:
So the empty _en could not be removed, it is there to keep the system sane. (I don't necessarily understand why, but I did not try to find out either.)

Locales are defined as they were but `PreferencesWindow` only shows the name. The _pt was renamed pt_BR, for clarity.

Sailsman63:
This looks good! Thanks!

PR #138, opened 2019-09-20
Creation depth of RootJoint corrected

peteihis:
The creation depth of the root bone was wrong. After this it is the working depth of the view as expected. Works correctly both in parallel an perspective modes.

This seems to have slipped through my fingers in the views update.

- Bug fix. 
- Target 3.1.1


Sailsman63:
Thaks for catching this!

PR #141, opened 2019-11-02
Splinemesh creation tool upgrade

peteihis:
Working on the primitives it caught my eye, that spline meshes are not shown during creation whereas primitives are.  Assuming, that the reason to that would have been processing power, when the code was written, I'd think those days are ancient history now. So here's what I got:

Changes:

- The object is drawn during creation
- Ctrl-down at MousePressed  produces centered drawing
- Shift as before
- Both keys together have combined effect
- UI reordred, so that shape selection is on top
- Shape selection chances suggested u and v values suitable for each shape.
- User given values for u and v are remembered per shape during session.

As I tested with u and v values, that I'd believe to be typical (up to 100 or so), I can't say for sure, that I would have seen any more delay, than when drawing primitives.


peteihis:
The last one is more of a philosophical update....

ArtOfIllusion/src/artofillusion/CreateSplineMeshTool.java
```diff
@@ -78,22 +82,58 @@ public String getToolTipText()
   @Override
   public void mousePressed(WidgetMouseEvent e, ViewerCanvas view)
   {
-    clickPoint = e.getPoint();
-    shiftDown = e.isShiftDown();
-    ((SceneViewer) view).beginDraggingBox(clickPoint, shiftDown);
+    clickPoint  = e.getPoint();
+    equilateral = e.isShiftDown();
+    centered    = e.isControlDown();
```
Sailsman63: This "Centered" is the one feature that I question. There does not seem to be an equivalent option for any of the other primitives, so not a matter of parity.
What is the intended utility? I'm having a hard time imagining where it fits in the workflow.
If we do end up keeping it, it needs to be documented in the window helptext.

Sailsman63:
Sorry for the delay. I finally got around to reviewing the details here. mostly looks good. See review comment for the one concern I have.

ArtOfIllusion/src/artofillusion/CreateSplineMeshTool.java
Sailsman63: Ahh... I see that #128 would add this to the other tools. I'm still wondering about workflow utility. One issue that I could see is that too many "Hidden" (Low-key, at least) Tool options can get to be as overwhelming (in a totaly different way) as too many buttons in your face.

peteihis: > I'm having a hard time imagining where it fits in the workflow.

Then, I have always _(or always since the first time I thought of it)_ found it annoying that when drawing a primitive it is always guess work, where exactly it end's up or where exactly to start dragging to draw a primitive where I wanted it. Usually either the size or the position ends up wrong. 

Most tools that have anything defined by dragging on screen have one option to start from center. Then it's in anybody's imagination where or whether to use it. I use those options pretty often. Many things are symmetrical by nature.

I'll admit that after playing with it a bit, it did at first feel like there were almost too many things happening through the modifier keys but I bet that after you use it for a while and find a few good use cases it won't be bad at all. And if you don't need the option just forget it. :) ALL software that I have ever used have plenty of options available that I have never used. 

I believe the perceived oddity is more just that it's a bit of a rarity to have the options available so easily: The usual UI is that you open a menu under the tool icon and select the center based option to draw _or_ that you open the tool settings dialog and check a box or two to draw. The what happens is that to do something else, you _(after first drawing with the last used settings and undoing what yo just did wrong)_ need to open the tool again and uncheck the same boxes, which just takes more time and disturbs the work flow.... Compared to that I'd say, this approach is extremely handy.

> it needs to be documented in the window help text.

Oh yes, I forgot the help texts! I'll add those to the PRs. Tomorrow or so.

Sailsman63: Y'know, after thinking about it a bit, I do believe you are right... I'm too used to just assuming that you create an object, and then move it to where it is needed.

I'll support this feature!

PR #144, opened 2019-12-22
NVIDIA crash bug tackle

peteihis:
Handling a bug in GL/Nvidia/Win10 environment that caused Java to crash. This is by no means a "fix" to the bug, that is outside our reach. This just prevents the unhandled error from happening by setting some initial values to an entity that the Nvidia drivers seem to expect to find.

Changes of #143 are included here, but this branch is derived from the master.

**Target 3.1.1**

### DIFF
--- a/ArtOfIllusion/src/artofillusion/view/GLCanvasDrawer.java
+++ b/ArtOfIllusion/src/artofillusion/view/GLCanvasDrawer.java
@@ -1,4 +1,6 @@
-/* Copyright (C) 2005-2017 by Peter Eastman
+/* 
+Copyright (C) 2005-2017 by Peter Eastman
+Modifications copyright (C) 2019 Petri Ihalainen
 
 This program is free software; you can redistribute it and/or modify it under the
 terms of the GNU General Public License as published by the Free Software
@@ -943,6 +945,11 @@ public class GLCanvasDrawer implements CanvasDrawer
           useTextureRectangle = false;
         }
       }
+
+      // Tackling Nvidia bug 2019
+      FloatBuffer fb = Buffers.newDirectFloatBuffer(1);
+      gl.glVertexPointer(2, GL.GL_FLOAT, 0, fb);
+      gl.glNormalPointer(GL.GL_FLOAT, 0, fb);
     }
 
     @Override
@@ -974,9 +981,9 @@ public class GLCanvasDrawer implements CanvasDrawer
       if (view.isPerspective())
         minDepth = view.getCamera().getDistToScreen()/20.0;
       else
-        minDepth = Math.min(-0.01, depthRange[0]);
+        minDepth = Math.min(0.0, depthRange[0]);
+      maxDepth = Math.max(minDepth, depthRange[1])+0.01;
       minDepth -= 0.01;
-      maxDepth = depthRange[1]+0.01;
       if (view.getTemplateShown())
         drawImage(template, 0, 0);
       gl.glClear(GL.GL_DEPTH_BUFFER_BIT);


Sailsman63:
This code does work okay on Linux for me, but I don't really understand _why_ it works... It's entirely possible that the driver is correct, and we've been miss-using open GL.

@peastman can you do a quick check to make sure this does not break Mac?
(Are we still using the old FFP from the early versions of openGL?)

peteihis:
> I don't really understand why it works... 

Well, there is absolutely nothing in those added lines that I would expect to be useful..... 

The lines did not help with null for the FloatBuffer or 0 for the size of it or <2 for the vector size, so obviously what ever is created by those lines has to exist. Though, if it/they were actually used, I'd expect an error (or at least strange behavior) since I used the same 'fb' both of the different pointers _('prepareBuffers()' in GLCD uses dedicated FloatBuffers)..._

peastman:
This looks reasonable to me, though I also don't see what code path would have led to it executing drawing calls without first preparing the buffers.  Stepping through it in a debugger might reveal something.

peteihis:
@ peastman:

If you mean, how it happened inside Art of Illusion, that's easy: In `GLCanvasDrawer` there are plenty of drawing methods like `drawBorder(), drawLine(), renderBox()`..., and to my surprise `renderMeshTransparent()`, that don't call `prepareBuffers()`. Buffers are not prepared in the `CanvasListener.init() either`. Instead `renderWireframe(), renderMesh(), renderBoxes(),... ` (that is, _the actual 3D-drawing methods_) do call `prepareBuffers()`. 

Now if it happens, that there is nothing on a particular view, that needs 3D-wireframe rendering or 3D-mesh rendering first, the view will try to at least `drawBorder()` without ever preparing the notorious buffers. Somehow that has not been a problem in the last 20 years but it started to become one this autumn.

It is a possibility, that a smaller bug (#143) could have covered this problem to some extent but I'm not buying that, because a) that fix has not been released yet and still crash cases were reported, and b) I was using a version with that fix for a while before the problem appeared to me.

peastman:
Yes, but `drawBorder()` doesn't do anything that uses vertex buffers.  So that shouldn't be a problem.

Sailsman63:
"stepping through in a debugger" are you thinking in java-land, or an actual native-side debugger?

peastman:
In Java.  It might be possible to identify the exact call that triggers the crash.

Sailsman63:
I'll see if I can duplicate the bug on Windows (I do have an NVidea card) and try to pinpoint where things go wrong...

peteihis:
> I'll see if I can duplicate the bug on Windows (I do have an NVidea card) and try to pinpoint where things go wrong...

If it does not happen with the original 3.1.0 release, you won't be able to duplicate it. In all cases, that it happened it was after an automatic update to Windows 10. So far the victims I know about are, Maksim, Jean-Claude, myself and Steve D,. (...and possibly Ivain, need to check my correspondence with him...)

I of course tested only to find what would prevent the crash, but several GL-calls were able to trigger it. When I commented out one, then the next one took over. I can send a list of my findings next week. (Did not take the test hack with me on holidays. :) )

Sailsman63:
It'd be nice to know which calls triggered the crash... What we need to see is if any of them are using vertex buffers under the hood...

I suspect that the NVidea driver itself is using a modern, shader pipeline to implement the old FFP, and forgot to initialise the buffers themselves.

peteihis:
It turned out to be slightly different from what I thought at first. The problem has nothing to do with any of the GL2 drawing calls. Instead in `CanvasListener.init()` the two lines:

~~~~
      gl.glEnableClientState(GL2.GL_VERTEX_ARRAY);
      gl.glEnableClientState(GL2.GL_NORMAL_ARRAY);
~~~~
need the    `gl.glVertexPointer()` and `gl.glNormalPointer()` calls to accompany them.

What is interesting is, that it does not matter, when in the program path they appear as long as they do.  I tried moving the hack lines to a couple of different places and actually even if I moved them as teh last thing in the `display()` method (after all the drawing instructions had been sent to GL2) they still prevented the crash.

Sailsman63:
It's possible, then, that this really is our bug, and some previous drivers let it through the cracks.. I'm merging now.

peastman:
That's an NVIDIA bug then.  Here's the documentation for `glEnableClientState()`:

https://www.khronos.org/registry/OpenGL-Refpages/gl2.1/xhtml/glEnableClientState.xml.

And here's what it says about `GL_VERTEX_ARRAY`:

> If enabled, the vertex array is enabled for writing and used during rendering when glArrayElement, glDrawArrays, glDrawElements, glDrawRangeElements glMultiDrawArrays, or glMultiDrawElements is called. See glVertexPointer. 

The vertex array shouldn't need to be set until you call one of those particular drawing functions.

PR #146, opened 2019-12-24
Converted manual to Sphinx format

peastman:
There's probably still a lot of cleanup that could be done, but for a first pass I think it's already looking  pretty good.  To build the manual, cd to the `manual` directory and type `make html`.  It will get put into `_build/html`.

If you want to experiment with different themes, it's set in conf.py.  Edit the line

```python
html_theme = 'pyramid'
```

Also see https://www.sphinx-doc.org/en/master/usage/theming.html for a description of the customization options for the built in themes.


peteihis:
A few comments:

- The html -> rst translation is now probably as good as it can be. I could have this as a starting point.

- How about creating a dedicated folder for the manual / manual-src? I think, the docs folder(s) are on the .gitignore list for a reason.

- Is it now possible to split long sections of the manual to smaller pieces? Shorter files would be easier to maintain but....<br>
Currently every file becomes a h1-level topic of it's own and the markup of headers does not follow any predefined convention. Instead the meaning of the markup is redefined per each file. _(The first found underlined header always becomes the model for h1 and the next one that is different from the first for h2 and so on... )_ This way we'd have to collect the already long files into even longer ones to keep the structure resonable.

- Then about the themes: Maybe not surprisingly,  we'll eventually have to create our own theme.<br>
While for exmple the pyramid looks quite descent, none of the built-in ones provide the kind of an UI, that I would expect to have on a large manual. Some third-party ones have some good features but so far I haven't come a cross the exact perfect combination. ;)

peastman:
> Currently every file becomes a h1-level topic of it's own and the markup of headers does not follow any predefined convention.

This is inherent in the format: every file has to contain at least one top level chapter.  I converted them with the convention of `###` for top level, `***` for second level, and `===` for third level.  But I then deleted the first header from some files where it really didn't make sense and led to strange numbering.

> Maybe not surprisingly, I think, we'll eventually have to create our own theme.

We certainly won't ever "have to" do that!  To me, it sounds like way too much work for very little benefit. :)  Most themes have a lot of customization options, in some cases even letting you provide your own style sheet.  I'd prefer to just pick one we like, perhaps customize it a little, then leave it at that.

peteihis:
I did not mean making one from scratch... :) Customizing and hopefully going copy/paste from some good ones. So far I haven't even tried to look at how those things work, but rather sieving the good good stuff from the not so good.

OOT: Somehow I ended up here https://peastman.github.io/statmech/index.html :D I have no idea how, it just noticed in on a browser tab.

peastman:
> OOT: Somehow I ended up here https://peastman.github.io/statmech/index.html :D I have no idea how, I just noticed it on an open browser tab.

Good book!

Sailsman63:
>But I then deleted the first header from some files where it really didn't make sense and led to strange numbering.

Perhaps a heads up that we need to think about the logical structure a bit...

I noticed that a lot of the images have two issues:

1. They have a wierd, orangy background/margin. (Possibly a built-in buffer to work better with the old format?)
2. They're lossy/not as crisp as I would like, especially for screenshots of menus/text dialogs. (Many are jpeg - I suspect that png would be a better choice for many of these shots)

I'm considering going through and re-screenshoting many, if not all, these images. Would you prefer iterating on this branch a bit, or should we just merge & do further work on top of master?

peastman:
The orange is part of the images.  That's the background color used in the current manual.  I thought about writing a script to remove it from them, but since a lot of them also have shadows blending into that color, it would be kind of complicated.  If you're up for it, redoing them as PNGs would be an improvement.

> Would you prefer iterating on this branch a bit, or should we just merge & do further work on top of master?

If there's no urgent changes, I think we can go ahead and merge.  That way anyone else who wants to make changes can create new PRs of their own.

peteihis:
> shadows blending into that color

And texts outside the image area and so on .... At the time the images were made, the usual tool for transparency was still gif-style "on/off" transparency and that does not work with small texts.

Actually even with alpha channel trasparency texts written directly on transparent background don't look good. What happens is, that the edges are blended to BOTH the background color AND transprency. -- To avoid this you need keep texts on their own layers and write on _the transparency mask_ rather than on the actual color layer. (My experience with GIMP)

peastman:
It's probably best to keep the images mostly opaque.  If we want to be fancy we can have transparent shadows.

Sailsman63:
Okay, we a can all work on it now.

@peteihis do you want to convert the image-handling documentation you created for the other version, or shall I?

peteihis:
> do you want to convert the image-handling documentation

Just did. Did not add it to the index page though.

PR #147, opened 2020-01-03
The rst-text for the new image management

peteihis:
This is in a separate file, that could be merged where needed. Images are in a folder of their own folder. 

There is also a folder named text_icon. My plan was to put there icons, that could be inserted into the text as 'characters'. Currently there is only a right-arrow in indigo color to replace the tinyish "&rarr;" (`&rarr;`). I used it in the text.

Somehow "text_icon" seems to be asking for renaming... "char-icon", "glyph", "text-symbol".. ???


Sailsman63:
Thank you! I was reviewing this earlier. (need to figure out where to put it still) Very to the point & understandable. I'll try to get this merge soon.

Sailsman63:
Okay, figured out where it should be for now.

I hope you don't mind that I tweaked the grammar just a bit. Does this still come across the way you intended?

Any other thoughts before I merge?

Sailsman63:
Not sure what I think about the right-arrow thing, but I'll leave it for now. We'll see how we like it going forward.

peteihis:
> Does this still come across the way you intended?

Looks good.

> Not sure what I think about the right-arrow thing, but I'll leave it for now. We'll see how we like it going forward.

The story behind it is that --> looks a bit too sketchy to me and `&rarr;` (&rarr;) looks usually too faint for it's purpose. In .rst it is very easy to include images like this, which I foud out doing the conversion. 

The arrow and what ever might be added later should of course fit the font that we'd be using, which is not necessarily the case with that particular arrow.... (Wonder if Mr. Yardstick would care to have a look.)

peastman:
> and \&rarr; (→) looks usually too faint

Did you try making it bold?

peteihis:
>>  and &rarr; (→) looks usually too faint

> Did you try making it bold?

The bold setting does not affect symbols, that are not actual letters, numbers or punctuation. Not in HTML or any other text editing. ...or at least I'm not familiar with a font where it would.

Besides I did not come up with a way to express "arrow" in rst. Just tested. Everything comes through as written text. Don't know if it's about the theme, my way of doing it or what, but also any attempts to set raw-html are failing.

Sailsman63:
There's a whole set of pre-defined shortcuts for things like this: [see](https://docutils.sourceforge.io/docs/ref/rst/definitions.html)

The one we want is in [isonum.txt](https://docutils.sourceforge.io/docutils/parsers/rst/include/isonum.txt):

`|rarr|` (or use a character picker, and insert the UTF-8 character directly)

For local reference, all of these files should be under `docutils.parsers.rst.include` of your Sphinx installation.

These _should_ respond to the em/strong tagging. If they do not, we need to look at the `.css` to make sure that they are not receiving special handling.

PR #149, opened 2020-01-09
Center click to see through hidden faces

### LINKED ISSUE #148: Modeling-Mode rotation.
Just noticed something:

When editing a triangle mesh, you can hide parts of the mesh so they don't get in your way. If you do so, the center-click to adjust rotation center still triggers on the nearest part of the total mesh, even if that part is hidden.

For some reason,  polymesh behaves better (triggers only on the visible parts of the mesh)

peteihis:
Resolves  #148 on AoI side. PolyMeshViewer can be updated later if necessary.

I tested doing just 

~~~~java
      if (v instanceof TriMeshViewer)
        hideFace = ((TriMeshEditorWindow)((TriMeshViewer)v).getController()).getHiddenFaces();
~~~~

in the ClickedPointFinder, but it does not look like good practice to me and it would limit the effect to TriangleMeshes only. Done the way it is now in the PR, it allows adding new object types without having to add each one of them individually to CPF.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/view/ClickedPointFinder.java
+++ b/ArtOfIllusion/src/artofillusion/view/ClickedPointFinder.java
@@ -1,4 +1,4 @@
-/* Copyright (C) 2016 by Petri Ihalainen
+/* Copyright (C) 2016 - 2020 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -15,6 +15,7 @@ import artofillusion.ui.*;
 import artofillusion.object.*;
 import artofillusion.*;
 import java.awt.*;
+import java.util.ArrayList;
 
 public class ClickedPointFinder
 {
@@ -23,8 +24,9 @@ public class ClickedPointFinder
   private int w, h;
   private Vec3 cameraOrigin, cameraZ;
   private boolean perspective;
-  
-  public ClickedPointFinder(){};
+
+  public ClickedPointFinder()
+  {};
 
   /** 
      Return the closest point on a surface of an object that is found under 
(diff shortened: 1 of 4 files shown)


Sailsman63:
Thanks for the very quick patch!

Unfortunately, I think I misunderstood the original Issue. Updating the bug report soon.

peteihis:
> I think I misunderstood the original Issue

No you did not. It was there.

That is not to say there could not be something else. I'd suggest to merge this fix (assuming it is acceptable) and make the other bug another issue. I have hunch that is might be in the RotateViewTool OR in some undetected interaction between ... something.

Sailsman63:
Okay, I'll test a little more.

With this patch, I was getting a bunch of `nullpointerexception`s spamming the terminal, though things seemed to be _functional_

peteihis:
> With this patch, I was getting a bunch of nullpointerexceptions spamming the terminal, though things seemed to be functional

I'll have another look. It uses null for the hide list whenever there is nothing marked hidden (the same practice as with canvas drawers) but I though I had handled it properly. Obviously I didn't.

Sailsman63:
I think if figured out the NPEs. Apparently, if you compile (java 8 compatible) code with jdk11, and then _run_ it on jdk8, you get some weird results.

peteihis:
The NPE:s should come from `ClickedPointFinder`. Which line? -- Or possibly `TriangleMaths`.

If this is J8 <--> J11 forth/back compatibility thing, then I'm not sure how to tackle it or if it is worth it really worth it, but it'd still be nice to know where the clitch is.

peteihis:
I did this into the same branch. It is a really a different issue but appears very similar.

Now the center clicks in model space actually hit the surface of the visible meshes. Hidden ones are ignored. Also the depth of the click is now correct in all combinations of local or scene coordinates and the entire scene shown or not. 

`ClickedPointFinder` was restructured quite a bit, which allowed some of the processing to be simplified.

ArtOfIllusion/src/artofillusion/view/ClickedPointFinder.java
```diff
@@ -36,52 +35,48 @@ public ClickedPointFinder()
...
-      hideFace = v.getHiddenFaces();
+      surface = info.getPreviewMesh();
+      hideFace = view.getHiddenFaces();
+      toContext = contextTransform(view, info);
 
-      for (int j = 0; j < rMesh.triangle.length; j++)
+      for (int f = 0; f < surface.triangle.length; f++)
       {
-        if (hideFace != null && hideFace[j])
+        if (hideFace != null && hideFace[f])
```
Sailsman63: I figured out where my exceptions were coming from.

This code assumes that `hideFace` is the same length as `surface`, if it exists at all. However, `surface` is a `RenderingMesh`, and `hideFace` is keyed on the control mesh. They only match if the mesh is in 'Smoothing:none' or 'Smoothing:shaded' mode. (Or if _all_ of the edges and vertices are 'Smoothness:0.0')

When they _don't_ match, `surface` will definitely be longer, and this line will throw `ArrayIndexOutOfBoundsException` and the click-to-center will do nothing until you change mesh mode, or show the entire mesh.

Probably want to use the control mesh instead.

peteihis: Hmm... The canvas dawers get a rendering mesh and the hidden faces list. I'll check again how they are used there.

peteihis: So, `hideFace[]` in a CanvasDrawer is not the same `hideFace[]` that appears in `TriMeshEditorWindow`

The correct hide list was produced by `drawSurface()`in `TriMeshViewer`. All I had to do was to lift it from the method to class level and use that one. Renamed the relevant parts.

peteihis:
Thanks for the finding, I think got it now. Squashed it all into one commit as the first one was basically written over by the second one any way. -- A couple of force pushes before I had it all cleaned up. :)

peteihis:
And got that too.

Now most of the `ClickePointFinder` seems to be rigged with `view instanceof ObjectViewer && ((ObjectViewer)view).somethig....`, which does not look like very good programming to me. 

It might be best split into different classes for different viewers or at least have different constructors for different hosts. But not now, I'm afraid.

Sailsman63:
Thanks! This one works as expected for me.

Not sure that it needs to be broken out into classes. Maybe a couple of private helper methods? Not my first concern at the moment. There are only five `instanceof ObjectViewer` in there. I kind of want to stew on that for a bit.

Sailsman63:
Merged as-is for now. If we want to do a refactor, it can be a separate project.

PR #150, opened 2020-01-26
3D-Manipulator improvements

peteihis:
Here are some improvements to the Compound3DManipulator.
- Handle distance to the axes stays constant regardless of the size of the manipulator
- UV-directions are now U to right and V up
- NPQ mode renamed PQN mode. _The N stands for normal in model coordinates and object-Z in scene._
- In parallel viewing the move follows the handle correctly also, when there is depth movement included.
- In perspective mode the move and the manipulator tend to get ahead or be left behind of the mouse depending on the relative direction. To have it precisely follow the mouse looks like a very complex process. _(I tested a bit but I'm literally back at the drawing board with it)_
- <strike>In perspective mode it is not a 100% success otherwise either but hopefully better than before.</strike> Actually it works as intended.
- Removed some unused or redundant code but there is more in that department to do. 

If you find something wrong, let me know.  On my behalf this could be merged as is but I may push updates too as soon as I get any done.

Future plans: 
- Drawing order logic. I know how, but this tempts me to get into a pretty thorough remake of a lot of other things first.
- Closer look at the data structure in general. (Like get rid of the `Manipulator` interface, which is not designed for a 3D-manipulator.)

### DIFF
--- a/ArtOfIllusion/src/artofillusion/MoveScaleRotateMeshTool.java
+++ b/ArtOfIllusion/src/artofillusion/MoveScaleRotateMeshTool.java
@@ -69,7 +69,7 @@ public class MoveScaleRotateMeshTool extends MeshEditingTool
   public void drawOverlay(ViewerCanvas view)
   {
     BoundingBox selectionBounds = findSelectionBounds(view.getCamera());
-    if (!dragInProgress && manipulator.getViewMode() == Compound3DManipulator.NPQ_MODE && selectionBounds != null)
+    if (!dragInProgress && manipulator.getViewMode() == Compound3DManipulator.PQN_MODE && selectionBounds != null)
     {
       // Calculate the axis directions.
 
@@ -89,7 +89,7 @@ public class MoveScaleRotateMeshTool extends MeshEditingTool
       else
         updir = avgNorm.cross(Vec3.vy());
       updir.normalize();
-      manipulator.setNPQAxes(avgNorm, updir, avgNorm.cross(updir));
+      manipulator.setPQNAxes(updir.cross(avgNorm), updir, avgNorm);
     }
     manipulator.draw(view, selectionBounds);
     if (!dragInProgress)
@@ -210,7 +210,7 @@ public class MoveScaleRotateMeshTool extends MeshEditingTool
       if (mode == Compound3DManipulator.XYZ_MODE)
         manipulator.setViewMode(Compound3DManipulator.UV_MODE);
       else if (mode == Compound3DManipulator.UV_MODE)
-        manipulator.setViewMode(Compound3DManipulator.NPQ_MODE);
+        manipulator.setViewMode(Compound3DManipulator.PQN_MODE);
       else
         manipulator.setViewMode(Compound3DManipulator.XYZ_MODE);
       theWindow.updateImage();
--- a/ArtOfIllusion/src/artofillusion/MoveScaleRotateMeshTool.java
+++ b/ArtOfIllusion/src/artofillusion/MoveScaleRotateMeshTool.java
@@ -184,12 +184,12 @@ public class MoveScaleRotateMeshTool extends MeshEditingTool
     int selected[] = controller.getSelectionDistance();
     Vec3 v[] = new Vec3 [baseVertPos.length];
     for (int i = 0; i < v.length; i++)
-      {
-        if (selected[i] == 0)
-          v[i] = transform.times(baseVertPos[i]).minus(baseVertPos[i]);
-        else
-          v[i] = new Vec3();
-      }
+    {
+      if (selected[i] == 0)
+        v[i] = transform.times(baseVertPos[i]).minus(baseVertPos[i]);
+      else
+        v[i] = new Vec3();
+    }
     if (theFrame instanceof MeshEditorWindow)
       ((MeshEditorWindow) theFrame).adjustDeltas(v);
     for (int i = 0; i < v.length; i++)
--- a/ArtOfIllusion/src/artofillusion/MoveScaleRotateObjectTool.java
+++ b/ArtOfIllusion/src/artofillusion/MoveScaleRotateObjectTool.java
@@ -141,18 +141,24 @@ public class MoveScaleRotateObjectTool extends EditingTool
       int dx = dragPoint.x - clickPoint.x;
       int dy = dragPoint.y - clickPoint.y;
       if (e.isShiftDown() && !e.isControlDown())
-        {
-          if (Math.abs(dx) > Math.abs(dy))
-            dy = 0;
-          else
-            dx = 0;
-        }
+      {
+        if (Math.abs(dx) > Math.abs(dy))
+          dy = 0;
+        else
+          dx = 0;
+      }
       Vec3 v;
       if (e.isControlDown())
         v = view.getCamera().getCameraCoordinates().getZDirection().times(-dy*0.01);
       else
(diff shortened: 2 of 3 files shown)


Sailsman63:
I've been wondering about some tweaks here for a while... I'll review later.

peteihis:
I have been on and off this for a long time. One problem here is that it is really hard to restrict changes just to the thing that you try to do. Small changes require somethign else and then that something else looks like it could be done better...

A couple of more things:

- The 3D-manipulator ignores `ObjectInfo.isLocked()`. The `MoveObjectTool` does not check `ìsLocked()`  but it does not move locked objects either. I don't know yet whee the difference is but the check couldbe added to "MoveScaleRotateObjectTool".

- One more thing to change is to make rotating natural. The current implementation is a kind of brilliant in its  simplicity but it has been outdated since it was new. _(It might be handy for small rotating knobs on that you expect to turn several full rotations at a time but it definitely does not belong to a 3D-manipulator.)_

Sailsman63:
>make rotating natural

How would you like this to work? Despite #152, I find rotating very natural, especially when working with meshes. Don't be afraid to upload a sketch/picture of what you have in mind.

Sailsman63:
This is looking better.

One thing I noticed when selecting an object in the Layout Window: If there is a perspective/camera view that has not had it's mode changed, the manipulator is rendered much too big. (You only see the center and a bit of the axis markers)

I'm not sure that this is due to _this_ change, as the old version acts weird in these circumstances as well.

Will probably merge tomorrow.

peteihis:
> the manipulator is rendered much too big.

In perspective mode the manipulator is now rendered in its default size in the current working depth of each view. The size or the "inner spere" (= rotation hadles) obeys the scaling of the perspective and the move and scale handles are kept at fixed distances according to screen scale.

The released version takes perspective scaling to account but ignores the different view angle and its effect on scaling.

One problem is that if you select the bound Camera of the view or an object that is located in the same psition OR even behind the camera,  the manipultorr is still drawn on that view. That should not happen as in that case you obviously can not use the manipualtor in any sensible way.

I tested a quick mechanism to block drawing in those case but got it wrong the first time. It is on my "list" (just in my head relly) though.

I got the drawing order logic done. I'll push in the afternoon, but I'm tempted to squash all the stuctural and visible the chanes into one commit as it is really impossible to keep the content of each single commit sensible. You always come up with something that seems to belong two or three commits before. (Can alwasy amend to the the previous one that is still open, but not before that....)

Sailsman63:
>You always come up with something that seems to belong two or three commits before. (Can alwasy amend to the the previous one that is still open, but not before that....)

There are ways to do it. If you find more white-space/format changes, make them in their own new commit. I can use an interactive re-base to reorder them, and then combine them into the correct logical commits.

peteihis:
> There are ways to do it. If you find more white-space/format changes, make them in their own new commit. I can use an interactive re-base to reorder them, and then combine them into the correct logical commits.

It is not about white spaces.

To have the code make better sense I needed to rename things, make some local variables class wide, delete some entirely, write out loops, create loops .... If and attempt was made to move a change back in history, there would would be places where it should affect but it wouldn't know to and places to affect that would not be there yet.  I don't see automatic (or assisted) rebasing as a realistic option here. 

Making code changes, when a lot of things need attention is a bit like painting a picture.... No stroke of the brush is ready alone. What needs to be done first is often revealed only after you manage to get a version of the actual goal working. ... And the preparations that might be needed don't necessarily make much sense without seeing what they are for,

peteihis:
> How would you like this to work?

Currently the rotation direction turns back if I try to rotate the handle around the center. I know how it works, but dragging to one direction to rotate does not feel right to me. In my expectation it would be WYSIWYG. WYSIWYG rotation was also the case with my day-job-tools from 2000 to 2016. In my current SW a 3D-tool does not exist and that sucks! :)

[Rotate.zip](https://github.com/ArtOfIllusion/ArtOfIllusion/files/4188248/Rotate.zip)

Sailsman63:
I think that we misunderstood each other regarding "Drawn Too Big"
This is what I mean:
Manipulator_Error

I've found that it only happens (so far) when you open a file that has the camera too close to the object already, (clipping plane starting to cut into it) so probably not that big a deal for the moment. If you have an idea off the top of your head of why this is happening, great. If not, we can skip it for now.

RE: rotating - Now I see what you mean. Not sure how I feel about that/how much it would take to make it work. definitely for another time, however.

peteihis:
>  it only happens (so far) when you **open a file** that has the camera too close to the object already

That was the key word! ... And there are a couple of things that get it back on track.

- Change the view to another SceneCamera
- Change to another view preset and **then back** to the camera
- Use FIt View to Selection

The scaling system should handle Scene Camera like business as usual but obviously something is now missed, when the file is just opened and there is no prior reference for scaling with SceneCamera. What I can tell is that the entire scaling is wrong, not just the size control of the center.... what ever it should be called... 'inner layer of elements'?: The move/scale handles too are too far from the axis tips. I'll try to figure this out. I would not expect it to be too difficult as the ingredients of the fix (as it turned out out) are already in the code....

> I'm not sure that this is due to this change

It is. The old manipulator (on the same file) appears as the tiny ball of yarn it always has....

> How much it would take to make it work. 

Not that much really. I tested with a quick prototype _(that was still missing view angle compensation of mouse drag)_. What caught my eye was that in that implementation the rotation feedback disc needs to be drawn from the point where you grab the handle. Otherwise the effect feels a bit like driving from passenger's seat.... And by nature (= using `Math.atan2()`)  it of course only draws 180° of the feedback before the disc flips side. The rotation angle (for the feedback disc at least) should be calculated cumulatively up to `Math.PI*2`, so it feels more steady.

Sailsman63:
>The scaling system should handle Scene Camera like business as usual but obviously something is missed, when the file is just opened and there is no prior reference for scaling with SceneCamera. 

I've been reviewing the code a bit, and I'm coming to the (tentative, preliminary) feeling that the entire approach to scaling in this file takes is... overly-complicated?

It certainly works, in the main, and I appreciate the effort that you've put into it. It's an area where I'd be looking at possible refactors if we ever get around to re-doing how view-decoration stuff is applied.

peteihis:
Found & fixed and fixed another one with SceneCamera too, that I had not checked earlier.

The bug that supersized the manipulator was not in the manipulator but somewhere else. However it was not necessary to use the faulty value here and I don't think it's used anywhere else either, so can be fixed separately some other time.

>  the entire approach to scaling in this file takes is... overly-complicated?

I know what you mean.... Unfortunately, with the current code base, you are looking at the minimum amount of scaling code. This same thing is happening in sub classes of ViewerCanvas and in mouseDrag methods but the manipulator is a special case in the sense that scaling happens first from screen to the 3D-space and then back, so there is (a kind of) double amount of code.

The amount of special cases can be reduced and I have a pretty good idea how but let's get back to that after the next release.

peteihis:
One more short coming of the manipulator:

TriMesh
PlyMesh

The AoI version takes a short cut in determining the surface normal direction. It always uses the average of the vertex normals in the selection. The PME one uses normals of their own for edges faces an vertices. Of course the AoI manipulator needs to handle all all different kinds of meshes. To have them all treated according to their special needs might even extend the object tool to sub classes but this is something to think about.

As said, not now. :)

Sailsman63:
Thanks!
RE: Selection Normal - Hmmm... Never noticed that before. Maybe open a new issue? I have a theory about the "Clean" way to handle that, but it would definitely be an API version issue.

PR #151, opened 2020-02-01
Optimized deleting parts of a mesh

peastman:


Sailsman63:
Definitely makes a big difference!

PR #153, opened 2020-02-16
Improved performance of beveling meshes

peastman:
I also did a bit of code cleanup while I was at it, including replacing tabs with spaces.  Tell it not to show whitespace changes to get a better idea of what changed.


Sailsman63:
Working great!

This one seems to have made a greater difference than the Deletion case. (Tested on the same mesh)

PR #157, opened 2020-02-23
Normal direction of surface text

### LINKED ISSUE #140: Surface text sides inverted
Just recently I noticed, that when you use the Text tool in Surface mode, and put a layered texture on it, Front and Back are opposite to, what you'd expect.

peteihis:
Resolves  #140.


Sailsman63:
Thanks!

PR #161, opened 2020-02-28
CreatePolygonTool upgrade

### LINKED ISSUE #156: Match polygon drawing tool modifier keys with primitives and spline mesh
In the Polygon tool holding Ctrl draws the polygon filled.

To have consistency with  #128 and #141 I'd suggest to put 'Expand from center' under the Ctrl in Polygon too and move the choice for filled drawing into the dialog. 

The dialog of the polygon tool is probably opened most of the times anyway, when the user wants to draw a polygon, so I'd say the choice would be in an obvious place there. And personally I have recently become to appreciate it very highly that tools that do somehow similar things are handled exactly the same way. (One of the things whose value you notice, when it's missing.)

peteihis:
Upgraded to work with the same logic as the primitives and Spline Mesh. Resolves #156.

- Draws the unsmoothed shape during drag for reference as overlay.

Let me know if you find anything funny, the modifications were prettey heavy.


ArtOfIllusion/src/artofillusion/CreatePolygonTool.java
```diff
@@ -66,155 +71,220 @@ public String getToolTipText()
...
     cosine = new double [sides];
     minsine = mincosine = 1.0;
     maxsine = maxcosine = -1.0;
-    for (i = 0; i < sides; i++)
+    for (int i = 0; i < sides; i++)
     {
-      sine[i] = Math.sin((i+0.5)*2.0*Math.PI/sides);
-      sine[i] = 1e-10*Math.round(sine[i]*1e10);
+      sine[i]   = Math.sin((i+0.5)*2.0*Math.PI/sides);
+      sine[i]   = 1e-10*(Math.round(sine[i]*1e10));
```
Sailsman63: Only thing I've done so far: you've got a redundant pair of parentheses here.

ArtOfIllusion/src/artofillusion/CreatePolygonTool.java
```diff
@@ -66,155 +71,220 @@ public String getToolTipText()
...
+      xscale = yscale = Math.max(xscale, yscale);
 
-    // Find the object's coordinate system.
+    Vec3[] vertex = ((Mesh)(objInfo.object)).getVertexPositions();
+    for (int i = 0; i < sides; i++)
+    {
+      vertex[i].x = sine[i]*xscale;
+      vertex[i].y = -cosine[i]*yscale;
+    }
+    ((Mesh)(objInfo.object)).setVertexPositions(vertex);
```
Sailsman63: And another here. `objectInfo.object` does not need to be wrapped.

ArtOfIllusion/src/artofillusion/CreatePolygonTool.java
Sailsman63: Line 83

Sailsman63:
I've come down on the side of consistency. I think we'll be better off with this than without.

ArtOfIllusion/src/artofillusion/CreatePolygonTool.java
peteihis: Oops. I tried somethig different here, this happened at returning the original.

ArtOfIllusion/src/artofillusion/CreatePolygonTool.java
peteihis: Thanks.

peteihis:
Fixed.

PR #165, opened 2020-03-08
Oss plugin update to work both with JDK 9 and older

makiam:


Sailsman63:
Where does the file `OSSpecific/lib/orange-extensions-1.3.0.jar` come from?

makiam:
https://ymasory.github.io/OrangeExtensions/

peastman:
This seems like quite a lot of new code for something that ought to be trivial...  There's only a single block of seven lines that's specific to the JDK8 API.  Why not just check the JDK version, then execute either those lines or a similar set of lines that use the JDK9 API?

makiam:
This code is not called from AOI code but registered and called as callback by OSX. So my idea is try to create JDK8 code and if it failed try to create JDK9 code.

Classes under src-ext is intended for compile JDK9 code with JDK8

Sailsman63:
So most of it is a compile-time dependency?
I did see that the built jar is *significantly* larger.

I don't see a need to include the Orange-Extensions. As long as we support JDK-8, any official binaries will be *built* on JDK-8, which has the apple classes included.

(This prevents issues with method calls into the standard library being resolved to newer, more specific versions that are not in the older version of JDK. This is a real thing that I've run into, mostly in the context of the JOGL libraries)

Also, what is the purpose of `DefaultPluginImplementation`? Looks more like a case for default methods in the Interface.

@peastman Due to the way the JDK API was constructed, We do need to use this callback style to get anything done.

Our other option is to just say that the "nice" stuff for mac should wait until we move on to java 12...

OSSpecific/src/artofillusion/osspecific/MacOSPlugin.java
```diff
@@ -11,98 +11,107 @@ Changes copyright (C) 2017 by Maksim Khramov
...
+import buoy.widget.MenuWidget;
+import buoy.widget.Widget;
+import java.awt.Color;
+import java.awt.Font;
+import java.awt.Rectangle;
+import java.io.File;
+import java.lang.reflect.InvocationTargetException;
+import java.util.prefs.PreferenceChangeEvent;
+import java.util.prefs.PreferenceChangeListener;
+import java.util.prefs.Preferences;
```
Sailsman63: Please do not expand package wildcard imports into these overly long lists of specific classes. As a rule of thumb, if you need more than two or three classes, import the whole package.

(This is an AOI style thing - other projects may differ)

makiam: Resolved

makiam:
> Also, what is the purpose of DefaultPluginImplementation? Looks more like a case for default methods in the Interface.

I have an idea to update other plugins to use this class. Extend and override only needed event handlers looks more object-oriented style for me vs implementing big switch method...

> I don't see a need to include the Orange-Extensions. As long as we support JDK-8, any official binaries will be built on JDK-8, which has the apple classes included.

We need it if we want to build AOI not only on OSX platform with only JDK8 where native implementation of this classes available

> I did see that the built jar is significantly larger.

I add only 4 new classes  (3 adapter classes + DefaultPluginImplementation). Some classes produces couple of anonymous classes at build. DefaultPluginImplementation is to be build with core application

Sailsman63:
>I have an idea to update other plugins to use this class. Extend and override only needed event handlers looks more object-oriented style for me vs implementing big switch method...

Then, that's a refactor/discussion for another time. It does *not* belong in this PR.

>We need it if we want to build AOI not only on OSX platform with only JDK8 where native implementation of this classes available

The legacy implementation did not have this limitation. (Used reflection) I'm not sure why you built a JDK-8 adapter, as it's kind of deprecated anyway. Probably best to nest the existing code inside a `try` and if if fails, move on to the new way.

makiam:
I borrow this solution from NetBeans sources repository. They used OE library for compile time references for JDK8 code and some classes from JDK9 to build JDK9 code with JDK8.

Sailsman63:
I think that you are missing my point. We have a perfectly good solution for jdk8 that does *not* require the compile-time binary. Why are we moving to a different solution there?

I totaly get that we will need the jdk9 reference classes for the *Desktop* method.

makiam:
IMO using OE library gives us more clear and easy to understand code.
What is the significant reason not to use library code at compile time?

Sailsman63:
Primarily, you're adding *another* dependency (to support a new way to do a job that was already done with internal code.) All of our existing dependencies meet needs that we cannot fulfill easily internally, this one just adds to the chain for cleaning up a *very* little bit of code.

The new code will probably have is own quirks. (I've spotted one bug that will only trigger on Mac w/jdk8)

You've also created an entire new API level just to support a *single* OS/jdk version combination that is going to be less and less relevant as time goes on. If we were actively developing more functionality for this plugin, and needed it to work on jdk 8 & 9+, it might be different.

As it is, I'd prefer something more like: 

Combine DesktopAdapter & DesktopAdapterJDK9 into a single class.

main code on application startup:

```
if (Runtime.class.getPackage().getImplementationVersion() < 9 && os == mac)
{
  // legacy reflective code goes here. Maybe _some_ cleanup if that makes it more readable
} else if (Runtime.class.getPackage().getImplementationVersion() >=9)
{
   //load and register the desktop adapter
}
```

Sailsman63:
By the way, I do think that the `DefaultPlugin` idea has merit. However, there I think there is a better way to do it, which is why I want that change pulled out and discussed separately.

OSSpecific/src/artofillusion/osspecific/DesktopAdapterJDK8.java
```diff
@@ -0,0 +1,55 @@
...
+import com.apple.eawt.PreferencesHandler;
+import com.apple.eawt.QuitHandler;
+import com.apple.eawt.QuitResponse;
+
+
+/**
+ *
+ * @author maksim.khramov
+ */
+public class DesktopAdapterJDK8 extends DesktopAdapter implements AboutHandler, OpenFilesHandler, PreferencesHandler, QuitHandler {
```
Sailsman63: Line 22 can't work. The `implements` clause will throw `ClassNotFound` X4 if we try to load on JDK 8 - the interface classfiles need to be present to load correctly.

OSSpecific/src/artofillusion/osspecific/MacOSPlugin.java
```diff
@@ -13,96 +13,86 @@ Changes copyright (C) 2017 by Maksim Khramov
...
-    win.getComponent().getRootPane().putClientProperty("Window.documentModified", win.isModified());
-    Scene scene = win.getScene();
-    if (scene.getName() != null)
-    {
-      File file = new File(scene.getDirectory(), scene.getName());
-      win.getComponent().getRootPane().putClientProperty("Window.documentFile", file);
-    }
+    view.getComponent().getRootPane().putClientProperty("Window.documentModified", view.isModified());
+    Scene scene = view.getScene();
+    if(null == scene.getName()) return;
```
Sailsman63: Please don't use void early-return like this.
`if(guard) return;` really messes with how source is read.

When I see an if, I'm expecting

```
if(okaytowork)
{
  Do stuff;
}
```

OSSpecific/src/artofillusion/osspecific/DesktopAdapterJDK8.java
makiam: Right. And I catching this on plugin onApplicationStarting code to fallback to new JDK9 implementation...

Sailsman63: How? These interfaces were added as part of Java 9... Even marked so in the source.

makiam: No. Apple specific com.apple.eawt.* classes available in JDK8 on Mac platform and dropped and replaced with java.awt.Desktop in JDK9. The new JDK9 interfaces from java.awt.* in DesktopAdapter9 has same names but other method parameters.
The OE library allows us to see and use this com.apple.eawt.* on other platforms and/or JDK9+

Sailsman63: Ahh...

peastman:
> Due to the way the JDK API was constructed, We do need to use this callback style to get anything done.

We can still do it with reflection using `java.lang.reflect.Proxy`.  Let me see if I can put something together.

PR #166, opened 2020-03-08
Minimum zoom distance

peteihis:
This prevents zooming into "crash zone".

`Fit View to Selection` selection, when the selected object(s) return a zero-sized bounding box would blank the view. Now the minimum distance from Camera to drawing plane is set to 1.0E-12.  Surprisingly with this magnification and the view set to parallel mode you can still create primitives and rotate the view normally. Some display modes (transparent and shaded) don't work correctly any more though.

The limit is set to the scroll wheel zoom and the magnification field of the ViewerCanvas. It is _possible_ to scroll wheel to 1.0E-12 distance of a point (= 2.0E15 scaling) but it _does_ take some patience....

### DIFF
--- a/ArtOfIllusion/src/artofillusion/view/ViewerScaleControl.java
+++ b/ArtOfIllusion/src/artofillusion/view/ViewerScaleControl.java
@@ -1,5 +1,5 @@
 /* Copyright (C) 2007-2009 by Peter Eastman
-   Modifications copyright (C) 2017 Petri Ihalainen
+   Modifications copyright (C) 2017-2020 Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -25,7 +25,6 @@ public class ViewerScaleControl implements ViewerControl
 {
   private ViewerCanvas view;
   private ValueField scaleField;
-  private double newScale;
 
   @Override
   public Widget createWidget(final ViewerCanvas view)
@@ -35,6 +34,7 @@ public class ViewerScaleControl implements ViewerControl
     this.scaleField = scaleField;
     scaleField.setText("100");
     scaleField.setMinDecimalPlaces(1);
+    scaleField.setValueChecker(new ZoomChecker(view));
     view.addEventLink(ViewChangedEvent.class, new Object() {
       void processEvent()
       {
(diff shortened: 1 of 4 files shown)


peteihis:
Added the safety to two more methods. One deprecated and one whose calling methods already do the check.

ArtOfIllusion/src/artofillusion/view/ViewerScaleControl.java
```diff
@@ -88,4 +96,34 @@ public String getName()
...
+  {
+     PerspectiveChecker()
+     {}
+
+     @Override
+     public boolean isValid(double value)
+     {
+       return value >= ViewerCanvas.MINIMUM_ZOOM_DISTANCE;
+     }
+  }
```
Sailsman63: Could you show me what the "Checker" code looked like when it was misbehaving? I think that you are running into some inner/outer class fine points. This *should* work with just a single checker.

To properly use OuterClass variables, the InnerClass constructor must be something like `zoomChecker(ViewerScaleControl control)` and then line 41 or so would be:

`scaleField.setValueChecker(new zoomChecker(this));`

Then you would call `control.view.isPerspective()` inside the checker.

peteihis: It would have looked, basically what I just pushed but without arguments in the constructor and the view of the `ViewerScaleControl` set public (or default) -- (not the `view` inside the widget). Probably had `else` there but I removed that. 

I only set the ViewerCanvas to the checker, that's all that it needs.

peteihis: Thanks. :)

Sailsman63: looks much better now!

peteihis:
I remembered one more tool that needed this.

To test:
- Draw a cube
- Edit all dimensions of it to about 8E-14
- Fit View to that.
- Zoom in and out in all the various ways
  - In perspective mode you won't see it because of the cut off distance. Only the magnification value field responds.

Sailsman63:
Good catch! I've never actually managed to create any of these crashes... You seem to have a talent for finding them (or a different approach to your workflow)

PR #167, opened 2020-03-12
Default empty plugin implementation

makiam:


Sailsman63:
Two Points:

1. While this does look cleaner from a certain perspective, is it really enough of an improvement to merit adding the code? Of the core plugins, only OSSpecific uses this interface. Of the other plugins that I have checked, only a couple use it. The API is not exactly a source of huge code bloat.
2. To use this as written, you are forced to extend this base class. That may not be what you want to do. If we go for this concept, I would be very tempted to implement it as default methods within the `Plugin` interface itself. (Seems controversial in some circles - I this case, it would prevent us from using the interface as a 'functional interface,' but I'm not sure we have a use case for that...)

Sailsman63:
On further thought (revisiting, I know it's been a long time) I'm going to say that this doesn't make a sufficient difference at this time.

makiam:
Sounds like never (((

Issue #163, opened 2020-02-28
NPE from AutoScroller


makiam:
Steps to reproduce:
Open AOI.
Click on Objects pane with RMB to invoke popup menu.
Click on pane again. The NPE is outputs in log:

java.lang.NullPointerException
	at artofillusion.ui.AutoScroller.mouseReleased(AutoScroller.java:51)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
	at buoy.internal.EventLinkRecord.dispatchEvent(EventLinkRecord.java:81)
	at buoy.event.EventSource.dispatchEvent(EventSource.java:140)
	at buoy.widget.Widget.dispatchEvent(Widget.java:326)
	at buoy.internal.EventLinkAdapter.dispatchEvent(EventLinkAdapter.java:212)
	at buoy.internal.EventLinkAdapter.mouseReleased(EventLinkAdapter.java:108)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6578)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3343)

I see that mouseReleased method calls twice and on second call tries to interrrupt already released scrollThred.
Easy to fix but not very clear why this method calls twice...

Sailsman63:
Can you clarify your instructions here? I have not been able to reproduce this exception...

peteihis:
I have seen this a lot, but I did not pay attention to how it happened. 

1) Click Objects Pane with Right MouseButton **on the empty the area past the object list**.
2) Click again with any button.

This pops the alert up. As a new symptom the NullPointer some times froze the UI entirely but that could be related to me trying to capture the event on video.

makiam:
null object is AutoScroller scrollThread  which already nulled

peteihis:
> null object is AutoScroller scrollThread which already nulled

[...]

PR #168, opened 2020-03-12
Fix NPE from autoscroller

makiam:
Issue #163 fix


ArtOfIllusion/src/artofillusion/ui/AutoScroller.java
```diff
@@ -48,8 +49,11 @@ private void mousePressed(MousePressedEvent ev)
 
   private void mouseReleased(MouseReleasedEvent ev)
   {
-    scrollThread.interrupt();
-    scrollThread = null;
+    if(null != scrollThread)
+    {
```
Sailsman63: I'd rather have this as `if(scrollThread != null)`

But otherwise looks good.

Sailsman63:
Thanks!

PR #170, opened 2020-03-15
MacOSPlugin works with JDK9+

peastman:
This is a simpler alternative to #165.

### DIFF
--- a/OSSpecific/src/artofillusion/osspecific/MacOSPlugin.java
+++ b/OSSpecific/src/artofillusion/osspecific/MacOSPlugin.java
@@ -1,4 +1,4 @@
-/* Copyright (C) 2002-2011 by Peter Eastman
+/* Copyright (C) 2002-2020 by Peter Eastman
    Changes copyright (C) 2017 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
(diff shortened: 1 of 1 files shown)


OSSpecific/src/artofillusion/osspecific/MacOSPlugin.java
```diff
@@ -40,15 +40,35 @@ public void processMessage(int message, Object args[])
...
+        {
+          // Use the Desktop API introduced in Java 9.
+
+          appleApi = false;
+          Class aboutHandlerClass = Class.forName("java.awt.desktop.AboutHandler");
+          Class openHandlerClass = Class.forName("java.awt.desktop.OpenFilesHandler");
+          Class preferencesHandlerClass = Class.forName("java.awt.desktop.PreferencesHandler");
+          Class quiteHandlerClass = Class.forName("java.awt.desktop.QuitHandler");
+          Object proxy = Proxy.newProxyInstance(Desktop.class.getClassLoader(), new Class[]{aboutHandlerClass, openHandlerClass, preferencesHandlerClass, quiteHandlerClass}, this);
+          Desktop desktop = Desktop.getDesktop();
```
makiam: > Desktop desktop = Desktop.getDesktop();

How to expect compile this with JDK8? If compile this with 9 what the reason set handlers via proxy?

Sailsman63: The `Desktop` class itself has been around since Java 6, so this will compile. For the time being, or official builds will be made with Java 8, but we can't guarantee that it will be run with 8.

peastman:
I tested this code on both JDK 8 and JDK 13.

Sailsman63:
And I tested on both Linux and Windows, so we're good!

PR #171, opened 2020-03-17
Update class lookup and instantiating with varargs style

makiam:
More array to varargs style call changes

### DIFF
--- a/ArtOfIllusion/src/artofillusion/procedural/Module.java
+++ b/ArtOfIllusion/src/artofillusion/procedural/Module.java
@@ -291,7 +291,8 @@ public class Module
    * procedure at a new point.The default implementation does nothing.
    * Subclasses whose output depends on the point should override this method.
    *
-   * @param point
+   * @param point - a point in space for which a module value is being evaluated.
+   * 
    */
   
   public void init(PointInfo point)
--- a/ArtOfIllusion/src/artofillusion/procedural/Module.java
+++ b/ArtOfIllusion/src/artofillusion/procedural/Module.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 2000-2011 by Peter Eastman
+   Changes copyright (C) 2020 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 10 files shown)


peastman:
Nice, definitely cleaner that way.

ArtOfIllusion/src/artofillusion/procedural/Module.java
```diff
@@ -285,11 +286,15 @@ public boolean edit(BFrame fr, Scene theScene)
...
 
-  /** This method initializes the module in preparation for evaluating the procedure at a
-      new point.  The default implementation does nothing.  Subclasses whose output depends
-      on the point should override this method. */
+  /**
+   * This method initializes the module in preparation for evaluating the
+   * procedure at a new point.The default implementation does nothing.
+   * Subclasses whose output depends on the point should override this method.
+   *
+   * @param point
```
Sailsman63: @param annotations in the javadoc should add information beyond what can be discerned from the type signature. Non-obvious use details, etc.

Tests/src/artofillusion/procedural/ModuleTest.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * To change this license header, choose License Headers in Project Properties.
+ * To change this template file, choose Tools | Templates
+ * and open the template in the editor.
+ */
```
Sailsman63: Please do add a proper license header when creating a new file. It helps with clarity, and I've heard that in some jurisdictions, lack of a license is interpreted in such a way that @peastman or I could (theoretically) face legal difficulties if we added the file to the project.

Sailsman63:
I do like this style for these sorts of cases!

ArtOfIllusion/src/artofillusion/procedural/Module.java
```diff
@@ -332,17 +338,18 @@ public void getColor(int which, RGBColor color, double blur)
...
   {
     try
     {
-      Constructor con = getClass().getConstructor(new Class [] {Point.class});
-      return (Module) con.newInstance(new Object [] {new Point(bounds.x, bounds.y)});
+      Constructor con = getClass().getConstructor(Point.class);
+      return (Module) con.newInstance(new Point(bounds.x, bounds.y));
     }
-    catch (Exception ex)
+    catch (IllegalAccessException | IllegalArgumentException | InstantiationException | NoSuchMethodException | SecurityException | InvocationTargetException ex)
```
Sailsman63: I've already merged, and then had to revert this in master.

How did you get the multi-catch to compile? It's Java7+ and the ant file locks in at Java6

peastman: Aren't we now using Java 8 as our minimum?

Sailsman63: I thought that we had decided to wait that until releasing the latest bugfixes (Possibly my misunderstanding)

The build files still reference `source 1.6`, so this should not have compiled...

PR #172, opened 2020-03-21
Document viewcontrols Resolves #154

### LINKED ISSUE #154: Document View Controls
The 3.1.0 release contained a major re-work of how viewpoints in the various editors are managed. (Thanks, @peteihis!)

These changes are extensive enough that they really should be documented in the manual

Sailsman63:
Gah! I've now got a very visceral understanding of why so many otherwise nice Open Source projects have such limited documentation! I lost count of the time spent on this so far.

This one adds an entire mini-tutorial on using the View-Control System.

Feedback appreciated!

I'm really feeling the need to do a customized template for the manual - I'm liking *bits* of several of the sphinx included themes.


peteihis:
A little directions here: How exactly do I get somebody else's branch for review? --About a million discussion threads about it and all different...

peteihis:
> I'm really feeling the need to do a customized template for the manual - I'm liking bits of several of the sphinx included themes.

Just what I said!

peteihis:
BTW, I though that the `Resolves #somehtig` would be good in the text (not needed in the headline)?

Sailsman63:
> A little directions here: How exactly do I get somebody else's branch for review? --About a million discussion threads about it and all different...

In your AOI directory, `git remote --add <name><url>` to make sure that you have a reference to the fork - once you've done it once, never need to again for that person/repo.
So, for me: `git remote --add sailsman(or whatever you want to call me) https://github.com/Sailsman63/ArtOfIllusion.git`

Then, when you want to grab a specific branch: `git fetch sailsman <branchname>` followed by `git checkout <branchname>

(Not surprised that there are a lot of answers out there. This command has a *lot* of options!)

Sailsman63:
> nice-to-have features of the included themes.

We should probably put together a tracking issue, or discuss on sourceforge if we want community feedback
>BTW, I though that the Resolves #somehtig would be good in the text (not needed in the headline)?

It would. I just hadn't added to any of the previous commits, so typed in the headline to make the connection obvious.

peteihis:
Looks good at firs sight. Everything seems to be covered.

A couple of points that caught my eye

~~~~src
Center Click in Object List | Fit object to active view. Set working depth to center of object.
~~~~
This works for the current selection. You can select multiple objects on the list before center click. It uses `ViewerCanvas.fitToObjects()` as the 'Fit To Selection' menu command does.

The images at 13.2.2.2. and the following parts: To me it would be easier to understand if each image sequence (two or three per mode) would be shown on one row. It wold be easier to follow the movement from frame to frame. Now when there is an image, then text and then another image you a kind o lose what the difference between the images is and in the case of three the firs one is may be already outside the screen when you see the last one.

I wonder if those would be best presented as inline animations (gif?). A "bit" more work but I'd expect it to be helpful for understanding.

peteihis:
You have added this as the last thing in the manual for now but I suppose the UI should be explained somewhere in the beginning.

Now there is the is the "Getting started" part which actually introduces the UI. (Though I would definitely not put a model of a fox in the woods as the first thing there for someone who just opened the manual for the first time.) This would seem to need a different discussion. The view controls do need their own section, it can not be too deep in the header structure.

peteihis:
"Viewpoint Management"

How about just "View Management" _(or -Manipulation or Controls..)_ Somehow the "Viewpoint" feels .... I don't know ... too detailed or abstract to me. As a user I would not be thinking of handling the viewpoint just the view.

peteihis:
And more minor additions:

Tray is the default for the Scene view but Space is the default for object editors. This is a choice I made based on my own experience. (Think about modelling a paws of an animal on the underside ...)

How about the Scale/Zoom field? You can type numbers into that. (Though it seems to forget to update something.. Parallel/Perspective transition does not always happen correctly after using the numeric input. Need to have a look.)

Sailsman63:
>it would be easier to understand if each image sequence (two or three per mode) would be shown on one row. 

Probably.  The current themes don't have a good way to make several figures (images with associated text) play well together on one line. This is more a themeing thing than a document structure thing. To put them on one line, I'd currently have to duplicate what I did with the scroll cues... text in the image, and that's less than ideal. 

>I suppose the UI should be explained somewhere in the beginning

Possibly.  I'm a little torn here.  Right now, I'm thinking of it as a bit of an appendix. The introduction should give enough information to get started with,  but I don't want people to feel as if they have to slog through the entire chapter before they can move on and 'get to the fun stuff' (it's huge!  15k of text, not counting the images!)

It's possible that some bits of this page belong in the introduction, and vice versa.

It would definitely need some discussion and thought,  but what about a manual structured as a fairly lightweight tutorial,  with links to various appendices for advanced/detailed information? 

I do agree that the introduction needs some... rework,  and not just in terms of the images.  The introduction to a lightweight, intuitive, easy-to-use modeling suite should not be 4 paragraphs of marketing drone. (at least, that's how it reads to me - I'm coming to appreciate that getting it right isn't easy!)

>How about just "View Management" 

I'd love something just a little smoother... To me, though, 'view management' has more to do with the stuff in the view menu... Grids, display method, etc.

Maybe a case for some of that too be moved into this page?

We really need to get a rough idea of what the structure should be...

peteihis:
> We really need to get a rough idea of what the structure should be...

I agree.  -->  #173

> more a themeing thing than a document structure thing

Yes it is ... I know that some per file styling can also be done, but my knowledge is limited. And I have not found Spinx documentation to be of the most helpful kind ever ....

>  To me, though, 'view management' has more to do with the stuff in the view menu... Grids, display method, etc.

That did cross my mind. 

> Maybe a case for some of that too be moved into this page?

Sure everything is connected ... It is a multidimensional thing to understand it all. A good way of presenting it certainly not easy to come up with.

peastman:
> The current themes don't have a good way to make several figures (images with associated text) play well together on one line.

Could you do it by putting them in a table?

> what about a manual structured as a fairly lightweight tutorial, with links to various appendices for advanced/detailed information?

What about starting out with the lightweight tutorial, but then continuing on to the detailed manual starting in the second chapter?

Sailsman63:
>Could you do it by putting them in a table

Maybe. I played with it a bit today, and I'm finding it frustratingly fussy. It doesn't want to scale them the same...

Sailsman63:
Okay, got a table behaving the way I wanted. (Actually, two)

And I'd like some feedback. Please look at the Tray and space sections as I've got them laid out. Do you prefer Row- or Column-based?

docs/manual/view_management.rst
```diff
@@ -280,22 +285,33 @@ Zoom/Z-axis Likewise works the same as Tray_ mode.
...
+   current view direction.
 
-   Left & Right pan are relative to the current view direction.
+.. |space3| image:: view_controls/Space3.png
 
-.. figure:: view_controls/Space3.png
-   :width: 45%
+.. |spacetext3| replace:: Space mode can barrel roll! (see the `mouse
+   control table. <mousecontrols_>`_) Up & Down rotation are relative
+   to the current view direction.
```
Sailsman63: Why all these substitution definitions? It's a serious pain to format text to fit into the cells in the row-based format, and would make diffs for any future changes to phrasing hard to read.

This also makes it simpler to re-order the table any time we choose.

It's not nearly so bad in the column-based version above.

Is the non-locality worth it?

Sailsman63: Gahh! The inline diff look horrid. Please click on the revision-number to get the side-by-side version. It makes much more sense!

peteihis: > Why all these substitution definitions?

The substitutions style is the only thing that makes the least bit sense here. The table definition in .rst is just stupid to begin with.

peteihis:
What I had in mind was the style on Space mode, but actually the Tray mode style is very easy to follow. Either one would tell the story.

Two things about the pictures: The images look rather murky to me. That tends to happen with screen captures and it is not easy to combine screen capture and a rendered image... If you don't mind I could try something one of these days.

How about the colors of the rings? I'd use the same ones as the 3D-mainuplator: blue for forward/back, green for left/right, red for barrel roll?

peteihis:
One more finding:

"In this projection, the zoom field represents scale in a percentage of the default AOI scale. (100px/scene unit)" --> Something like "... view scale as pixels/scene unit. The deafult is100"

I'd drop the %s. The thinking does work as percents of 100 but that sounds like a tricky way of thinking, when what really happens is directly, 1 unit is stretched to that many pixels. Might even be simpler expressed that way "...scale: 1 scene unit takes this many pixels on screen."

docs/manual/view_management.rst
Sailsman63: Well, any other table definition setup that I've ever seen required lots of markup tags... Think of a `.html` `<tr><td></td></tr>` nest.... Not very readable, whereas the `.rst` format is at least human-understandable.

Sailsman63:
>The images look rather murky to me

Which part of them? The piping and cube were Ray-traced, the camera screenshot. I was to trying to avoid the very stark white background that many diagrams have...I find them a bit painful.

If you've got thoughts, though, feel free. You can actually make a PR against my branch, if we haven't resolved the rest of this by then.

Sailsman63:
>I'd drop the %s. The thinking does work as percents of 100 but that sounds like a tricky way of thinking, when what really happens is directly, 1 unit is stretched to that many pixels

Hmm... Maybe better that way... As long as it then stays that way!

Sailsman63:
Looked a little more at table formats. I think that the vertical arrangement works better with just two images - if I put them side by side, they will be scaled to fill the space, and the table will then be rather large vertically without being any more readable.

The three-image case (Space) does work better side-by-side.

I've also found two different ways of writing the tables internally - please take a look.
Please be aware that I intend to rebase/squash all of the table experiments into the original PR once everything settles out.

If you suspect that I'm getting too into table formatting: Maybe. But I also think that we're going to need more of this across the manual, so let's figure out the most maintainable way to do this!

peteihis:
> If you suspect that I'm getting too into table formatting

Better to do it right the first time. Patching up things that were never properly thought though later is something that I hate with all my guts. (I get to do my share of that at work ... see my point? :) )

> As long as it then stays that way!

The `ViewerScaleControl` has always been using `view.getScale()` directly, not %. I don't see why that would have to change. :)

Sailsman63:
Okay, finally got the table formatting figured out, and the commit history cleaned up. I think the tables are a lot neater than the original format.

If I don't hear from anyone, I'll merge within a day or so.

PR #174, opened 2020-03-21
Change to full TOC in manual sidebar

Sailsman63:
This I kind of feel strongly about, no matter what other decisions we make regarding templates, formatting, etc:

The sidebar should have a global toc! I want to see what my options are in whatever I'm reading!

### DIFF
--- a/docs/manual/conf.py
+++ b/docs/manual/conf.py
@@ -95,7 +95,7 @@ html_static_path = ['_static']
 # default: ``['localtoc.html', 'relations.html', 'sourcelink.html',
 # 'searchbox.html']``.
 #
-# html_sidebars = {}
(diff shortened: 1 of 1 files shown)


peteihis:
Yes!

Also the sidebar should stay put on the screen, when you scroll the content. Peter said earlier that with long files you always have the sidebar but that is not the case with the current theme: You have to scroll back up to see the sidebar and then you no longer see where you just were

Some of the themes actually paint with a different color/background on the contents list the part(s) that you are currently viewing. That is a really nice feature and helps to keep to understand where you are. I have no idea, how that is done though...

Sailsman63:
Pretty much all in the CSS.

I also want to get the sidebar in a fixed position, but need to get the styling settled out a bit first.

PR #176, opened 2020-03-27
Refactor build definition to use global properties

Sailsman63:
Should have done this the last time I messed about with the build files... can't remember why not.

Makes updating defaults much easier, and can still over-ride on the command line if needed.

### DIFF
--- a/ArtOfIllusion.xml
+++ b/ArtOfIllusion.xml
@@ -3,16 +3,19 @@
 <project name="ArtOfIllusion" default="dist" basedir=".">
 
   <!-- set global properties for this build -->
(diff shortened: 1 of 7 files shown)


makiam:
It is possible to move common properties initialization to shared xml file (ex. common,xml) and than just import this file
`	<import file="common.xml"/>`

Sailsman63:
Sure, but to what purpose? I don't see a need to jump into full `.xml` format for a simple set of key-value pairs.

makiam:
Some common taslks also may be moved to common ant-code referencing current project by ${ant.project.name}". This reduces repeated code in build scripts.

PR #177, opened 2020-03-29
Removing a confusing instruction

peteihis:
I noticed these strange instructions in the Rendering section.

I don't know what the relevance of that would have been in using the rendering dialog in the first place and even stranger, what the setting zoom to 100 had to do with anything: Would you not have liked to preview the camera as it was set? With the current code that would not even work.

### DIFF
--- a/docs/manual/rendering.rst
+++ b/docs/manual/rendering.rst
@@ -167,8 +167,7 @@ the dialogue that appears:
 
 The **Width** and **Height** define the size (in pixels) of the 2D image created.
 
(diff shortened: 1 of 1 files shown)


Sailsman63:
Probably, there was a time when the previews were 'parallel' mode only... I think we need kind of determined that?

Sailsman63:
Merging this, as it's definitely wrong/out of date.

I do think the rendering chapter, though, as a whole is the target for the next cleanup pass through the manual. Is that something you'd be interested in doing? (Just so we don't both start work on the same thing)

PR #179, opened 2020-03-29
objectinfo lock to be saved

### LINKED ISSUE #178: "Locked" status of ObjectInfo is not saved
I have had this in my mind for a while now, but I did not hurry to bring it up, thinking that the next release would be mainly a bug fix without format changes. Now that it seems like it's going to be 3.2 with the new GL, I think this should be fixed at the same.

So, I had a look a the code and `locked` is never mentioned in the write/read methods of `Scene`. I'd expect to look like how `visible` is handled.

I don't know how this should be fixed as there is no built in version numbering for `ObjectInfo`, but from user's point of view I'd expect to reopen my scenes in the state I last saved them in. I don't (and I'm not alone with this) see the locked parameter as a per session thing.

This is not the only thing that is wrong with object locking but this would the one that makes the greatest difference for the feature to be in any way useful.

peteihis:
Save `ObjectInfo.locked` status to disk with the scene.

Contains a format change. Target release 3.2

Resolves #178

### DIFF
--- a/ArtOfIllusion/src/artofillusion/Scene.java
+++ b/ArtOfIllusion/src/artofillusion/Scene.java
@@ -1,6 +1,6 @@
 /* Copyright (C) 1999-2013 by Peter Eastman
    Changes copyright (C) 2016-2020 by Maksim Khramov
-   Changes copyright (C) 2017 by Petri Ihalainen
+   Changes copyright (C) 2017-2020 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -1227,7 +1227,7 @@ public class Scene
     Class cls;
     Constructor con;
 
-    if (version < 0 || version > 4)
+    if (version < 0 || version > 5)
       throw new InvalidObjectException("");
     loadingErrors = new StringBuffer();
     ambientColor = new RGBColor(in);
@@ -1457,7 +1457,9 @@ public class Scene
     info.setId(in.readInt());
     if (info.getId() >= nextID)
       nextID = info.getId() +1;
+
     info.setVisible(in.readBoolean());
+    info.setLocked(version < 5 ? false : in.readBoolean());
     Integer key = in.readInt();
     obj = table.get(key);
     if (obj == null)
@@ -1567,7 +1569,7 @@ public class Scene
     int i, j, index = 0;
     Hashtable<Object3D, Integer> table = new Hashtable<Object3D, Integer>(objects.size());
 
-    out.writeShort(4);
+    out.writeShort(5);
     ambientColor.writeToFile(out);
     fogColor.writeToFile(out);
     out.writeBoolean(fog);
@@ -1698,6 +1700,7 @@ public class Scene
     out.writeUTF(info.getName());
     out.writeInt(info.getId());
     out.writeBoolean(info.isVisible());
+    out.writeBoolean(info.isLocked());
     key = table.get(info.getObject());
     if (key == null)
       {


Sailsman63:
I think this is good, but I'm going to wait a bit for two reasons:

1. I don't know if 3.2 is going to be next, or if there will be a 3.1.1 in between. Waiting for feedback on using the uncompleted JOGL 2.4.0 RC builds
2. We should do a quick survey, and see if there are any other File-Format breakages we want to get out of the way. I'd rather do as many as possible at once.

peteihis:
Yes, pretty much what I thought.. :)

peteihis:
I thought I'd go deeper into cleaning up what was left from the view manipulation upgrade _after_ the bug fix round was out. I can't say if there would necessarily be any format/API breakers in there but I thought, it would be time to introduce `BoundinSphere` into the code and use it on a few things. _(I have not seen that particular BoundingSphere solver idea used by anybody else, but based on what I have tested, I'd say, there probably are worse ones out there.)_

Sailsman63:
That sounds like a very different kind of thing... what do you see using a BoundingSphere for? They have their place, but what would you be hoping to simplify?

peteihis:
> That sounds like a very different kind of thing...

That's right.

> what do you see using a BoundingSphere for?

For one thing better view fits. Also some issues with the 3D-manipulator would just go away if it didn't rely on strangely oriented bounding boxes. I don't think it as something that objects would need to keep track of, just as a tool that is used when needed. Those for now.

makiam:
>     2. We should do a quick survey, and see if there are any other File-Format breakages we want to get out of the way. I'd rather do as many as possible at once.

IMO another option to discuss to add to ObjectInfo is an renderable flag to separate object visibility in editor and renderable

peteihis:
>  to separate object visibility in editor and renderable

As a matter of fact, I have sometimes missed the possibility to get wire frame objects and light source rendered with the normal scene.... but that'd be a very different case.

Thinking of the uses of separating interactive visibility and renderability, I'd vote "no". 

I can think that sometimes you'd need to use reference objects on the editor, that you don't want to show up on the final image/movie and it would be nice to be able to mark them not renderable, but really I believe that, to most users, it would just become a source of problems.

In professional (engineering) tools there may be several different states of presence defined for an object ... visible, active, enabled, suppressed, reference... They all have their uses but at the same time they add complexity to using the software. About 99,9% of time visible/not is what I use and all the rest just causes trouble.... Another one I use a lot is the "grounding pin" or "lock" as it is called in AoI.

Sailsman63:
>issues with the 3D-manipulator would just go away if it didn't rely on strangely oriented bounding boxes.

Please look at opening a different issue it survive discussion about that. This does not seem like something that would change the file format, even if we do go that direction.

Sailsman63:
>separate object visibility in editor and renderable

Does that simplify the interface or workflow in any way? If not, I agree with Pete: just adds complexity. AOI is focused on being simple and intuitive. Every extra UI element tends to work against that, so it takes a strong need to justify it's existence.

Sailsman63:
>reference objects on the editor, that you don't want to show up on the final image/movie

IIRC, this was one of the intentions behind the selection plugin, when it was first written.

>get wire frame objects and light source rendered with the normal scene

What I've been doing is render the scene, then hide the renderable objects and take a screenshot. Overlay in GIMP/Photoshop.

Do you see use-cases beyond diagrams for the manual or tutorials?

peteihis:
Oh, I forgot, but there is at least one thing that would change something.... I'd call it `ViewCamera.´ interface (suggestions welcome about the name). I see this as a necessity. 

Currently we have some "if-stacks" in the code to sort out what kind of an object is assigned as a camera to a view. For SceneCamera the zoom and fit parameters need to be calculated in ViewerCanvas and it is not possible to add new view camera objects without hard coding the changes to several places in the code. One such new object would be the user defined preview point, that was requested already in the beginning of the view manipulation project. Plugin cameras would need to extend one of the existing recognized camera objects, that might not have anything to do with the object that it really is.

The idea is that any object that can be controlled by assigning to a view should be able to produce `distToScreen` (or projection the distance used inside theCamera), FoV-angle(s), the information which one is the driving parameter... There might be a few others.  

Currently SceneCamera does not calculate dist to screen and the others can't calculate FoV. It would help to just `getVerticalFoV(this view)` and not think if that object knows it or not.

Sailsman63:
>`ViewCamera.´ interface (

Do you see this changing the file format version? If not, it's definitely a different issue... If so, how so?

peteihis:
> Do you see use-cases beyond diagrams for the manual or tutorials?
 
Generally speaking presentations ... But I was thinking of it more as an advanced screen capture feature or something like that rather than a rendering feature (but where's the line....?) To me it actually sounds more like a plugin than anything built into the renderers.

> Do you see this changing the file format version?

I was afraid it might, but thinking of it, I see no reason why it would. --> Different issue. Just adding API.

peteihis:
> Please look at opening a different issue

Just raking my memory now ... I may have a thing or two on the waiting list. :)

peteihis:
One potential one in #182. Needs study and is a much bigger job.

Sailsman63:
Okay, I've come to the conclusion that we probably won't be adding other file-format changes. (If we do come up with any, it's simple to add them.)

PR #181, opened 2020-04-10
Joint handling updates

### LINKED ISSUE #164: The DOF Graph drawing
Hi.

As we happened to be in the neighborhood, this old pebble in my shoe caught my attention again. There are a number of things that seem to be a bit off with the DOF Graph.

DOFGraph

The image is a scaled up blend of two screen captures. Believe it or not the value it displays is 90°, not e.g. 88,5°. The larger gray disc is the one that is displayed, when the DOF is locked. It is bigger than the active dial but it is slightly misplaced. Further more, in the beginning of the code I find:
~~~~Java
    private static final int SIZE = 64;
    private static final int RADIUS1 = 64;
~~~~
...ehhh... diameter equals radius?

I don't know what has been available at the time the DOF Graph was coded but for now I'd suggest:

- Make it a bit bigger.
- Use an odd value for the sizes, so the center point is in the middle of the middle pixel. The graphs look a lot sharper this way.
- Draw in `Graphics2D` (Use a transparent  `BufferedImage` as canvas)
- Use anti aliased drawing.
- Use the options that draw in double values.
- Use `AffineTransform` to move the 0,0 to the center pixel. This simplifies the rest a lot.
- Consider revising colors, like red(dish) tone for restricting the total range, yellow(ish) for comfort range, green(ish) for the bend angle...(if any color at all, might be too much graphic information.)
- Try to have better separation between the handles of the zones.

Of course mouse event handling would be affected accordingly.

peteihis:
The DOF-editor as discussed. Closes #164. 

- Now the handles can be separated even if the opposing handles are on top of each others just by picking on the side that you wish to drag.
- The general "jumpiness" is tamed down a bit.
- The colors have dark/light contrast that should be sufficient even to reduced color vision. I avoided using direct green/red contrast.

EDIT: I'll try to see if I can reduce the size of the editor a bit as in it's current design it covers so much of the screen that you can not see the bones, that you are working on....

### DIFF
--- a/ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
+++ b/ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
@@ -266,7 +266,7 @@ public class JointEditorDialog extends BDialog
       comfortBlock.add(comfortBox, boxgap);
       comfortBlock.add(comfortRow, rowgap);
 
-      stiffBlock.add(new BLabel("Stiffness"), rowfirstontop);
+      stiffBlock.add(Translate.label("Stiffness"), rowfirstontop);
       stiffBlock.add(stiffnessSlider, rowgap);
 
       // Make sure, that all the name labels are the same size. The text is not translated.
--- a/ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
+++ b/ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
@@ -54,9 +54,10 @@ public class JointEditorDialog extends BDialog
     RowContainer nameRow = new RowContainer();
     nameRow.add(Translate.label("Name"));
     nameRow.add(nameField = new BTextField(joint.name, 20));
-    content.add(nameRow, BorderContainer.NORTH, new LayoutInfo());
+    content.add(nameRow, BorderContainer.NORTH, new LayoutInfo(LayoutInfo.WEST, LayoutInfo.NONE, new Insets(5, 5, 5, 5), null));
+
     FormContainer center = new FormContainer(4, 2);
-    center.setDefaultLayout(new LayoutInfo(LayoutInfo.CENTER, LayoutInfo.NONE, new Insets(5, 5, 5, 5), null));
+    center.setDefaultLayout(new LayoutInfo(LayoutInfo.WEST, LayoutInfo.NONE, new Insets(5, 5, 5, 5), null));
     center.add(makeBorder(ang1Panel = new DOFPanel("X Bend", -180.0, 180.0, joint.angle1)), 0, 0);
     center.add(ang1Panel.getGraph(), 1, 0);
     center.add(makeBorder(ang2Panel = new DOFPanel("Y Bend", -180.0, 180.0, joint.angle2)), 2, 0);
@@ -65,6 +66,7 @@ public class JointEditorDialog extends BDialog
     center.add(twistPanel.getGraph(), 1, 1);
     center.add(makeBorder(lengthPanel = new DOFPanel("Length", 0.0, Double.MAX_VALUE, joint.length)), 2, 1);
     content.add(center, BorderContainer.CENTER);
+
     if (joint.parent == null)
       lengthPanel.setEnabled(false);
     RowContainer buttons = new RowContainer();
@@ -72,6 +74,7 @@ public class JointEditorDialog extends BDialog
     buttons.add(cancelButton = Translate.button("cancel", this, "doCancel"));
     content.add(buttons, BorderContainer.SOUTH, new LayoutInfo());
     setResizable(false);
+    addEventLink(WindowClosingEvent.class, this, "doCancel");
     pack();
     setVisible(true);
   }
@@ -118,7 +121,8 @@ public class JointEditorDialog extends BDialog
 
   /** Inner class representing a panel for editing a single degree of freedom. */
 
-  private class DOFPanel extends FormContainer
+  //private class DOFPanel extends FormContainer
+  private class DOFPanel extends ColumnContainer
   {
     public ValueField valField, minField, maxField, minComfortField, maxComfortField;
     public ValueSlider stiffnessSlider;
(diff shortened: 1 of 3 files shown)


peteihis:
Size reduction done. Grouping of the elemets should also be clearer now.

Sailsman63:
Sorry for the delay on reviewing this. I like the results, will be doing a pass through the code shortly.

ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
```diff
@@ -114,7 +121,8 @@ private void valueChanged()
 
   /** Inner class representing a panel for editing a single degree of freedom. */
 
-  private class DOFPanel extends FormContainer
+  //private class DOFPanel extends FormContainer
```
Sailsman63, line 124: Cleanup point: please, no commented-out code

ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
```diff
@@ -125,18 +133,19 @@ private void valueChanged()
 
     public DOFPanel(String name, double min, double max, DOF dof)
     {
-      super(3, 7);
+      //super(3, 7);
```
Sailsman63, line 136: And here.

ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
```diff
@@ -348,72 +401,166 @@ public DOFGraph getGraph()
   private class DOFGraph extends CustomWidget
   {
     private DOFPanel panel;
-    private int dragging;
+    private boolean dragging0, dragging1, dragging2, dragging3, dragging4;
```
Sailsman63, line 404: Any particular reason that you broke these up into multiple Booleans? If I understand correctly, only one of them can be true at a time. If that supposition is true, a more idiomatic structure here would be a `private enum` of the various dragging 'modes'. (With descriptive names)

ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
```diff
@@ -348,72 +401,166 @@ public DOFGraph getGraph()
...
+      valueVector1 = getValueVector(minLimit, ARM_R2);
+      valueVector2 = getValueVector(maxLimit, ARM_R2);
+      dragging1 = onHandle(valueVector1, mouseVector, HANDLE_R);
+      dragging2 = onHandle(valueVector2, mouseVector, HANDLE_R);
+
+      if (dragging1 && dragging2)
+      {
+        if (maxLimit-minLimit > 180)
+          if (Math.abs(maxLimit-mouseAngle) < Math.abs(minLimit-mouseAngle))
+            dragging1 = false;
```
Sailsman63, line 503: And places like Here...

ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
```diff
@@ -348,72 +401,166 @@ public DOFGraph getGraph()
...
-      if (dragging > -1)
-        disableUpdating = true;
+      valueVector1 = getValueVector(minLimit, ARM_R3);
+      valueVector2 = getValueVector(maxLimit, ARM_R3);
+      dragging3 = onHandle(valueVector1, mouseVector, HANDLE_R);
+      dragging4 = onHandle(valueVector2, mouseVector, HANDLE_R);
+
+      if (dragging3 && dragging4)
+      {
+        if (maxLimit-minLimit > 180)
```
Sailsman63, line 537: Here...

ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
```diff
@@ -432,47 +579,57 @@ public void run()
...
+      }
+
       double val = panel.dof.pos;
       boolean range = panel.rangeBox.getState();
       double minv = (range ? panel.dof.min : panel.min);
       double maxv = (range ? panel.dof.max : panel.max);
       double minComfort = panel.dof.minComfort;
       double maxComfort = panel.dof.maxComfort;
-      if (dragging != 1 && angle < minv)
+      if (!dragging1 && angle < minv)
```
Sailsman63, line 603: And down through here, we would `case ... switch` on the enum.

ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
Sailsman63: I guess the core question is: How many of these booleans can be true at the same time?

peteihis: Actually, when I built it the way I did I had a slightly different plan on how to decide which handle to move. Also the graphic appearance is pretty much parametrized. It was possible to have the handles overlap in their initial positions more than they do now.

I can change it back closer to what it was.

ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
peteihis: Actully there is a short sections a few lines down from this, where switch could be used. Not really anywhere else.

ArtOfIllusion/src/artofillusion/animation/JointEditorDialog.java
Sailsman63: These are each one of the handles, correct? So Joint, CW_hardstop, CCW_hardstop, CW_softstop & CCW_softstop?

If you're moving one of the hard-stops, the update code should, of course, validate if the  soft-stops and the joint are still at valid locations. You are still only *dragging* one handle.

This is, in fact, almost a perfect textbook case of what the `enum` structure was invented for. I think @peastman used an `int` originally because `enum` was not part of the java language yet. (That's an old, C-lang way of faking an enum)

Would you like me to run up a PR against your branch to show you what I mean?

At the very least, we should give each of them a descriptive name...

peteihis: > You are still only dragging one handle.
> give each of them a descriptive name...

Yes, I was just deriving from the existing code but logically it should separate `onHandle` and `dragging` tell better exactly what.

> Would you like me to run up a PR against your branch to show you what I mean?

Go ahead. :)

Sailsman63: I'm actually going to leave this sub-project for now. I still think that something along those lines is a better target, but I don't want to hold up these user interface changes anymore.

I'll come back to it once we've got a reasonable release finished.

PR #183, opened 2020-04-11
initial twist angle

peteihis:
Addressing a long term bug in creation of new bones.

The "twist" angle is in principle calculated correctly from the various directions and coordinates that are involved but often the resulting angle is 180° or -180° instead of 0° (or for example -123° instead of 57°).

The near 180 degree angles prevent using angular restrictions and often force the movements to happen over the ±180° limit, which is a bit difficult to handle. A typical place to set limits to twist angle would be a spine structure, where each bone may be allowed to move only a few degrees. To add restrictions you'd have to turn the bone around but if the bone already has children, that usually is not a reasonable option. 

- This makes sure that the initial Z-rotation of a new bone is always between -90° and 90°.
- This does not affect the the creation or functionality of any children of the joint.

Target release: The next to come.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/animation/Joint.java
+++ b/ArtOfIllusion/src/artofillusion/animation/Joint.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 2002,2003 by Peter Eastman
+   Change Copyright (C) Petri Ihalainen 2020
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -41,6 +42,8 @@ public class Joint
       length = new DOF(0.0, Double.MAX_VALUE, d);
       calcAnglesFromCoords(false);
     }
+    if (twist.pos >  90) twist.pos -= 180;
+    if (twist.pos < -90) twist.pos += 180;
     length.fixed = true;
     angle1.loop = angle2.loop = twist.loop = true;
     id = -1;


Sailsman63:
I think I see where this is going, and on the surface it makes sense. I've just never run across the issue that it's supposed to solve (I don't do much rigging)

Do you have an example?

peteihis:
Like this. -- Just make something and start drawing bones to a variety of directions. Rotate the view and draw more. Nothing special is needed to make it happen.

[Bone-twist_example.zip](https://github.com/ArtOfIllusion/ArtOfIllusion/files/4556157/Bone-twist_example.zip)

On this one the bones that are drawn to the right side (in front view) have this happening. I haven't tried to figure out a rule to which directions would result which twist, but as the view orientation, drawing direction and parent coordinates are all factors (3 coordinate systems) that'd be very hard. It is not unusual that in the exact same conditions you may get 0, +180 or -180.

Sailsman63:
Thanks for the example file!

I've played around with this, (Both the example, and a new scene, both with and without this fix.) and I'm still not getting it. The results that I see from the existing code make sense to me, and I've not seen any differences when building a new example with the changed code.

Could you maybe do one with a single joint that you feel is wrong, and then an equivalent with it arranged the way you feel it should be?

peteihis:
I was sure I had answered thos but seem that I never sent the answer after all.

> Could you maybe do one with a single joint that you feel is wrong, and then an equivalent with it arranged the way you feel it should be?

Just see the example file. The twist angle for all of the child joints is supposed to be 0°. The ones on the right side have 180° or -180°, which makes using restriction/comfort zones impossible and also makes normal editing difficult as even a tiny ±5° forth and back shifting difficult as it will have to go over the ±180° limit.

Sailsman63:
Oh, I think I see it now. Less about the joint itself, and more about the interface causing issues.

I'm seeing two bugs in how the interface handles this:

- Joint Orientation "Snags" at the 180/-180 mark, rather than smoothly overflowing into the other half of the circle.
- Moving the Limit handles from this position causes out-of-range values, (like -210, instead of wrapping back to 150) and seems not lo like the idea of a limit range that crosses the 180/-180 delimiter.

This change will mitigate both of those, but not completely fix them. I'll probably merge this in a bit, but we need to keep them in mind.

Sailsman63:
And it's also more intuitive for a joint's default rotation to be '0', so that's another argument in favor of this.

peteihis:
> Joint Orientation "Snags" at the 180/-180 mark, rather than smoothly overflowing into the other half of the circle.

That's a part of the old editor code, fixed in #181. Without restrictions activated the movement over the 180° mark is smooth and the threshold for the jump to the other side is lower than before. -- It would be logical for it to switch to the opposite side handle (at least in some cases) rather than to turn the current handle almost a full circle back, but that would add quite a few more rules to the already complex logic, so I did not want hold the PR back for unsure results ... 

> Moving the Limit handles from this position causes out-of-range values, (like -210, instead of wrapping back to 150) and seems not lo like the idea of a limit range that crosses the 180/-180 delimiter.

Another editor issue, fixed in #181.

PR #184, opened 2020-04-11
Finnish translation

peteihis:
A small fix.


Sailsman63:
The latest change should be made over in #181.

peteihis:
Well, I had them all in the same file and they eventually will be anyway...

peteihis:
... And they are all more or less related to #181. Consider this an after patch?

peteihis:
I have changed this according to your request in the last push -f

PR #185, opened 2020-04-12
Reconnect delete-key

peteihis:
In the early versions of AoI delete was under Delete key as you would expect. At some point it was moved to the Backspace-key, which still after all the years is confusing to me and I believe to most users anyway....

This PR adds the function back to Delete-key as well. There's nothing wrong with using BS for the same purpose but a dedicated key, that ROW uses, should not be blank.

..So now it would be much like any text processor. Either Del or BS will delete the current selection.

### DIFF
--- a/ArtOfIllusion/src/keystrokes.xml
+++ b/ArtOfIllusion/src/keystrokes.xml
@@ -14,6 +14,22 @@
 			else
 				window.deleteCommand();
 		}
+	<keystroke name="Delete Selection" code="127" modifiers="0" >if (window instanceof LayoutWindow)
+		{
+			score = window.getScore();
+			if (score.getSelectedKeyframes().length == 0)
+				window.clearCommand();
+			else
+				score.deleteSelectedKeyframes();
+		}
+		else if (window instanceof MeshEditorWindow)
+		{
+			if (window.getToolPalette().getSelectedTool() instanceof artofillusion.animation.SkeletonTool)
+				window.deleteJointCommand();
+			else
+				window.deleteCommand();
+		}
+	</keystroke>
 	</keystroke>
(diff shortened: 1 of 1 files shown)


Sailsman63:
I think that I like this. I'm not sure why @peastman chose the backspace key as default - he set that when he implemented the scriptable keymappings.

Sailsman63:
Just tested this, and found a snag:

AOI re-generates the default keystrokes only when there is no `keystrokes.xml` saved in `~/.artofillusion/`

So, this only takes effect for *new* installs, or if the user deletes the keystroke file, which resets any custom script shortcuts that they may have set up. Not sure how to handle the deploy on this.

Sailsman63:
Okay, I'm merging this. The upgrade path will need to be handled separately. (Probably a startup script that removes itself after running.)

peteihis:
Yes, I thought about that start-up script possibility too....

Then there is another thing to check: As part of the view manipulation upgrade some short cuts were changed. Those too are in use only in entirely new installs. They would be Ctrl+F for "Fit to Selection" in **all** windows, Ctrl+H for free hand selection (ex Ctrl+F in mesh editors) and Ctrl+Shift+F to Fit to All.

The script could probably do the change and the notify the user that this was done, if something was done. -- Basically I would not leave it for the user to decide on the go but maybe the script should make a safe copy of the short cuts first...

Sailsman63:
> why a very basic function... would be user-editable?

Any of the below:

- People can't agree on what key to use, even if you or I think it's obvious. (See Peter preferring Backspace for his own workflow, even though this has been a perennial question since 2.3)
- Even when there is a good default, this key is not always comfortably accessible on every keyboard
- Why have two mechanisms for key-mapping? If you're going to have a system for custom key-mappings, why would you continue to carry around the hard-coded mappings?

(Incidentally, if you do run across hard-coded key-mappings anywhere, I'd like to know about them.)

PR #186, opened 2020-04-14
ObjectSet save fixed

peteihis:
As per subject.


Sailsman63:
Thanks for the quick fix!

Question: did you try it *without* the setter? I Wasn't *Certain* that we would need it, just seemed like it would be possible, depending on the underlying code...

peteihis:
> Question: did you try it without the setter?

I did. The ObjetSets came back empty, only with their names.

peteihis:
I remember now that the spec does say, that it needs a no-args constructor and dedicated set and get method pairs for the data.

Sailsman63:
Thanks for checking. That actually makes sense - it generates a `new ObjectSet()` and then uses the setters to actually restore the data.

PR #188, opened 2020-05-03
Precise module render

makiam:
Using Floats produces module outlines render better aligned


Sailsman63:
Why would we want to align module barriers by a fraction? We use whole numbers as they align to the nearest pixel.

PR #189, opened 2020-05-03
Update module duplicate result type

makiam:
Update procedural module's duplicate() method declaration to be more specific. This reduces dependencies count of deprecated Module class

### DIFF
--- a/ArtOfIllusion/src/artofillusion/procedural/BrickModule.java
+++ b/ArtOfIllusion/src/artofillusion/procedural/BrickModule.java
@@ -259,7 +259,7 @@ public class BrickModule extends ProceduralModule
   /* Create a duplicate of this module. */
 
   @Override
-  public Module duplicate()
+  public BrickModule duplicate()
   {
     BrickModule mod = new BrickModule(new Point(bounds.x, bounds.y));
 
--- a/ArtOfIllusion/src/artofillusion/procedural/CellsModule.java
+++ b/ArtOfIllusion/src/artofillusion/procedural/CellsModule.java
@@ -258,10 +258,10 @@ public class CellsModule extends ProceduralModule
     return true;
   }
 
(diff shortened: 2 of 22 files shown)


Sailsman63:
No. This is a part of the public API. It *Must* return the more general type, which is still `Module.`When we are ready to remove the deprecated code, it will then need to return `ProceduralModule`

makiam:
This fix is not violates Module contract because more specific type returns is still Is-a module.

peastman:
I think this is ok.  Covarient return types have been supported all the way back to Java 5.

Sailsman63:
You know, you may have a point. I've always stayed away from them because it felt wrong, even though it's legal.

I've never seen a compelling use case - do you find that this makes the code easier to read?

peastman:
It can avoid unnecessary type casts, so yes, it makes the code easier to read.  And it's generally safer.  The more specific the return type is, the more information the compiler has for catching errors.

Sailsman63:
> avoid unnecessary type casts

That's true if the calling code cares about what type of module it gets.

I'm having trouble imagining an instance where you would call this method *and know/care what kind of module you get.* To me, the more generic return type is part of the documentation. It says "Hey, this is part of a bigger system. If you're thinking of calling this, you really should be planning on handling *any* type of module."

peastman:
If you're writing generic code to process procedures you probably won't care.  But if you're writing, for example, a plugin that builds procedures programmatically, there's a good chance you'll care about the return type.

Just as a matter of principle, I think it's always good to make signatures as specific as possible.  If you know the return value will always be a particular class, say so.  It will never hurt and will sometimes help.  In fact, I'd vote for making this same change to all `duplicate()` methods (objects, textures, etc.).

Sailsman63:
Okay, I see where you're coming from.

Let's go for it.

PR #190, opened 2020-05-16
A bug fix to view manipulation

peteihis:
I have noticed, when playing wit zoom, center click, perspective etc... that sometimes the view is not responding the way I expected but I could not exacty tell where it went wrong and the problem always seemed to fix itself by the following view manipulation actions.

Now I found it! 

This makes sure that changing from parallel to perspective keeps screen dimensions unchanged at working depth, after you have selected a new point of focus by center click. This is the way it was originally supposed to be.


Sailsman63:
This sounds right. I've always had issues reliably duplicating the glitches I came across, but this fix "Feels" right when I tried it out. We'll go for it!

Thanks!

PR #191, opened 2020-05-18
Update minimum supported java version

Sailsman63:
I think we settled this already... are there any last minute concerns?


peastman:
Fine with me!

Sailsman63:
Okay, in it goes!

PR #192, opened 2020-05-23
Procedure owner defaults

makiam:
Move default method implementation to ProcedureOwner defaults


makiam:
> 
> 
> Tests should be of meaningful work. These methods are essentially hard-coded configuration options, testing them is a lot like testing against the contents of a configuration file.

Tests are just a proof of no regression introduced

peastman:
> I think it's only implemented the way it is is because configuration objects could get a little weird in this context.

I think it's implemented the way it is because default implementations didn't exist yet when I first wrote it. 😀  Eliminating duplicated code is generally a good thing.

Sailsman63:
Quite probably, and also that there's no obvious way to specify a set of configuration options in an interface. I do like simplifying/reducing code, as long as it does not lose clarity along the way.

My point is that these methods aren't really "Methods" at all. They're a stand-in for a set of hard-coded configuration flags. (not configuring the `ProcedureOwner`, but configuring the `ProcedureEditor`)

The proposed tests are not testing behavior, because there *is no behavior to test* - it's just comparing a configuration against another, manually entered copy of the configuration information.

Sailsman63 (approved):

PR #194, opened 2020-06-17
Update JOGL to 2.4.0-rc-20200307

Sailsman63:
Jogl hasn't actually gone stable with 2.4.0, but the remaining issues on their roadmap shouldn't affect us.

Why make this change? This should fix the `inconsistant ld.so` issue that has come up with many recent builds of openjdk.

If you're building from source, you'll need to manually copy the `*.natives.*` jars for your platform from the InstallerSrc directory to the `Live_Aplication/lib` directory.

I'm also comitting linux libraries for arm here. (mostly Pi type boards, etc)

thoughts?


Sailsman63:
Has anyone gotten a chance to test this? I also posted a build on sourceforge, if you don't want to deal with building it yourself...

Should I just bring it into mainline? @peastman  @peteihis ?

peteihis:
I haven't tried to build it, but I copied the files into my current AoI non-development AoI. No problems so far though I haven't really used it that much. I have been mostly focusing on the development cases.

Sailsman63:
>I copied the files into my current non-development Ao

Do you have openGL enabled? Does it work as 'available'? I'd be surprised if the files do work with a build that was not compiled against them...

peteihis:
Yes, yes and it does.

peteihis:
And I removed all the old content of lib.

Sailsman63:
Just also figured out that this fixes a bug that was preventing AOI from running correctly on mac with newer builds of Java 8, and any build of higher java versions. Since I've got confirmation that this works on all three major platforms, I'm going to go ahead and merge.

Issue #193, opened 2020-05-26
Raytrace rendering bug with large objects


peteihis:
There is a bug in ray tracing that is triggered by the size of the object. After some limit size rays seem to return only the color of the ambient light from the surface ignoring the portion of light that should come form lights. 

The triggering size depends on surface type (triangles, or primitive) and there are other factors that affect the result like camera orientation, lighting direction number of lights, distance form the camera to the surface.... It seems that all light produce their own patterns of missing light but sometimes when a pixel rejects light, it rejects it from all lights. Sometimes just one. Another interesting finding is that on a sphere the pattern is mirrored in Y-direction.

![Sphere](https://user-images.githubusercontent.com/22843957/82940303-4b495380-9f9d-11ea-94c6-ff4801b0bd7b.png)

Sphere, R = 80

![Cylinder](https://user-images.githubusercontent.com/22843957/82940315-4f757100-9f9d-11ea-9701-39b6a9832bc2.png)

Cylider, R = 80, H = 80 

![Pyramid_5](https://user-images.githubusercontent.com/22843957/82940329-53a18e80-9f9d-11ea-8ab9-0db3562fc097.png)

Pentagon based pyramid, Edge = 2000

A cube would have to be aroud 10000 units in dimensions before the bug appears. On the pyramid camera distance seems to have a pretty random effect on how bad it gets... Just one Alt+scroll-tick may have the defect almost disappear or ruin the entire image. The case above is somewhat intermediate.

On the outside this seems to be a part of the same family that has produced colorless pixels in the past but there is the difference that the pixels do have a color which allows further processing. The colorless pixels had the tendency of infecting the neighboring pixels as well if for instance filters were applied. Anti aliasing may mask the problem in milder cases but not when the image is noticeably damaged.

Now the process of collecting information from the surface for a ray to return is a pretty complex operation and for the time being I can't begin to imagine where it goes wrong. Some directions, where to dig, would be welcome.

Sailsman63:
Huh...

Findings:

1. This is not *random* noise. for a given object/light/camera position setup, the exact pattern is repeatable. I've even saved a file, and re-started to make sure.
2. Primitive objects do not seem to be affected by surface accuracy settings, or by switching to low-memory mode.
3. The patterns seem to have a very strong directional component. For Spheres, it seems to be related to Camera View direction. For Cylinders, it seems to be related to the rotation of the cylinder itself.

Sailsman63:
[...]

PR #196, opened 2020-07-03
Raytrace large objects

peteihis:
Partial solution to things discussed in #193.

These at least push the boundaries, where the problems start, a bit farther. 

Unfortunately, a completely flat object on the scene may still drive the rendering into a loop, when the general size of the scene is somewhere between 25000 and 50000 units. Before that there already are serious quality issues on smaller sizes.  I have not been able to figure out, where _that_ loop happens but it is instant, when you launch the rendering.

The loop catcher in the commit does work for that particular case. You can see the result in the image, when it happens: Instead of transparent pixels you get pixels with the wrong color spread around the problem area.

<b>EDIT:</b> I have tested so many things that I got confused with what I had where: <b>&rarr;</b>

I have tested relative tolerance on  _a few more things_. With that modification I managed to render a somewhat demanding case without very serious problems or _(this is noteworthy! **\***_) looping treads: The case took over 5 hours with an 8-thread 4,0 GHz engine. Millions of triangles in the scene, a reflective water texture, 8/16 (or maybe 16/32) anti aliasing and 16 ray ambient occlusion, the largest object was with 20000 unit diameter _(= sea 10 km to the horizon)_. If the process had been bound to crash or loop, the probability is rather high, that it would have. 

Those changes were mostly in `Raytracer`. Then there is still the `OctreeNode`. Among other issues, its `contains` can not seem to work with the same rules to rays and nodes.... <b>&larr; edit</b>.

Until the next update. :)

_**\***) It is really hard to think of anything worse in that moment, when you have about 90% of a movie rendered and then CPU usage drops to show activity of one thread and never gets forward. You know that the image and therefore the entire movie will never be finished because the renderer can't get over just one damn ray in one damn pixel!_

### DIFF
--- a/Renderers/src/artofillusion/raytracer/OctreeNode.java
+++ b/Renderers/src/artofillusion/raytracer/OctreeNode.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 1999-2015 by Peter Eastman
+   Modification copyright (C) 2020 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -11,7 +12,6 @@
 package artofillusion.raytracer;
 
 import artofillusion.math.*;
-
 import java.util.*;
 
 /** This class represents a node in an octree, used for sorting the objects by location in
@@ -28,9 +28,10 @@ public class OctreeNode
   public RTObject obj[];
   public float minx, maxx, miny, maxy, minz, maxz;
   public float midx, midy, midz;
+  private double nodeTol, rayTol;
   
   private static final int CELLS = 64;
-  private static final RTObject EMPTY_OBJECT_LIST[] = new RTObject [0];
+  private static final RTObject EMPTY_OBJECT_LIST[] = new RTObject[0];
   private static int leftCount[] = new int [CELLS+2];
   private static int rightCount[] = new int [CELLS+2];
 
@@ -48,8 +49,13 @@ public class OctreeNode
     this.maxz = maxz;
     boolean inside[] = new boolean [tri.length];
     int count, i;
-
     parent = parentNode;
+
+    nodeTol = maxx-minx;
+    if (maxy-miny > nodeTol) nodeTol = maxy-miny;
+    if (maxz-minz > nodeTol) nodeTol = maxz-minz;
+    nodeTol *= Raytracer.TOL;
+    rayTol = nodeTol*0.5; // Otherwise completely flat objects might be entirely missed.
 
     // Find which objects are actually inside this node.
 
@@ -149,6 +155,7 @@ public class OctreeNode
 
     if (!contains(pos))
       return null;
+
     current = this;
     while (current.obj == null)
       {
@@ -157,16 +164,16 @@ public class OctreeNode
             if (pos.y > current.midy)
               {
                 if (pos.z > current.midz)
-                  current = current.child[7];
+                    current = current.child[7];
                 else
-                  current = current.child[6];
-              }
+                    current = current.child[6];
+            }
             else
               {
                 if (pos.z > current.midz)
-                  current = current.child[5];
+                    current = current.child[5];
                 else
-                  current = current.child[4];
+                    current = current.child[4];
               }
           }
         else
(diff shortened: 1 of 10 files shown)


Sailsman63:
Thanks for putting in the effort!

I'm not sure that finishing a render with speckles is the way to go here... would it be better for the render to abort? Or have a setting for the user to chose whether to abort or continue? (not a setting that I would want to spend a lot of time on)

peteihis:
> Or have a setting for the user to chose whether to abort or continue?

That setting actually is there already: The user can press "Cancel" any time during the process if he doesn't like what he sees happening. And the user can abandon the ready picture, if the case was a fast one.

And definitely it needs to be up to the user to decide. An occasional misplaced pixel may not destroy the entire image. If the quality really looks like "the soup" hit the fan, then you might think otherwise, but I could not tell on anybody's behalf, what the acceptable limit would be in each case.

Testing with primitives I've got good some pretty strong indication that the situation can still be improved. The difficult piece seems to be the `RTTriangle`. 

Now I'm more worried about the possibility, that a thread can fall into a loop which makes the process immune to Cancel button. That probably is not a hard case to rig with a safety switch as soon as I find where that happens.

Sailsman63:
> I'm more worried about the possibility, that a thread can fall into a loop which makes the process immune to Cancel button. 

Don't worry about that unless you actually see it happen. It's really not a good idea to add escape hatches in the worker threads. If we see this in the wild, we should look at the overall coordination picture.

> Testing with primitives I've got good some pretty strong indication that the situation can still be improved. The difficult piece seems to be the RTTriangle.

Is this in regard to the node issue, or the intersection-miss issue? Remember that, though they look similar, they come from different areas.

peteihis:
> ..unless you actually see it happen

I see it frequently with some very simple test cases and I know a sure way to make it happen -- which is why I'm sure that I can find where it is. I don't like the idea of slowing down the ray process with extra checks either and it may be possible to avoid that if I find what causes it, but on the other hand there already are about a squillion if-checks in the ray handling path, so I feel that one more would be a small price to pay for a saved rendering case.

> Is this in regard to the node issue, or the intersection-miss issue?

Those are related and in the bottom they actually use the same method, where decisions are made, though the route there is different. I have a couple of ideas what to try next with it...

peteihis:
So, I'm a kind of back where I was a couple of weeks ago, but now I may have a fuller picture of what is happening:

Those rays that "miss" the object are not actually traced at all. `RaytracesRenderer.spawEyeRay`is called pixel-many times, but it does not always call `traceRay`. That is because it first call `OctreeNode.findFirstNode`, which calls `findNode`, which calls `contains`, whics return `false`. &rarr; no node, no ray even though the ray _did_ find a point, where it hit the object.

Where things seem to go wrong is, when `findFirstNode` moves he point to check 'just inside' the node. Sometimes `TOL` is so small compared to the dimensions that it disappears into numerical error. This starts to happen somewhere at 1000's of units for object size, which is not really that much, say for a landscape behind a building...

Now one troublesome special case is a completely flat triangle with one of the coordinate dimensions 0.0. The root node around it would be 2 × TOL + next Floats thick. It is much easier to miss than anything thicker.

I think I'll calculate a size dependent tolerance for each node and see if using those will help in finding the first and also the following nodes.

Sailsman63:
This sounds like a spot where creating some unit tests, fuzz/property testing style, of the specific methods that are failing here...

If you like, I could try to look at that in a bit.

peteihis:
> some unit tests

Might be worth trying. There seem to be parts that I have been missing so far and if it helps to see what's happening or even confirms (or proves wrong) what is already known it helps to solve the case.

peteihis:
Some new findings:

Basically the tolerance, that is based on the size of OctreeNode works perfectly well, when there is only one object to render. When you add more objects and increase the scale, a couple of other problems emerge:
- Infinity loops (again)
- Flat objects disappearing either entirely or partially

The infinity loop happens in the `RaytracerRenderer.traceLightRay`, where the do-loop starts to find the same node again, though it should realize that it is at the last node. It is almost exactly like the while loop in `Raytracer.traceRay`. 

The flat object thing seems more complex and the outcome depends on relative distances of the objects. I'm not sure what happens there (or where 'there' actually is) but in _some_ cases there are `SearchlistClassLoader` lines printed into `System.out`.

I also found out that the relative objectTolerance of `Cube` may be somehow wrong in the PR. Needs another look.

peteihis:
Äärrghh.... I had made a setting to `OctreeNode.findNextNode` that ended up recalculating the relative tolerance into a kind of "tolerance of tolerance". This change was not included in the PR. Reverted that closer to what I originally though and the weird stuff seems to be gone. -- Except for the `RTCube`, where I had defined `relativeTol` but never used it.

I'll do some more testing and if I won't find anything alarming I'll, push some updates.

I'll also try to come up with some simple but effective way of making sure that the do-loop, mentioned above will always be terminated one way or another. I am not convinced at all that the numerical conditions that lead to _infinity_ loop would never be possible again. My experiences with AoI warn me otherwise. -- As the eventual culprit that caused the loop is `findNextNode`, which starts to find the node itself instead of the next one, the confirmation that the found node actually is the next one should be built in there. That should also clean any possible remaining quality issues off.

peteihis:
I think I'm done unless somebody finds something wrong with it. That resolves #193 except for the new finding in the sea scene. My comments:

- I have not checked to detail how it now responds in the smaller ens of the scale but along the way problems started to gradually  emerge after 1E-4-dimensions. Sometimes you could go as far as 1E-13.

- There are some white space changes in the first commit too, but mostly they are in the latter one. 

- For some reason there are a lot of places where git seems to get confused about what is new and what is not. -- I guess the files need to be compared side by side in an editor then.

- Now the actual "power fix" to the pixel problems and the infinity loops is the new `do`-loop in `OctreeNode.findNextNode` starting on line 263. The function is that if the "next node" is the same one that you are on, then extend the margins and retry. In the methods that call `finXXXNode` there are safety switches that will prevent looping if the size dependent tolerances fail and start to return the same node.

- Also, just a little detail, but though it'd look pretty powerful the code  to write `y = Math.max(Math.max(a, b, ), c)*x;` invoking `Math.max` seemed take a much longer time than a couple or consequential `if (a > b)`s and separate multiplication. I tested that only by scripts but I rewrote my earlier Math.maxes to if statements.

peastman:
I want to take some time to think about this.  The overall approach you've taken looks good, but the implementation has issues.  These changes are likely to have a significant impact on performance.  The bottleneck to raytracing performance is often memory rather than computation.  That's why I changed OctreeNode to store all its fields in single precision.  That reduces the memory it takes up, which improves performance.  You've added two new fields to it, and double precision ones at that.  Likewise you've added new fields to all the RTObject subclasses.  All of that is going to hurt performance.

peteihis:
Sure. 

I was aware that added capability would be exchanged to reduced performance to some extent. Then it is a question of "which wolf to feed" after the most critical issues are taken care of. All of those changes do have a purpose, but it does not necessarily mean that they could not be improved.  Suggestions are welcome.

A view from the users side just for background: I have ran into all of there issues before. It was always frustrating and infuriating to run into an invisible limit to which there did not seem to be any reason in the model. However, those were never _that_ critical and soon I pretty much forgot them -- either found cheat around it or just gave up..... But this time I ran to these issues doing real work for my employer and their client. Originally just to do a study that was best done by tracing rays, rather than manually modelling pieces in CAD, but then also to explain what exactly was done. In that it'd be useful to do a few images and possibly an animation of the process using the actual modeled environment form the CAD. And that is big.

This is where a little drop of performance is insignificant in comparison to _not_ having to worry if the rendering will ever get finished or if there will be holes in the model. Retrying over and over, trying to figure out, what the problem was this time and hoping for the best is nerve wrecking and consuming working time in the way that is just not justified. It has to work. Reliability is what counts.

While I was working on it, it did cross my mind that if the fixes turn out to be very heavy, it might be an option to have these changes as a "special cases renderer"  plugin an do just a necessity set in the built in one, but that also might seem a bit confusing to users.... 

_And BTW, my project manager seemed happy with my results so far. Reporting to the client some time after holidays... :)_

peastman:
Can you post a set of test scenes that demonstrate each of the problems?

Sailsman63:
Any differences between `Math.max(Math.max(a, b), c)` and Chained `if`s would need to be checked with a special benchmark test in pure java. Scripts are likely to add enough additional indirection that you can't tell the difference correctly...

Also, I thought we were intending to keep the changes for the "Primitive Tolerance" issue separate from the ones for the Octree intersection miss, as they really are two separate bugs?

peteihis:
> Can you post a set of test scenes that demonstrate each of the problems?

@peastman: I will but I'm now away from my development set for a week or so. Some of the cases will reguire some explaining as in what the SW set it where it reveals a problem. The first cases were described in the #193-thread, but I had some that did not explain there.

@Sailsman63: I could make them separate commits but now that I actually managed to track down the rootcauses to all of them, I'd choose to keep them together. They are closely related, basically result of the same structural issue in the back and I believe they'd be best optimized if they are kept together.

The RTObject tolerances really just push the limits where you can operate a litte larger, but the more serious reliability issue would still remain.

> but the implementation has issues

Back to this: I see some redundacy there now as I can approach it from inside to out. The double confirmation to prevent looping is unnecessary. The Octreenode fields could be single precision. One of them coule be calculated from the other one. Actually all of the added fields _could_ be turned into calculations (but then the values would need to be recalculated every time they are needed again). I think it would be possible to come up with somethig, that just caculates the value without ifs and maxes to make it faster.

peteihis:
Test cases attached!  - And sorry for the time it took. I have had most of these ready for a couple of months already. 

[Raytracer test cases.zip](https://github.com/ArtOfIllusion/ArtOfIllusion/files/5324058/Raytracer.test.cases.zip)

The test cases are all 1 second long movies that will have the raytracing process glitch somewhere within the second. Some of them will fail from beginning without some initial fixes to the code.

I am not sure what happened with the last two cases. What they are for did happen at some point of the process but I could not come up with the a setup that would produce the problem any more.

Sailsman63:
Just ran this branch to test against the example file for #239.

This branch added some glitchies that did not exist with the standard raytracer, so we will need to figure out some more solid testing.

peteihis:
> example file for #239.

Ehh... #239 is the anti-aliasing issue, there was no example file in there.

With this PR I worked from surface towards the root cause(s). It was the typical case that fixing one trouble immereiately caused a different one. Different combinations of corrections created new surprises.... 

I think I'd better start another PR starting from the root cause and then seeing, what other fixes might be worth keeping and what not. Fixing only the root casues (= returng the same `nextNode()` as last time) would lead to longer redering times when the first attempt to find a node fails, but I can not say how severe the effect would be. It never happened that all pixels suffered the issue, just a percentage of them, probably some 20-30 % even at worst.

And the branch seems to require rebasing.

Sailsman63:
Argh! I meant #236

Also, if you're re-looking at this, please separate out the prison fixes for primitive objects. I think I could review and merge them fairly quickly if I could see them separately.

peteihis:
> Just ran this branch to test against the example file for **#236**.

> This branch added some glitchies that did not exist with the standard raytracer, so we will need to figure out some more solid testing.

This branch should not change what happens with that test file. The problem exists with 3.1.0 and the result is exactly the same as with this one (just tested). If you actually see a difference, let me know.

> **Sailsman63:** Also, if you're re-looking at this, please separate out the prison fixes for primitive objects. I think I could review and merge them fairly quickly if I could see them separately.

>  **peastman:** .... You've added two new fields to it, and double precision ones at that. Likewise you've added new fields to all the RTObject subclasses. All of that is going to hurt performance.

So basically the fixes to the RT-objects would hurt performance in one way, but the other option would be to do run more calculations and repeating subroutines, which also hurts performance...

I guess, I'll try to find a way somewhere in the middle (doing some calculations locally instead of using fields). Separate PRs for the root cause and the RT-fixes.

Sailsman63:
> If you actually see a difference, let me know.

I do. With this patch I see some spurious pixels outside, but surrounding, the cylinder. These only show up in a few frames (basically, I watch the rendering process) This seems to come and go. I was not able to replicate just now, but I saw it earlier today.

> So basically the fixes to the RT-objects would hurt performance in one way, but the other option would be to do run more calculations and repeating subroutines, which also hurts performance...

> I guess, I'll try to find a way somewhere in the middle (doing some calculations locally instead of using fields). Separate PRs for the root cause and the RT-fixes.

You're mixing two different (related) bugs. I was referring to the scaling of TOL by the dimensions of the primitve objects, which will _not_ impact performance in any meaningful way, and has nothing to do with a fix for the triangle/mesh objects.

peteihis:
> ... some spurious pixels outside, but surrounding, the cylinder

A kind of fit with the picture. RTCylinder does not work right when it is very stretched or flattened. Bump maps also play their part in it (which they should not).

> I was referring to the scaling of TOL by the dimensions of the primitve objects

That was one of the questions in my mind. The fix does add a field (two in case of Cylinder) to each RT primitive but I'd expect scenes mostly to be rendered from RTTriangles. Above RT objects were mentioned generally though. -- I'll separate those and work the rest from depths up.

peteihis:
Ookay....

I downloaded the test cases zip back and found out that it was empty!

Right....

PR #197, opened 2020-07-04
Math.round to Math.rint in artofillusion.ui -- bug fix

peteihis:
I discovered that `ValueField` changed anything larger than 9223372036854776.0 into 9.223372036854776E15, when reading values from objects.

The problem was that the process used `Math.round()`, which returns `long` and its  `MAX_VALUE` is 9223372036854776. `Math.rint()` returns the value as `double`.

This changes all the found rounds into rints in `artofilusion.ui`. In some cases it does not really make any difference, but in principle it is the correct operation. I checked, that the value is never handled as `long` after the rounding. It is either becomes `double` again in the next operation or is converted `(int)`.

Did not check the rest of the code, but I suspect that there is a large population of `Math.round` all over it, when the expected return value would still be be `double`....

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ui/Compound3DManipulator.java
+++ b/ArtOfIllusion/src/artofillusion/ui/Compound3DManipulator.java
@@ -810,9 +810,9 @@ public class Compound3DManipulator extends EventSource implements Manipulator
         drag = view.getCamera().findDragVector(dragStartPosition, disp.x, disp.y);
       if (isShiftDown)
       {
-        drag.x = gridSize*Math.round(drag.x/gridSize);
-        drag.y = gridSize*Math.round(drag.y/gridSize);
-        drag.z = gridSize*Math.round(drag.z/gridSize);
+        drag.x = gridSize*roundValue(drag.x/gridSize);
+        drag.y = gridSize*roundValue(drag.y/gridSize);
+        drag.z = gridSize*roundValue(drag.z/gridSize);
       }
       Mat4 transform = Mat4.translation(drag.x, drag.y, drag.z);
       dispatchEvent(new HandleDraggedEvent(view, dragHandleType, dragAxis, bounds, selectionBounds, ev, transform));
@@ -868,7 +868,7 @@ public class Compound3DManipulator extends EventSource implements Manipulator
     if (isShiftDown)
     {
       dragDistance /= gridSize;
-      dragDistance = Math.round(dragDistance);
+      dragDistance = roundValue(dragDistance);
       dragDistance *= gridSize;
     }
     Vec3 drag3D = dragDir3D.times(dragDistance);
@@ -1016,6 +1016,11 @@ public class Compound3DManipulator extends EventSource implements Manipulator
   private Vec2 toVec2(Point start, Point end)
   {
      return new Vec2(end.x - start.x, end.y - start.y);
+  }
+
+  private double roundValue(double val)
+  {
+     return Math.abs(val) > 1.0e15 ? Math.rint(val) : Math.round(val);
   }
 
   /**
--- a/ArtOfIllusion/src/artofillusion/ui/ValueField.java
+++ b/ArtOfIllusion/src/artofillusion/ui/ValueField.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 1999-2007 by Peter Eastman
+   Modification copyright (C) 2020 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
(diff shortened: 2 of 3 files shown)


Sailsman63:
Boy, you sure are stretching the "workable" limits, here! I'm trying to imagine a circumstance when a value slider capacity of around 9 trillion is not large enough... 

Note that `Math.rint()` rounds exact half values (x.5) to the nearest *even* integer, which does mean a behavior change even for me everyday values. Let me stew a little about whether this is preferable to display overflows at a value that I suspect will rarely, if ever, be reached.

Particularly for the manipulator, is there any way that we'll ever get into the high value ranges?

peteihis:
> rounds exact half values (x.5)

Oh. So where `round` rounds x.5 always up, `rint` may round it up or down, which differs from the school convention. I guess the reason is somewhere in the bits level... A calculatory fact, that did not exist, when the classical approach was created. Actually, how rint works, may be closer to the statistical approach, where you were supposed to look at the next two digits instead of one, to decide which way to round (don't remember the exact rule any more) and round exact point-5's every second time up and the other times down to avoid statistical distortion. 

I can see that that the behavior of rint might be confusing at least with number fields, when you are working with smallish - say < 100000 values.

I was considering a more complex approach either by scaling "extra large" (> 1e6) values in a different way and/or cutting and pasting the string to be displayed but rint looked like the perfect solution.

In the manipulator rotation angles will never be that big but coordinates may well be: It is possible to create objects larger than Long.MAX_VALUE, display them in wireframe mode and move with the manipulator. The grid step (shift down move) does not then work right when there is the `Math.round` in the equation.

peteihis:
Now thinking, it might work to switch to rint when the abs(value) is >1e15 and use round by default and leave the angles and trigonometrics as they are.

peteihis:
And move object and move view don't work either when you fit the view to something in 1e20 size... Hmmm.

peteihis:
I think this would be more reasonable. 

I still don't know why move tools stop working in those sizes. Some strange flickering starts to appear already with slightly smaller values.

Sailsman63:
Can you point me at a specific location where I can check this?

peteihis:
> Can you point me at a specific location where I can check this?

Location? You mean the move object / move view tools?

The camera distance from origin is probably a factor but there is no special location. I just drew a cube in the origin and fit the view to that. When the cube edge length is 1E15 units, there is some jumpiness in the movements, especially if you rotate the view a bit first (some 45° left/right & 30° up maybe). If you set the cube size to 1E16 and fit the view to that, the tools stop working. -- I did not try to define the exact limit, where it happens but it is not the size of the cube that causes it: Zooming closer to about  half the distance all of a sudden releases the lock.

Of course viewing has to be is in wire frame mode, to draw objects that big.

Sailsman63:
I was thinking more of places where I can see a `ValueField` that is affected by this change. This merely adjusting how large numbers are displayed in the UI, is it not?

peteihis:
> The width of text fields is always a tradeoff. You want it to be wide enough for the values people will typically be entering. Making it much wider than that wastes screen space.

Of course I'm aware of that and I'm not suggesting extreme. 

However it is not uncommon that even relatively usual values are too wide for the fields. Sometimes (when the size is not critical) even changing the text field width form 5 to 7 would help a lot basically so that a number with three decimals and minus sign would fit.  _(I don't know how that value actually defines the width. Usually a quite bit more than the given digit count actually fits in.)_

When it comes to wasting screen space I'd say that generally speaking it is not the width of text fields that is the problem Quite often it is just the general arrangement that leaves empty bare background on dialogs. Of course that can not always avoided, as the UI also has to have some kind logic to the arrangement of elements. Also compressing wasted space would easily mean a pretty thorough redesign of the UI to replace the easy (and definitely very helpful) 'semi products' that AoI provides.

The latest fashion seems to be that dialogs come without a background. The buttons and fields are just cloud of "floating bricks" on the screen and sometimes with content that is changing according to the situation.... I can think of one word that in most cases describes those brick clouds comprehensively: "Disaster".

Sailsman63:
Okay, now I see the "limit to `long.MAX_VALUE` bug.

I only see it when editing through the modal dialog, not when using the Object Properties panel. Also, If the modal dialog clips the value like this, *but not if I enter a smaller, normal value,* The `ObjectPropertiesPanel` does not update the shown value.

The discussion on good field widths needs a whole separate discussion...

Sailsman63:
And even with this fix, something goes weird with the manipulator at those high scales... I think we can do the UI elements, but the change to the manipulator probably is just masking something deeper.

PR #200, opened 2020-07-14
Camera distance value copy fix

makiam:
Fix SceneCamera copy code to properly copy distToPane value from source to target.
Also tests for SceneCamera object and camera's filter FilterParameterTrack

### DIFF
--- a/Tests/src/artofillusion/object/SceneCameraTest.java
+++ b/Tests/src/artofillusion/object/SceneCameraTest.java
@@ -1,0 +1,504 @@
+/* Copyright (C) 2020 by Maksim Khramov
+
+   This program is free software; you can redistribute it and/or modify it under the
+   terms of the GNU General Public License as published by the Free Software
+   Foundation; either version 2 of the License, or (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful, but WITHOUT ANY 
+   WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A 
+   PARTICULAR PURPOSE.  See the GNU General Public License for more details. */
+
+
+package artofillusion.object;
+
+import artofillusion.Camera;
+import artofillusion.Scene;
+import artofillusion.animation.FilterParameterTrack;
+import artofillusion.animation.PoseTrack;
+import artofillusion.animation.WeightTrack;
+import artofillusion.image.ComplexImage;
+import artofillusion.image.filter.ImageFilter;
+import artofillusion.math.BoundingBox;
+import artofillusion.math.CoordinateSystem;
+import artofillusion.test.util.StreamUtil;
+import java.io.DataInputStream;
+import java.io.DataOutputStream;
+import java.io.IOException;
+import java.io.InvalidObjectException;
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+import org.junit.Assert;
+import org.junit.Test;
+
+/**
+ *
+ * @author MaksK
+ */
+public class SceneCameraTest
+{
+  
+  @Test
+  public void testGetCameraBounds() {
+    SceneCamera camera = new SceneCamera();
+    BoundingBox cBox = camera.getBounds();
+    Assert.assertEquals(-0.25, cBox.minx, 0);
+    Assert.assertEquals(0.25, cBox.maxx, 0);
+    
+    Assert.assertEquals(-0.15, cBox.miny, 0);
+    Assert.assertEquals(0.20, cBox.maxy, 0);
+    
+    Assert.assertEquals(-0.20, cBox.minz, 0);
+    Assert.assertEquals(0.20, cBox.maxz, 0);
+  }
+  
+  
+  @Test
+  public void testCreateSceneCamera() {
+    
+    
+    SceneCamera camera = new SceneCamera();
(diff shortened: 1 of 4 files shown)


Sailsman63:
Can you point to what happens without this fix? What does the use see that is wrong?

peteihis:
> Can you point to what happens without this fix?

The camera that the values are supposed to be copied from would get the value from the camera that they were supposed to be copied to. This line should be doing the opposite to what is done in the `duplicate()` mode above it. I seem to have just copied the line and not reversed the action.

makiam:
copy() function in general used to restore edited object values from Undo/Redo stack. As the camera's distToPane is not editable this fix has no good visible effect. (I just found this with unit test)

peteihis:
> As the camera's distToPane is not editable this fix has no good visible effect.

Actually it does. `distToPlane` is the working depth of  the view. The rotation center of the view is at this distance from the camera and any mouse actions are assumed to happen at this distance. You "edit" the value by zooming the view in or out or centerig it to a new object or point on an object surface. 

If the value is wrong you see it in the behavior or the view. The view just will not do what you expect. It rotates in predictably and objects will be drawn in wrong scale and in wrong depth. It is really THE parameter in the logic of view manipulation.

And thanks for the finding! I had missed it entirely.

Sailsman63:
That's about what I thought. Good catch.

Tests/src/artofillusion/object/SceneCameraTest.java
```diff
@@ -0,0 +1,504 @@
...
+   Foundation; either version 2 of the License, or (at your option) any later version.
+
+   This program is distributed in the hope that it will be useful, but WITHOUT ANY 
+   WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A 
+   PARTICULAR PURPOSE.  See the GNU General Public License for more details. */
+
+
+package artofillusion.object;
+
+import artofillusion.Camera;
```
Sailsman63, line 14: Please collapse imports of more than two or three classes in the same package.

Tests/src/artofillusion/object/SceneCameraTest.java
```diff
@@ -0,0 +1,504 @@
...
+
+/**
+ *
+ * @author MaksK
+ */
+public class SceneCameraTest
+{
+  
+  @Test
+  public void testGetCameraBounds() {
```
Sailsman63, line 41: Is there anything that would break if we changed the size or geometry of the visual mesh that is used to represent a `SceneCamera`? If not, I'd submit that we have no need to run such a test, and the test will just make it harder to change the visuals in the future if we need to.

Tests/src/artofillusion/object/SceneCameraTest.java
```diff
@@ -0,0 +1,504 @@
...
+  }
+  
+  
+  @Test
+  public void testCreateSceneCamera() {
+    
+    
+    SceneCamera camera = new SceneCamera();
+    ObjectInfo camObj = new ObjectInfo(camera, new CoordinateSystem(), "Camera");
+    Assert.assertTrue(camera.isPerspective());
```
Sailsman63, line 61: Which leaves this as the one assertion that is actually specific to a new `SceneCamera`

So. Is the fact that a new `SceneCamera` defaults to Perspective mode significant and important to test for?

If so, we should rename the test "newCameraPerspectiveByDefault()" or similar, to reflect what we are actually testing.

Sailsman63: Okay, better. I'm still not sure why it's important.

Tests/src/artofillusion/object/SceneCameraTest.java
```diff
@@ -0,0 +1,484 @@
...
+    SceneCamera camera = new SceneCamera();
+    Assert.assertTrue(camera.isPerspective());
+
+    
+  }
+  
+  @Test
+  public void getEmptySceneCameraFilters() {
+    SceneCamera camera = new SceneCamera();
+    Assert.assertTrue(camera.isPerspective());
```
Sailsman63, line 52: This we've already tested. No need to assert again.

Tests/src/artofillusion/object/SceneCameraTest.java
```diff
@@ -0,0 +1,484 @@
...
+  public void сreateSceneCameraFromStreamDoubleGoodFilter() throws IOException {
+    ByteBuffer wrap = ByteBuffer.allocate(200);
+    wrap.putShort((short)1); // Base Object3d version
+    wrap.putShort((short)3); // Scene Camera object version
+    wrap.putDouble(123); // distToPlane
+    wrap.putDouble(45); // fov
+    wrap.putDouble(100); // depthOfFieled
+    wrap.putDouble(50); // focalDistance;
+    wrap.put((byte)1);  // Perspective Boolean treats as byte 
+    wrap.putInt(2); //One Camera filter
```
Sailsman63, line 392: Do we really want to be exercising the format-version code with hand-constructed byte streams? Is there a better way to test this, or should be be bothering at all?

@peastman, I'd appreciate some input.

Sailsman63:
I'm thinking about cherry-picking 559d652, to get it into our next nightly. The test require some further thought and review.

PR #201, opened 2020-07-23
Bug fix to Cylinder pose track keyframe editor

peteihis:
On the keyframe-editor of Cylinder Pose track the radii fields were set to display size, not radius. The fields were called 'Bottom Radius (X)' and '-(Z)' though.

I have sometimes though that there something seemed funny about how cylinder behave, but this was it. The object editor is displaying the radii as it should.


Sailsman63:
Thanks! Good catch!

PR #203, opened 2020-08-11
Syntax editor

makiam:
Common script editors code moved to base class. Old Editor control replaced with new RSyntaxTextArea.
Added new library jar


ArtOfIllusion/src/artofillusion/script/ScriptWindowBase.java
```diff
@@ -0,0 +1,190 @@
...
+import javax.swing.filechooser.FileNameExtensionFilter;
+import org.fife.ui.rsyntaxtextarea.RSyntaxTextArea;
+import org.fife.ui.rsyntaxtextarea.SyntaxConstants;
+import org.fife.ui.rtextarea.RTextScrollPane;
+
+/**
+ *
+ * @author MaksK
+ */
+public class ScriptWindowBase extends BFrame
```
Sailsman63, line 40: Not a fan of this approach. As originally implemented, the script editor is simply a widget that provides the text editor functions. It composes well with any window arrangement that you want to embed it in.

As you've written this, one has to extend the `ScriptWindow`. Favor composition over inheritance.

ArtOfIllusion/src/artofillusion/script/ScriptedObjectEditorWindow.java
```diff
@@ -22,7 +22,7 @@ Changes copyright (C) 2020 by Maksim Khramov
 
 /** This class presents a user interface for entering object scripts. */
 
-public class ScriptedObjectEditorWindow extends ScriptWindowBase
+public final class ScriptedObjectEditorWindow extends ScriptWindowBase
```
Sailsman63, line 25: Why the change to final?

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
```diff
@@ -18,7 +18,7 @@ Changes copyright (C) 2020 by Maksim Khramov
 
 /** This class presents a user interface for entering scripts to be executed. */
 
-public class ExecuteScriptWindow extends ScriptWindowBase
+public final class ExecuteScriptWindow extends ScriptWindowBase
```
Sailsman63, line 21: And here?

ArtOfIllusion/src/artofillusion/script/ScriptedObjectEditorWindow.java
makiam: This way (Plus ScriptWindowBase is marked as package-private) I prevent unintended extension  this classes.

Sailsman63: Again, why?

If this is in response to the concern about the structure that I expressed the first time through, you seem to have misunderstood me...

makiam: Just not found good solution with composition style against inheritance style

peastman: Why do you want to prevent people from extending it?

The point was that if the editor is a widget, anyone can embed an editor inside any window.  But if the editor is a window, it can only be used in that window and no other.

makiam: As @Sailsman63 wrote

> As you've written this, one has to extend the ScriptWindow

I close this views (Script editor and ScriptObject editor) from extending nd reduce visibility of their base class. 
... Maybe I do not understand the idea in general

Sailsman63: I can run a PR against your branch to show you what I mean. It may be a few days.

PR #206, opened 2020-08-25
Use same API name and signature as all other AOI objects. (writeToFil…

makiam:
…e vs writeToStream ) Simplified scene images save code

### DIFF
--- a/ArtOfIllusion/src/artofillusion/image/ExternalImage.java
+++ b/ArtOfIllusion/src/artofillusion/image/ExternalImage.java
@@ -378,12 +378,12 @@ public class ExternalImage extends ImageMap
   }
 
   /** @deprecated <br> 
-      Use writeToFile(DataOutputStream out, Scene scene) instead. <p>
+      Use writeToStrean(DataOutputStream out, Scene scene) instead. <p>
       This will save the ExternalImage to the stream, but the relative path will 
       be replaced by absolute path as the path to the saved scene is missing */
 
   @Deprecated
-  public void writeToFile(DataOutputStream out) throws IOException
+  public void writeToStream(DataOutputStream out) throws IOException
   {
     writeToStream(out, null);
   }
--- a/ArtOfIllusion/src/artofillusion/image/ExternalImage.java
+++ b/ArtOfIllusion/src/artofillusion/image/ExternalImage.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 2017 by Petri Ihalainen
+   Changes copyright (C) 2020 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -12,7 +13,6 @@ package artofillusion.image;
 
 import artofillusion.*;
 import artofillusion.math.*;
-import buoy.widget.BStandardDialog;
 import java.awt.*;
 import java.awt.image.*;
 import java.io.*;
@@ -21,8 +21,7 @@ import javax.imageio.*;
 
 public class ExternalImage extends ImageMap
 {
-  private File externalFile;
-  private String loadingError, imageType;
+  private String imageType;
   private ImageMap imageMap;
   private int w, h;
   private String type, lastAbsolutePath, lastRelativePath;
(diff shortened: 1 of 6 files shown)


peastman:
Given that they write to a `DataOutputStream` instead of a `File`, I think `writeToStream()` is a better name.  It also is currently the more common name: it looks like we have 50 methods called `writeToFile()` and 91 called `writeToStream()`.  So I would make all of them use that.

ArtOfIllusion/src/artofillusion/image/ExternalImage.java
```diff
@@ -384,19 +378,18 @@ public ExternalImage(DataInputStream in) throws IOException
   }
 
   /** @deprecated <br> 
-      Use writeToStream(DataOutputStream out, Scene scene) instead. <p>
+      Use writeToFile(DataOutputStream out, Scene scene) instead. <p>
```
Sailsman63, line 381: Missed one on standardizing on `writeToStream`

ArtOfIllusion/src/artofillusion/image/ExternalImage.java
```diff
@@ -384,19 +378,18 @@ public ExternalImage(DataInputStream in) throws IOException
...
       This will save the ExternalImage to the stream, but the relative path will 
       be replaced by absolute path as the path to the saved scene is missing */
 
-  @Override
   @Deprecated
-  public void writeToStream(DataOutputStream out) throws IOException
+  public void writeToFile(DataOutputStream out) throws IOException
```
Sailsman63, line 386: And especially this line...

ArtOfIllusion/src/artofillusion/image/ImageMap.java
```diff
@@ -251,5 +253,5 @@ public void setName(String newName)
       
       which reconstructs an image from its serialized representation. */
 
-  public abstract void writeToStream(DataOutputStream out) throws IOException;
+  public abstract void writeToStream(DataOutputStream out, Scene scene) throws IOException;
```
Sailsman63, line 256: Changing the Public API of one of our core classes. I think I see why, but do we know if this would break anything?

@peastman

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/image/ExternalImage.java
```diff
@@ -21,8 +21,7 @@
 
 public class ExternalImage extends ImageMap
 {
-  private File externalFile;
-  private String loadingError, imageType;
+  private String imageType;
```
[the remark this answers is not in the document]
makiam: imageType used both in read and write methods for this class

ArtOfIllusion/src/artofillusion/image/ImageMap.java
makiam: As AOI has no Plugin API to connect other image formats I think this is no-breaking change

ArtOfIllusion/src/artofillusion/image/ExternalImage.java
```diff
@@ -384,19 +378,18 @@ public ExternalImage(DataInputStream in) throws IOException
   }
 
   /** @deprecated <br> 
-      Use writeToStream(DataOutputStream out, Scene scene) instead. <p>
+      Use writeToStrean(DataOutputStream out, Scene scene) instead. <p>
```
Sailsman63, line 381: Spelling.

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/image/ImageMap.java
peastman: It's impossible to guarantee it won't break something, because we don't know what other code might be calling this.  The question is how important we think that is.

Sailsman63: This is part of the serialization code, though. Not sure I've ever run across a Plugin or Script that wanted to serialize *part* of a scene, Not sure I even can think of a realistic use case for that... It's normally called when writing out an entire scene.

I was more concerned about Plugins adding new ImageMap types, but Makiam is right that this is not currently possible (At least without overriding parts of AOI core code.)

Your call.

peastman: I agree it's pretty unlikely to break things.  Probably not worth worrying about.

Sailsman63: Sounds good to me.

PR #207, opened 2020-08-25
Made Scene textures and materials available as lists.

makiam:
+ Small update of code with modern style


Sailsman63:
I could see some instances where having the material and textures available as lists might be handy. Did you have any particular points in mind, it is this just in general principles?

Also, again: this PR covers two separate concerns. They should at least have been separate commits...

makiam:
I try to create some UI Plugin to show textures/materials. It is possilbe to take it from scene one by one but get its as list looks better.

Sailsman63:
Okay, that's about what I thought. This looks good to me.

PR #208, opened 2020-08-28
Update Scene load errors code

makiam:


ArtOfIllusion/src/artofillusion/Scene.java
```diff
@@ -1159,21 +1165,6 @@ public Texture getDefaultTexture()
...
-    return errorsLoading;
-  }
-
-  /** Get a description of any errors which occurred while loading the scene. */
-
-  public String getLoadingErrors()
-  {
-    return (loadingErrors == null ? "" : loadingErrors.toString());
-  }
-
```
Sailsman63, line 1176: My one concern here is that these are public methods - just removing them could break existing scripts or plugins.

I'd rather see these two methods `@Deprecated` (with a note to use `getErrors()`) and then change the implementation to use `getErrors()` internally.

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/Scene.java
makiam: I did not find any usages of this methods across all plugins/scripts from aoisp plugin portal. So in my opinion it is safe enough removing this.

Sailsman63: On further consideration, I think that this one is safe.

PR #209, opened 2020-08-28
Scene assets load update

makiam:
Scene assets load update. Use common method to load both materials and textures

### DIFF
--- a/ArtOfIllusion/src/artofillusion/Scene.java
+++ b/ArtOfIllusion/src/artofillusion/Scene.java
@@ -1209,7 +1209,6 @@ public class Scene
     initFromStream(in, fullScene);
   }
 
-  @SuppressWarnings("UseOfObsoleteCollectionType")
   private static Vector<Named> loadSceneAsset(DataInputStream source, Scene scene, Supplier<Named> stub) throws IOException
   {
       int counter = source.readInt();
--- a/ArtOfIllusion/src/artofillusion/Scene.java
+++ b/ArtOfIllusion/src/artofillusion/Scene.java
@@ -45,11 +45,17 @@ public class Scene
   private TextureMapping environMapping;
   private int gridSubdivisions, environMode, framesPerSecond, nextID;
   private double fogDist, gridSpacing, time;
-  private boolean fog, showGrid, snapToGrid, errorsLoading;
+  private boolean fog, showGrid, snapToGrid;
   private String name, directory;
 
   private ParameterValue environParamValue[];
-  private StringBuffer loadingErrors;
+
+  private final List<String> errors = new ArrayList<>();
+
+  public List<String> getErrors()
+  {
+    return Collections.unmodifiableList(errors);
+  }
 
   public static final int HANDLE_SIZE = 4;
   public static final int ENVIRON_SOLID = 0;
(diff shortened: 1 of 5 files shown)


peastman:
Could you explain the design and reasons for this change?  What is `setNameFluent()` for?

makiam:
This is a just a syntax sugar to write object initialization code this way..
>>`Supplier<Named>> stub`
...
>> `stub.get().setNameFluent('<unreadable>')`

peastman:
What is "fluent" supposed to mean?

makiam:
https://en.wikipedia.org/wiki/Fluent_interface

ArtOfIllusion/src/artofillusion/Scene.java
```diff
@@ -1208,8 +1209,59 @@ public Scene(DataInputStream in, boolean fullScene) throws IOException, InvalidO
...
+          result.add((Named)stub.get().setNameFluent("<unreadable>"));
+          scene.errors.add(Translate.text("errorInstantiatingClass", className));
+        }
+      }
+      return result;
+  }
+
   /** Initialize the scene based on information read from an input stream. */
 
+  @SuppressWarnings("UseOfObsoleteCollectionType")
```
Sailsman63, line 1264: Whatever else we decide about this change, please reserve `@SuppressWarnings` for instances where we *cannot* change to a newer collection class/structure/safe cast/etc.

And even then, It should probably go at the line that uses the 'iffy' code, with a comment that explains *why* we have to stick to the old way...

Sailsman63:
Alright, finally got some time to consider this.

- Fluent: I kind of suspect that most fluent interfaces don't actually have "Fluent" in their names. An example of a fluent interface in java would be `java.util.stream.Stream`, in particular all of the methods that return a `Stream` (And arguably, those that return one of the specialized Stream variants) 
- I can see that you are trying to unify the loading of assets, but this `Named` Interface just seems clumsy. It's not in itself a self-explaining abstraction, especially when it's bolted on this way.
- If we were to try to unify the loading of assets, we're essentially saying that `Textures`, `Materials` or any other asset type are basically always mirrors of each other. This may not always be the case.

makiam:
For now I use setNameFluent because changing setName method return type gives me conflict with ProcedureOwner interface which also declares name getter and setter methods. In general ProcedureOwner looks responsible for many things which is not good IMO.

makiam:
For future...  Apply Named interface for all objects with name getter/setters.

Sailsman63:
Let me phrase this another way:

What purpose does the abstraction "named" serve? Is there any case where we would want to operate on a collection of "named" things, without considering what else they might be?


If you've got concerns about `ProcedureOwner` in general, please open a separate issue to explain them.

makiam:
Every object marked with this interface may return it's name and can change it (Which is used in this particular case)

Sailsman63:
I think we're having a serious communication issue here...

What makes this significant enough to justify a separate interface? Again, is there any use distinguishing between items that are "named" and those that are not?

makiam:
Old code for loading textures/materials has 2 points where object name are set. (One for fallback texture and one for material). Once I write common code to load both asset types I need common denominator to set name. The Named interface serves as this denominator.

peastman:
I don't think I've ever encountered a method with the name "fluent" in it before.  Most people will have no idea what it refers to.  I certainly didn't.  And if you want an API where calls can be chained together like that, you need to make all setter methods do it.  Having just one that can be chained when none of the others can doesn't help anyone.  It just makes the API confusing.

> Old code for loading textures/materials has 2 points where object name are set.

I could see an argument for creating a common interface implemented by both textures and materials.  Names are only a small part of what they have in common.  They also have IDs, can use ImageMaps, can be saved to files, etc.  If we want to consider that, let's enumerate everything that interface might include.  Then we need to figure out whether there's actually enough benefit to justify it.  If you almost always know whether you're working with a Texture or a Material, then it doesn't make much difference.  But if there's a lot of code that would benefit from being able to accept either Textures or Materials, then it might make sense.

Sailsman63:
These all got duplicated in #212, continue any discussion over there.

PR #211, opened 2020-09-08
Harden test implicit resolves #67

### LINKED ISSUE #67: ImplicitSphereTest rounding errors?
https://github.com/ArtOfIllusion/ArtOfIllusion/blob/0eddacaab0b8465085da35a45951496b6687206c/Tests/src/artofillusion/object/ImplicitSphereTest.java#L51-L53

In some recent runs of the unit tests (working on fixing the test suite) I just got the following errors:

`[junit] expected:<1.125680191712794> but was:<1.1256157239630205>`
on line 52 and, on a susequent run,
`[junit] expected:<-0.07972279469479109> but was:<-0.07970200347745438>`
on line 51.

This test randomly generates the radii for each test. This looks like rounding/overflow errors to me, but without knowing what the input numbers are, I cannot be sure. The test should at least print `r1` & `r2` values for the test sphere, and `x`, `y` & `z` values for that run of the test method.

Sailsman63:
As discussed in #67 
Well, this took me down a deeper rabbit hole than I was expecting. I've done the following:

- Restructured pretty much the entire test class.
  + I broke the various tests out into separate test methods, to improve localizing any failures. The test case name should tell exactly what kind of unexpected we've run into.
  + Added a readable printout of the <sphere, point> pair that failed the test
  + This would be much easier if we had a property testing framework (plugin to junit 5, but I'm not going there right now)
- Fixed two separate false failure modes, both as expected from the issue discussion
  + Scaled the allowable delta between the estimated and returned gradients proportional to the returned gradient
  + At the edge of the influence radius, some of the samples for the estimated gradient might be inside the influence, and some outside. This seriously skews the estimate. Instead of trying to do some really fussy bracketing, or trying to space the sample points tighter, I threw in the towel and just make sure that the estimate and the returned gradient are in the same octant of the sphere. (Make sure each of the components match in sign - pointing roughly in the right direction)

Was this overboard? Possibly. Unnecessary? Probably.

Learned a few things, and by the time I realized how much surgery I was doing, it was mostly done except for cleanup.

### DIFF
--- a/Tests/src/artofillusion/object/ImplicitSphereTest.java
+++ b/Tests/src/artofillusion/object/ImplicitSphereTest.java
@@ -103,16 +103,22 @@ public class ImplicitSphereTest
         double radius = sphere.getRadius();
         double value = sphere.getFieldValue(point.x, point.y, point.z, 0, 0);
         sphere.getFieldGradient(point.x, point.y, point.z, 0, 0, grad);
-        double step = radius*1e-3;
+        double step = radius*1e-4;
         double vx1 = sphere.getFieldValue(point.x-step, point.y, point.z, 0, 0);
         double vx2 = sphere.getFieldValue(point.x+step, point.y, point.z, 0, 0);
         double vy1 = sphere.getFieldValue(point.x, point.y-step, point.z, 0, 0);
         double vy2 = sphere.getFieldValue(point.x, point.y+step, point.z, 0, 0);
         double vz1 = sphere.getFieldValue(point.x, point.y, point.z-step, 0, 0);
         double vz2 = sphere.getFieldValue(point.x, point.y, point.z+step, 0, 0);
-        assertEquals(0.5*(vx2-vx1)/step, grad.x, 1e-2*value);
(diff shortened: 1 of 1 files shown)


Sailsman63:
Can someone check to make sure I didn't break the math?

So far, this iteration seems to be stable when run as part of the test suite.

peastman:
A bit of clarification on the finite difference estimation.  The original code was  this:

```java
assertEquals(0.5*(vx2-vx1)/step, grad.x, 1e-2*value);
```

The two points are separated from each other by `2*step`: one is at `x-step` and the other at `x+step`.  That's what the factor of 0.5 comes from.  It might be slightly clearer if it were written `(vx2-vx1)/(2*step)`.

In the new version you left out the factor of 0.5 in `estimateGradient()`, but multiplied the result by 0.5 in the assertion in `gradient_estimate_within_delta()`.  That's mathematically correct but confusing.

Otherwise this looks good!

Sailsman63:
Thanks! I did realize that it would be clearer as you suggested, I'll fix before merging.

PR #212, opened 2020-09-08
More Named interface usages

makiam:
More Named interface usages
Remove fluent suffix from Named intrface
Remove name getter/setter and canEditName methods from ProcedureOwner interface
Replace usage of canEditName with instanceof Named  idiom

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
+++ b/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
@@ -12,13 +12,11 @@
 
 package artofillusion;
 
-import artofillusion.animation.*;
 import artofillusion.image.*;
 import artofillusion.image.filter.*;
 import artofillusion.material.*;
 import artofillusion.math.*;
 import artofillusion.object.*;
-import artofillusion.procedural.*;
 import artofillusion.script.*;
 import artofillusion.texture.*;
 import artofillusion.ui.*;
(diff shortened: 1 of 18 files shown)


peastman:
Before you do any more work on this, we really need to settle the question of whether this is something we want to do at all.  What problem is this supposed to solve?  What are the situations where it's important to be able to manipulate objects of many different types, where the only thing they have in common is having a name?

makiam:
Most important thing: implementing this interface allows me to write single method to load scene assets. I do not expect further changes with this intrface.

Sailsman63:
If this is the only real use for this interface, I'm going to have to say that I'm not a fan.

This does not seem to be an abstraction that makes items in the scene easier to manipulate:
- `Objects` and `ImageMap`s also have names. Why are they not `Named`? If they're not relevant, `Named` is probably not a meaningful distinction.
- The 'Fluent' style of method call that returns an object: outside of the loading code that you've already worked in, where would we use that? I'm not sure I like a situation where we are often ignoring the return value of a method, simply because it's there to support a very narrow use case.

Sailsman63:
Also, a bit of a side point: Please do not base new work on top of unrelated PR branches that are still in process. Both this and #209 are both based on #208. This makes it harder to see the changes that are unique to this PR, and can cause merge problems if we merge #208 in the meantime.

makiam:
> Objects and ImageMaps also have names. Why are they not Named?

Just leaved this as is for now... But of course this may be  marked with Named interface too.

Sailsman63:
I've been considering this in the back of my mind for a while, and I don't feel like this is the way to go.

I'm not adverse to finding some common ground between all AOI 'asset types' and extracting it out, this interface, as conceived, is too narrowly focused to make a good API.

PR #216, opened 2020-09-13
Use new Scene api for textures

makiam:


ArtOfIllusion/src/artofillusion/image/ImagesDialog.java
```diff
@@ -757,6 +745,7 @@ private void selectNone()
       for (BCheckBox b: removeBox) b.setState(false);
     }
 
+    @SuppressWarnings("ResultOfObjectAllocationIgnored")
```
Sailsman63, line 748: What line specifically triggers this warning? As a general rule, it's best to apply 'suppress warnings' for as narrow a chunk of code as possible. (As little as a single statement, when we can.)

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/image/ImagesDialog.java
makiam: this method creates instance of PurgeDialog which is not assigned to any variable. Just opens dialog

Sailsman63: Then the 'suppress warnings' goes down above line 761, to apply only to that statement...

makiam: Annotation cannot be applied to single line of code.

Sailsman63: You may have to add a local variable to assign the new PurgeDialog to... Which will probably stop the warning outright...

Sailsman63:
Thanks!

PR #218, opened 2020-09-26
twist angle optimization recoded

peteihis:
The problem is demonstrated in the zipped .mp4 below. 

[twisanglefixbug.zip](https://github.com/ArtOfIllusion/ArtOfIllusion/files/5286776/twisanglefixbug.zip)

I should have realized that after modifying the angle the CoordinateSystem system needs to be recalculated. Otherwise the outdated CoordinateSystem takes over the next time the bones move.


Sailsman63:
Good catch! Thanks!

PR #219, opened 2020-10-01
Small clean-up of Modules Menu class

makiam:


ArtOfIllusion/src/artofillusion/procedural/ModuleMenu.java
```diff
@@ -336,14 +332,12 @@ public void add(Entry entry)
 
     public Entry(String name, Class moduleClass)
     {
-      this.name = name;
-      this.moduleClass = moduleClass;
-      args = new Object[] {null};
+      this(name, moduleClass, new Object[] {new Point()});
```
Sailsman63, line 335: Wait, why are we putting a`Point` in the array, when the array is supposed to be empty?

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/procedural/ModuleMenu.java
makiam: This array is not empty array but single element array with first null value.
If You see line 226 You see that array checks for null value and then replaces this null to Point()
This way we look for cunstructor with first Point class parameter

PR #221, opened 2020-10-01
Cache groovy script execution

makiam:


Sailsman63:
I think this looks good.

I'm getting very good response times once the script is in the cache, and it doesn't look like the cache itself is taking up much memory.

Next step is to get the startup process to scan and parse any groovy scripts in the keystrokes records.

PR #222, opened 2020-10-04
Clean Color module

makiam:


Sailsman63:
API breakage. You've removed a public constructor. 

Changes like that should not be sounds at random.

PR #223, opened 2020-10-04
No need to calculate ports count for Output module.

makiam:
No need to calculate ports count as Output module always has only single input port.Simple evaluate width anf height from name string


Sailsman63:
I don't see how this saves us anything. The runtime cost is literally too small to measure in any meaningful way, and you've had to duplicate the part of the parent code that really matters here.

PR #224, opened 2020-10-04
Convert multiple if's to switch. Added toString method for debug purpose

makiam:


ArtOfIllusion/src/artofillusion/procedural/IOPort.java
```diff
@@ -134,14 +137,35 @@ public void draw(Graphics g)
...
+    public String toString() {
+        StringBuilder sb = new StringBuilder();
+        sb.append("IOPort{valueType=").append(valueType == COLOR ? "Color" : "Number");
+        sb.append(", type=").append(type == INPUT ? "Input" : "Output");
+        sb.append(", description=").append(Arrays.toString(description));
+        sb.append('}');
+        return sb.toString();
+    }
+  
+
```
Sailsman63, line 169: Does this have any use outside of debugging? If not, it belongs in the TestClass that uses it.

Sailsman63 (changes requested):

PR #227, opened 2020-10-04
hide antialiasing scrollbars

peteihis:
Just a little something to hide the scroll bars that appeared to scroll just one item!


Sailsman63:
What screen resolution are you working with? I'm wondering how this will impact users with smaller screens...

(Though, honestly, I doubt anyone is running with a screen under 720p these days)

peteihis:
I'm on 1930 × 1080 or more but these are in the middle of dialog frame....

Sailsman63:
Playing with down-sized windows, it looks like we should be okay unless someone uses larger-than-standard UI fonts on a smallish screen....

PR #229, opened 2020-10-09
Soft rendering to truely comply with antialias selection

### LINKED ISSUE #198: Raytracer UI glitches
There are some unfinished features in the interacion between the RaytraceRenderer and the UI.

For example the Soft Shadows check box is grayed out when anti aliasing is off, but the function is still active bascally just making terrible mess without AA on.

Soft shadows UI Soft shadows no AA

Which way should it be? Should switching AA off disable the grayed out functions too or should the disabling be removed from the UI? -- I'd vote for the first mentioned.

Also the Illumination tab seems to have similar issues. At least the widgets on the Tab may or may not be grayed out when no illumination is selected ... And I have a vague feeling that some of those things too might stay active when the UI is grayed.

peteihis:
As it says. Resolves #198.

I was a little bit worried about putting an `if()` inside a `for` inside another `for`. With some admittedly light testing the effect seems to be some 5-10% more time spent in the soft shadows method. _(Stdevs of the results are much higher than that, but the trend seems consistent.)_

Taking the `if` outside the `for-for` would lead to some restructuring of that method. Whether it'd be worth it, I'd be happy to hear what anybody else thinks? Not that big a job really, but AFAICT there might be a couple of different ways, how to do it.

### DIFF
--- a/Renderers/src/artofillusion/raytracer/RaytracerRenderer.java
+++ b/Renderers/src/artofillusion/raytracer/RaytracerRenderer.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 1999-2014 by Peter Eastman
+   Modifications copyright © 2020 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


Sailsman63:
>seems to be some 5-10% more time spent in the soft shadows method

Exactly what is 5-10% more? Just the time spent in the Get Direct Light loop, compared to the current code? If so, I think we can live with that. It takes stupidly high shadow ray settings & low quality everything else for the soft shadows to really have a huge impact on overall rendering speed.

If you mean that the whole rendering timeframe is increased by 5-10%, I have not been able to duplicate that, but we'd need to look a little closer.

peteihis:
I tested the rendering time it with gloss off and maximum amount of shadow rays, so the time is mosty spent in getDirectLight. For a more precise test I should have rigged it with a nanosecond thing but I thoght this was good enough to start with. The rendering times varied between 11 and 18 seconds, with 8 repetitions, so  the difference would still disappear into error margins.

peastman:
The cost of one extra "if" should be negligible compared to the other things happening in the loop.

peteihis:
If you could not repeat or are not worried anout it, then I'm happy with it.... (might go back to study it some time though but we'll see...)

Sailsman63:
I suspect that your timing variance from run to run is at least partly due to 'windoze' being weird about thread scheduling, and more aggressive about CPU throttleing- I remember that when I was primarily running windows, my Linux VM was more stable, and sometimes a bit faster when running java code!

I think this one is good to go! Thanks.

PR #232, opened 2020-10-23
Save RenderSetupDialog settings to metadata 

### LINKED ISSUE #230: Saving Render Dialog settings
Just wanted to let you know that I'm working on that. Should not take too long.

- Currently the settings of the selected renderer are saved but settings that belong to the dialog, like the image size, are not. 
- Also , I found that when you open a file and use 'Render Immediately' before doing anything else, on the first time the renderer settings are not read from metadata. The second time they are.

peteihis:
Resolves #230.

Now it checks if settings have changed and if so call `setModified()`. An addition to the`Renderer` interface was needed get the information the renderers but it seems that renderers that have been compiled before the change would still work if they are otherwise compatible.

I came up with a kind of a hidden feature for cases where the user _(well myself at least)_ might like to save already changed settings on Cancel. Let me know what you think.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/RenderSetupDialog.java
+++ b/ArtOfIllusion/src/artofillusion/RenderSetupDialog.java
@@ -31,6 +31,8 @@ public class RenderSetupDialog
   private static final List<Renderer> renderers = PluginRegistry.getPlugins(Renderer.class);
   private final List<ObjectInfo> cameras;
   private final Scene theScene;
+  private Map<String, Object> rendererConfiguration;
+  private Renderer lastConfiguredRenderer;
   private BComboBox rendChoice, camChoice;
   private BRadioButton movieBox;
   private RadioButtonGroup movieGroup;
@@ -116,6 +118,8 @@ public class RenderSetupDialog
     loadRenderSettings(theScene);
     content.add(currentRenderer.getConfigPanel(), BorderContainer.CENTER, new LayoutInfo(LayoutInfo.CENTER, LayoutInfo.BOTH));
     enableMovieComponents();
+    lastConfiguredRenderer = currentRenderer;
+    rendererConfiguration  = currentRenderer.getConfiguration();
     PanelDialog dlg = new PanelDialog(parent, Translate.text("renderTitle"), content);
     if (dlg.clickedOk())
       doRender();
@@ -123,10 +127,12 @@ public class RenderSetupDialog
 
   private void doRender()
   {
+    checkModified();
     recordConfiguration();
     theScene.setMetadata("RenderSetupDialog settings", getConfiguration());
     if (currentRenderer.recordConfiguration())
     {
+      checkRendererModified();
       theScene.setMetadata(currentRenderer.getClass().getName()+" settings", currentRenderer.getConfiguration());
       ObjectInfo cameraInfo = cameras.get(currentCamera);
       Camera cam = new Camera();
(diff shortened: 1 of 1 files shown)


Sailsman63:
>seems that renderers that have been compiled before the change would still work if they are otherwise compatible.

No, they actually don't. They'll *load* just fine, but the moment that the dialog tries to check if any settings have been modified, things crash.

peteihis:
I seem to get varying results. I may have an idea, but I'll be back at it on the weekend....

I think a8eef0a should be OK as such though.

Sailsman63:
It should be.

A thought: I don't think that it would be all that difficult to include a `private class foo extends PanelDialog` that added a third button. You'd have:

- Okay (Save settings to file and render)
- Cancel (Dismiss dialog without saving)
- Save Settings for later (Save settings to scene & Probably? dismiss dialog)

This would save a whole bunch of complexity trying to figure out if settings have been changed, etc. and probably prevents any need to extend the Renderer interface.

peteihis:
> This would save a whole bunch of complexity trying to figure out if settings have been changed

Not really. I tried a _trick or two_....

Checking if any of the values are changing (`boolean`) is not much different from checking how many of them are changing (`ìnt`).

And because the `Renderer` interface does not provide any channel to check if values have changed and the renderers are only known as 'Renderer' it is not possible to get anything from the renderer to the core SW around the interface. -- And naturally it is not possible to `import RaytracerRenderer` or the other one into the dialog as they are plugins. ( `( RaytracerRenderer)currentRenderer.modified...` would seem simple but `can not find symbol`).

The two ways to do it are either to send the parent (with the scene included) to the renderer or to ask the renderer about values that are going to change. Somehow the latter seems lighter to me but the as this is not nearly the only place where `modified` is not updated correctly the uniform approach could just be to sent the main `LayoutWindow` everywhere.

OTOH I'm not aware of any plugin renderers that would work with the current version any way....  But I'll test a _third or so_ trick and get back.

The buttons then, if that's the way you'd like to go, to me "Save settings" and "Reset renderer" would make a nice pair to work with. A "Save and exit" would seem a kind of blunt/unnecessary as there already are four different ways to exit the dialog. -- And the reset would make life easier to someone who is new to it... It's pretty easy the get confused if you don't know what everything is about (as I vaguely recall ;) ).

Sailsman63:
>Checking if any of the values are changing (boolean) is not much different from checking how many of them are changing (int).

My point was to avoid checking this at all.

- Render *Always* writes metadata
- Cancel *Never* writes metadata
- Save Settings *Always* writes metadata (and could very well leave the dialog open)
- No "Reset" button. That's what Undo in the main window is for.

peteihis:
Oh, I see.... _(I may not be at my brightest at this our any more...)_ The point with the reset button would be that if you have tinkered with the settings before and already saved them in a file that you just opened, you would have a way to start over.... But coming to think of it, that is probably not very useful to most of us after all.

peteihis:
> (I may not be at my brightest at this our any more...)

As said -- after a badly slept week.... so:

The primary reason to check if values on the renderer panel have changed is to decide if `LayoutWindow.setModified()` should be called. That can not be avoided if we want the state to be correct. -- And I'm not particularly happy about thoise cases where AoI is either ignoring changes or marking the scene modeified just in case....

I came up with a way to handle the check inside `RenderSettingsDialog`. For the renderer configuration it means that the dialog has to `getConfiguartion()` from the renderer and store the map until "OK". (This is not my _third-or-so-option_ , it was yet different.)

I'll push an update in the evening after some more testing.

Sailsman63:
Okay, the current state looks like a good start to me.

We can consider whether we want to add a "Save Settings" later. Anything you'd like to address before I merge this?

peteihis:
Go ahead.

I though about the Save settings thing but the idea started to crawl towards "Manage settings". I think I have mentioned the idea before and I have a rough idea what I'd what to see there but not much more so far.

PR #237, opened 2020-11-07
Clean empty interface implementations

makiam:


peteihis:
What would changing the `RenderListener` methods from `public` to `default` do to plugin renderers? I don't see the `RaytracerRenderer` or `Raster` on the list either.

makiam:
This should not affect them

Sailsman63:
Good change. Sorry I missed this one. (Not sure how... It's very straightforward.)

PR #238, opened 2020-11-07
Clean Tracks defaults

makiam:


ArtOfIllusion/src/artofillusion/animation/Track.java
```diff
@@ -183,16 +184,16 @@ public Keyframe setKeyframeIfModified(double time, Scene sc)
...
-    return new Track [0];
+    return new Track[0];
   }
   
   /** Determine whether this track can be added as a child of an object. */
   
   public boolean canAcceptAsParent(Object obj)
   {
-    return false;
+    return obj instanceof ObjectInfo;
```
Sailsman63, line 196: This changes the default behavior of any `Track` that does not override this method.

Why the change? "Most Common Override" does not always translate to "Best Default"

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/animation/Track.java
makiam: We have 18 Track class implementations and only 2 of them returns false by default. So my update definitely looks for me as Best Default

PR #240, opened 2020-11-08
Save menu item enabled state

peteihis:
The Save-menuItem was staying enabled after save, if you didn't at least click a view after saving (by "Save"). Most actions on the UI `updateMenus()` but his did not.

There seems to be something not quite right with the Move and Rotate tools too that the `menuItem` does always get updated. If this PR is still here when I come up with a fix to those I'll add it here. If not, I'll PR that separately.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -1,6 +1,6 @@
 /* Copyright (C) 1999-2015 by Peter Eastman
    Changes copyright (C) 2016-2020 by Maksim Khramov
-   Changes copyright (C) 2017-2019 by Petri Ihalainen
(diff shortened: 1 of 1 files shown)


Sailsman63:
I can't duplicate your issue here. I've:

- Launch AOI
- Add a few objects
- Save
- After saving, no further clicks in UI:
    - Ctrl + S Shortcut does nothing
    - Save option in file menu is grayed out


Am I missing something?

peteihis:
To me it appears like this.

[SaveMenuItem.zip](https://github.com/ArtOfIllusion/ArtOfIllusion/files/5508511/SaveMenuItem.zip)

When Save is used either by the menu item or Ctrl+S the `modified` status on `LayoutWindow` is updated but not the menu item `setEnabled()`.

Sailsman63:
Got It! Thanks!

PR #244, opened 2020-11-12
Add Traditional Chinese translation file

microwave54:
Add Traditional Chinese translation file and edits to History.md and Translate.java files


Sailsman63:
Thank You very much!

PR #245, opened 2020-11-13
Alternate Syntax Highlighter implementation

Sailsman63:
Turns out that it's possible to replace the `ScriptEditor` component with some bits that come standard with RSyntaxTextArea

This version allows the various script editors to remain independent for the time being, rather than forcing them into an inheritance hierarchy.

Two questions:

- Are we safe in removing `ScriptEditor` completely, or is there a chance that plugins are using it directly?
- The backing library is a fairly sizable dependency compared to AOI as a whole. (1.14MB) Is the syntax highlighting worth the extra load?

See also #203


makiam:
I see no significant difference against my PR... Just less changes. Once I create my PR I move doubling code for file loading and UI creation to common class. Here this areas unchanged. 

No better composition code against my inheritance code

Sailsman63:
>Just less changes

Which was sort of the point. 

This may have been a miscommunication, but when I was reviewing your PR, it came across to me as "I needed to make these refactors to get this to work at all" rather than the more accurate "The provided components slip right in where the original custom solution fit, and I also thought it might be good to refactor the structure here" With said refactor being kept as a separate commit so that it can be evaluated on it's own merits.

PR #246, opened 2020-11-16
Handling of editing by ScriptedObject

### LINKED ISSUE #241: ConcurrentModificationException prevents opening the scene
Hi.

I made a tiny mistake with a scripted object. The object needs to read information from the scene. The I tried to return something into the scene to be able to look at the result (which I can not do inside the script). Ste object stopped working and after save the file can not be opened any more.

One time the log system alerted `java.util.ConcurrentModificationException`. After that never nothing. Depending on how I tried to open it nothing happened or AoI got into infinity loop. The error was then found in System output that I directed to files but no way to open the file.

I've got a safe copy of the scripted object, so no worries, now but I'm probably not the only one who will ever try something like that.

peteihis:
Resolves #241.

It notifies the user when the information is most likely to be understood and otherwise (it is supposed to) just tolerate the case.

There are still  a lot of calls to `Scene.setTime()` in the code that don't check if the scene content changed or update the ItemList accordingly. Though, when those calls happen, the user should already be aware of what is happening.

EDIT: I did not translate the notifications either. If you think, this is approach is OK, I may check the rest of the code and translate if you feel it's worth it.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -1249,6 +1249,8 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
   public void setTime(double time)
   {
     theScene.setTime(time);
+    if (theScene.itemListChanged)
+      rebuildItemList(); // The user sees the change. The problem has been pointed out before, so let's just show the effect.
     theScore.setTime(time);
     theScore.repaint();
     itemTree.repaint();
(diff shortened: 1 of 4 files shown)


peteihis:
The `AnimationPreviewer` seemed to be the only one requiring attention. The rest refer to their private scenes.

makiam:
Probably much simple solution to avoid ConcurentModificationException... Just to replace Objects vector with CopyOnWriteArrayList.
At least for my quick check I see that scene is successfully loaded and Score animatin is work again

But much deep testing is a good idea

peteihis:
> Just to replace Objects vector 

That's basically what I did. Copied the vector to a temporary vector. That took care of the `Concurrent`thing. Then I just compared the temporary Vector to the current situation to send a note to the user if necessary.

PR #247, opened 2020-11-16
StartTime of rendering Movie

peteihis:
A little _oops_ in the new bit. The rest seemed OK.


Sailsman63:
Thanks for the catch!

PR #248, opened 2020-11-27
Rendering primitives, Light ray fix

peteihis:
The light ray issue fix for actual primitives, `Sphere`, `Ellipsoid`, `Cube` and `Cylinder`.

I think the solution is now simpler than before. No fields needes to be and only one variable in one method.

I tested `Math.max/min` vs. `if( < > )` with a plugin. It turned out that the `max` and `min` functions in real Java (as in comparison to Groovy) are faster than `if`. It is hard to say exactly how much faster as results in Windows vary quite a bit, but significantly. For comparison I added a set of multiplying and dividing to the test set. The `max` and `min` seem to be speed wise just as good a those. 

I may have a look at `RTTriangle` too and add it to this if I can get the implementation simplified and also see what, if anything, to do with `RTImplicitObject`.

### DIFF
--- a/Renderers/src/artofillusion/raytracer/RTCube.java
+++ b/Renderers/src/artofillusion/raytracer/RTCube.java
@@ -1,4 +1,5 @@
 /* Copyright (C) 2004-2013 by Peter Eastman
+   Editions copyright (C) by Petri Ihalainen 2020
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -27,6 +28,8 @@ public class RTCube extends RTObject
   protected Mat4 toLocal, fromLocal;
 
   public static final double TOL = 1e-12;
+
+  private double cubeTol;
 
   public RTCube(Cube cube, Mat4 fromLocal, Mat4 toLocal, double param[])
   {
@@ -103,6 +106,8 @@ public class RTCube extends RTObject
     }
     bumpMapped = cube.getTexture().hasComponent(Texture.BUMP_COMPONENT);
     this.toLocal = toLocal;
+    cubeTol = (Math.max(Math.max(Math.abs(fromLocal.m14), (Math.abs(fromLocal.m24))),(Math.abs(fromLocal.m34))) + 
+               Math.max(Math.max(xsize, ysize), zsize))*TOL;
   }
 
   /** Get the TextureMapping for this object. */
@@ -128,6 +133,7 @@ public class RTCube extends RTObject
   {
     Vec3 rorig = r.getOrigin(), rdir = r.getDirection();
     Vec3 origin, direction;
+
     if (transform)
     {
       origin = r.tempVec1;
@@ -167,7 +173,7 @@ public class RTCube extends RTObject
         if (t1 < maxt)
           maxt = t1;
       }
-      if (mint > maxt || maxt < TOL)
+      if (mint > maxt || maxt < cubeTol)
         return SurfaceIntersection.NO_INTERSECTION;
     }
     if (direction.y == 0.0)
@@ -193,7 +199,7 @@ public class RTCube extends RTObject
         if (t1 < maxt)
           maxt = t1;
       }
-      if (mint > maxt || maxt < TOL)
+      if (mint > maxt || maxt < cubeTol)
         return SurfaceIntersection.NO_INTERSECTION;
     }
     if (direction.z == 0.0)
@@ -219,12 +225,12 @@ public class RTCube extends RTObject
         if (t1 < maxt)
           maxt = t1;
       }
-      if (mint > maxt || maxt < TOL)
+      if (mint > maxt || maxt < cubeTol)
         return SurfaceIntersection.NO_INTERSECTION;
     }
     int numIntersections;
     Vec3 v1 = r.tempVec1, v2 = r.tempVec2, trueNorm = r.tempVec3;
-    if (mint < TOL)
+    if (mint < cubeTol)
     {
       v1.set(rorig.x+maxt*rdir.x, rorig.y+maxt*rdir.y, rorig.z+maxt*rdir.z);
       mint = maxt;
--- a/Renderers/src/artofillusion/raytracer/RTCylinder.java
+++ b/Renderers/src/artofillusion/raytracer/RTCylinder.java
@@ -45,53 +45,53 @@ public class RTCylinder extends RTObject
     cone = false;
     transform = true;
     if (vy.y == 1.0)
-      {
-        if (vx.x == 1.0 || vx.x == -1.0)
-          {
-            rx = size.x/2.0;
-            rz = size.z/2.0;
-            transform = false;
-          }
-        else if (vx.z == 1.0 || vx.z == -1.0)
-          {
-            rx = size.z/2.0;
-            rz = size.x/2.0;
-            transform = false;
-          }
-        if (transform == false && ratio == 0.0)
-          cone = true;
-      }
-    else if (vy.y == -1.0 && ratio != 0.0)
-      {
-        if (vx.x == 1.0 || vx.x == -1.0)
-          {
-            rx = size.x/2.0;
-            rz = size.z/2.0;
-            transform = false;
-          }
-        else if (vx.z == 1.0 || vx.z == -1.0)
-          {
-            rx = size.z/2.0;
-            rz = size.x/2.0;
-            transform = false;
-          }
-        if (transform == false)
-          {
-            rx *= ratio;
-            rz *= ratio;
-            ratio = 1.0/ratio;
-          }
-      }
-    height = size.y;
-    halfh = height/2.0;
-    if (transform)
+    {
+      if (vx.x == 1.0 || vx.x == -1.0)
       {
         rx = size.x/2.0;
         rz = size.z/2.0;
-        if (ratio == 0.0)
-          cone = true;
-        this.fromLocal = fromLocal;
+        transform = false;
       }
+      else if (vx.z == 1.0 || vx.z == -1.0)
+      {
+        rx = size.z/2.0;
+        rz = size.x/2.0;
+        transform = false;
+      }
+      if (transform == false && ratio == 0.0)
+        cone = true;
+    }
+    else if (vy.y == -1.0 && ratio != 0.0)
+    {
+      if (vx.x == 1.0 || vx.x == -1.0)
+      {
+        rx = size.x/2.0;
+        rz = size.z/2.0;
+        transform = false;
+      }
+      else if (vx.z == 1.0 || vx.z == -1.0)
+      {
+        rx = size.z/2.0;
+        rz = size.x/2.0;
+        transform = false;
+      }
+      if (transform == false)
+      {
+        rx *= ratio;
+        rz *= ratio;
+        ratio = 1.0/ratio;
+      }
+    }
+    height = size.y;
+    halfh = height/2.0;
+    if (transform)
+    {
+      rx = size.x/2.0;
+      rz = size.z/2.0;
+      if (ratio == 0.0)
+        cone = true;
+      this.fromLocal = fromLocal;
+    }
     cx = fromLocal.m14/fromLocal.m44;
     cy = fromLocal.m24/fromLocal.m44;
     cz = fromLocal.m34/fromLocal.m44;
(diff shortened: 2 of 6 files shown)


peteihis:
I'm looking at a couple of further simplifications.

Renderers/src/artofillusion/raytracer/RTCube.java
```diff
@@ -128,6 +129,8 @@ public SurfaceIntersection checkIntersection(Ray r)
   {
     Vec3 rorig = r.getOrigin(), rdir = r.getDirection();
     Vec3 origin, direction;
+    double cubeTol = Math.max(Math.max(Math.max(maxx-minx, maxy-miny), maxz-minz), rorig.length()*0.01)*TOL;
```
Sailsman63, line 132: I thought, when we first discussed this, that this could be an instance variable that was calculated when the `RTCube` was constructed.

When did it acquire the dependency on the length of the Ray origin?

peteihis: > I thought, ... that this could be an instance variable

So did I at the time, but it has been like this for a while already. 

The original idea works as long as you stay relatively close to the world origin but when the objects travel further away the light ray problem returns. It happens independently in all coordinate directions. Some of the test scenes I posted demonstrate just the phenomenon, that both the size and the distance from origin have a part here independent of each others.

But when I now started to test a bit more, it turned out that at least in the case of `RTCube` I could drop the object size from the equation and use only the ray origin. For the more complex objects it seems a bit more complicated....

What is behind it is that numerical error with double precision numbers is usually around 1E-16 but that only applies to numbers close to 1.0. The error naturally increases while the number increases so for 1E5 the error is around 1E-11 and the ray origin is defined in world coordinates.

**So back to square one.** I think I'll  reset this branch and rethink.

For one thing, I don't like myself that the object surface tolerance is recalculated for every ray. I'm not too crazy about calculating the square root in the `.length()` either but that worked. On the other hand, if it was going to be done this way there was no reason to keep the object variable in any more. _(At some point I even had the instance variable and a couple of temporary variables before the final decision about the tolerance, but I compressed those out of the code.)_

Looking at it now, the idea that might work could be to find the point of the object (well, it's bounding box actually) that has the greatest distance from the world origin and select the max absolute value of the coordinates as the base of the tolerance. This tolerance would be used as the instance variable.

Then there is the problem with round surfaces of spheres and cylinders. The tolerance is compared to values calculated from `r2` so the numerical error increases (somehow) relative to the 2nd power of the size of the object while it increases in the 1st power of distance. I did some initial testing with `Cylinder`. What I got so far was not yet an universal solution but it seemed like a possible track.

Anyway the 1E-12 tolerance is extremely ambitious to begin with. In real world dimensions it would mean one nanometer for one kilometer (!). I doubt if anybody would ever render an image that requires that good a separation (well, I might, but...). We probably could afford to loosen it a couple of decimal positions if it comes to that.

peteihis:
That should do it. Tested with size factors 1E-8 to 1E+8.

Attached a test file. Nothing moving, just render with default setting without and with this PR.
[Raytest1.zip](https://github.com/ArtOfIllusion/ArtOfIllusion/files/5646709/Raytest1.zip)

Sailsman63:
This looks good to me so far.

@peastman, You're a lot more familiar with the raytracer than I am. See any specific points of concern? We're talking about adding one or two doubles to each of the RTObject primitives, initialized on object creation.

peastman:
Adding the extra fields will increase memory use and hurt performance.  I'm mainly worried about triangles, since there tend to be a lot more of them than primitives.

As a compromise, how about if we remove the field from RTTriangleLowMemory and compute the tolerance in `checkIntersection()`?  The whole point of that class is to use as little memory as possible.

I think the field could easily be a float instead of a double.  That would reduce the memory for some of these classes (though not RTTriangle, I think).

Sailsman63:
You've got a point about `RTTriangleLowMemory`

As for overall performance concerns: I've been running some tests with some fairly detailed triangle meshes. (total of around 40k triangles in the definition meshes, rendered with very high surface accuracy.) So far, I've not been able to notice any difference in the render times. (~5 minute render coming out within a second or so either way, well within variance that I would expect from outside factors.)

peteihis:
OK. I'll try those things: 
- LowMemory triangle to recalculate the value for every ray
- Object tolerance field to `float` ... And I'll do some performance testing. 

So far it has been looking like the constant recalculating would hurt a lot more than adding a field, but I'll try to put some numbers on a few things that happen in there.

Sailsman63:
The performance hit would be compared to current HEAD, with the hard-coded `TOL` and rendering glitches, not the recalculate-on-the-fly approach.

Go ahead and try the floats, but I'm not expecting much of a difference. I suspect that the Raytracer's heap access patterns are already cache-unfriendly enough that it'll just be lost in the noise...

peteihis:
> The performance hit would be compared to current HEAD, with the hard-coded TOL and rendering glitches, not the recalculate-on-the-fly approach.

Yes, the HEAD would make the baseline but while at it, I'd run a few other things too. 

From earlier life I have the experience that `float-float` calculations are just marginally faster than `double-double`. I would not expect `double-float` to be much different than `double-double`. Memory wise, I would expect the difference with or without the the field to be rather marginal as well but the idea I had for testing would be to eliminate any room for guessing. :)

peteihis:
@peastman: What would be your expectation, what would happen with these changes? How hurt is hurt and where does it show? And what exactly (or just generally) is the mechanism that should cause performance drop (otherwise than that there just is more stuff in the memory)?

I have been slowly working on a set-up that simply checks how much time was spent at each of the major steps of rendering and what the memory usage is between the steps. Some "pilot" runs would seem to suggest, that the price on the quality guarantee we'd obtain would be very reasonable.

peastman:
Just more stuff in memory.  Because Java leads to such inefficient memory access, rendering tends to be memory bound rather than compute bound.  If you increase the size of an object by 8 bytes, that's 8 more bytes it's taking up in the cache and another 8 bytes that need to be loaded every time we need one that isn't in the cache.

Sailsman63:
>Java leads to such inefficient memory access

Yeah, cache-aware programming is tricky in pretty much any context. Java just tends to hide it a little harder. I've run across references to performance-critical java code resorting to operations on buffers/primitive arrays (basically, faking C-style `struct`s)

peteihis:
> 8 more bytes it's taking up in the cache and another 8 bytes that need to be loaded every time we need one that isn't in the cache.

Any hints how I would build a scene that emphasizes this problem? So far I have tested with something with a lot of reflected rays and another one with heavy ambient occlusion (actually just AO and nothing else) but basically the results are inconclusive.

What always happens is that when a raytracing scene is built for the first time it takes about 1.5-2 times as long as for any subsequent images in the same session. Another obvious thing is that building the scene takes a tiny bit longer, when it calculates the object tolerances. However that added time is soon lost in the variation of total rendering times. 

Not much to go about but there are cases when with a precalculated `float` the total time was shorter OR the memory was filled less than without the variable. But only two test cases ran so far and variation is _wayyyy_ greater than the difference in performance.

...And I have been refreshing my memory about error margins, confidence levels and t-tests.. _(the Welch's t-test would be my tool of choise....)_ ... That stuff just doesn't stick to braincells for very many decades... :D

peastman:
> So far I have tested with something with a lot of reflected rays and another one with heavy ambient occlusion (actually just AO and nothing else) but basically the results are inconclusive.

The number of rays doesn't matter.  What's important is the number of triangles.  Render a complicated scene with a fine surface error tolerance.

Sailsman63:
Just did some (very preliminary) testing here. I actually found that I have access to some kernel-level tools that can track cache stats for a process.

I got a seriously detailed mesh rendering at high resolution, and so far... I've not found any significant difference in cache performance between current HEAD and this branch. Both ran at about 8% cache miss for the rendering settings I used on a quiet system. (closed out of other applications, etc. to reduce noise)

- There are some wide-ish error bars on that number, due to me having to manually start the stat in a separate terminal window at the same time as I started the render.
- I am certain that cache-locality was *a* bottleneck for this scene and settings. I was able to increase and decrease both the overall render time, and the cache-miss ratio by increasing and decreasing the allowed surface error.
- This does include the initial "Processing Scene" phase, which has (unsurprisingly) a much higher miss ratio. I've not yet figured out a way to do more fine-grained (timewise) logging in a useful fashion.

peteihis:
@peastman: My idea about those case was not the raycount but the fact that in thoses cases the (seconday) rays would travel pretty randomly to different parts of the scene. I would expect that to cause need for some in-and-out cache traffic. Now I also had more "triangle-intensive" scene but I need more runs ....

Over the holidays I wrote little hacks to `ArtOfIllusion` and `RaytracerRenderer`, that print to `System.out` some numbers about time consumtion ona memory usage. -- Obviously not about using cahce though. 

Anyway the testcase I eneded up with was basically: Call System.gc() before rendering, Render a movie of 12 images (1 sec / 11 fps), ignore what was reported for the 2 first frames and use the the rest to analyse. 

So far the preliminary result are very sronly suggesting that _**"This myth is busted!"**_. With the triangle-intesive case I got even lower T-values than with the ray-intensive ones (The higher the T then the more significant the result). The problem with T-tests of course is that you need to choose what you wish to believe, but I'd say that a T between two different cases should be at least 2×T between exactly same cases (preferably a max from a group of few runs) to mean any significance.

I'll do some more and put them through the machine before I report any results (and I'll put the test cases somewhere that you can get them if you wish to have a closer look, but not until I'm done.)

Sailsman63:
> the preliminary result are very srongly suggesting that "This myth is busted!".

I think that it would be more fair to say that the Raytracer's cpu cache performance is already so bad that this change will not make a discernible difference in most situations.

If my testing is anywhere near the ballpark, we're looking at 1 in 10-12 requests for data missing the cache entirely, and having to go all the way to main memory. This does *not* include intra-cache thrashing, (Cycling data in and out of L1 to L2 or L3) nor does it catch inter-thread contentions. (multiple threads trying to access the same cache line simultaneously)

[This stackoverflow post](https://stackoverflow.com/a/29188516/1922925) gives a bit of rough insight into the relative costs of cache misses.

> Obviously not about using cahce though.

Which is actually impossible inside the JVM, anyway. You can, of course, call out to binary code/tools, but the testing I was doing was with linux command line tools running in a terminal window...


---------------------

I'm about 90% convinced that this change would be fine as is, except for the LowMemoryRTT. (That one is specifically designed to fit in the smallest memory footprint possible, for running in memory-constrained situations.)

peteihis:
I finished one set of tests:

Those were on a 8GB-RAM, 2MB cache laptop, allocated 5GB for Java. Ran each case twice: HEAD as is and `double` and `float` for the precalculated variable, 10 last of 12 rendered frames included in the analysis.

In performance loss we are probably talking about 0,5 - 1 % as increased processing time. The earlier, not so well planned tests seem to agree with that.  Margin of error at 90% confidence level per each run varies between 1,04% and 2,12%. T-tests, when valid, are insisting that there is absolutely do difference, what so ever, between the cases.

Comparing just the scene building times the outcome is very similar with the difference that variation and MOEs are a bit higher. Any expected increase in time consumption there was not detected either. &rarr; Cache limit seems like a credible suspect there too.

Reading the post above I began to wonder if the result would be any different on a processor with larger cache? The practice (based on current catalogs) with processor manufacturers seems to be that they reserve 1MB / thread. It might be interesting to run just 4 treads on 12 MB cache, but I'm not planning to take the time to do that now.

> except for the LowMemoryRTT

I wonder if that'd be worth changing? The LowMemory triangle is anyway a bit of a remnant of the past but more importantly, I would not expect much difference in how memory is used. Instead redoing the 21-operation calculation every time you need it would just increase processing time unnecessarily.The other values that are _not precalculated_ for LMRTT require significantly fewer calculations. The field could of course be a `float`. That should help at least in theory.

peteihis:
I ran a set of tests with `RTTriangleLowMemory`. 

Without much analyzing, all that seems to happen is that time is consumed about 10% more than with the regular RTTriangle. MOE90 values vary rougly between 1 and 2 %.

The `totalMemory-freeMemory` after building the RTscene is in average only about 3% lower but the MOE90s are varying between 3,5 and 15 %, so nothing can be said with any certainty.

This, of course, was measured with all the existing memory set available for Java. I don't know what would happen if the memory allocation was limited but the difficulty would probably be to find a number where the RTTLM actually begins to affect without the process exceeding the limit. My best guess is that if the model is so large that the memory allocation is exceeded, then it is rather unlikely that using the Low Memory setting would help.

What do you think?

Sailsman63:
>I began to wonder if the result would be any different on a processor with larger cache?

The primary difference would be the size of scene before cache misses start to have a significant effect. It would definitely affect the *rate* of cache misses for any given pair {scene, renderingSettings}

>The practice (based on current catalogs) with processor manufacturers seems to be that they reserve 1MB / thread

It gets more complicated than that. My (Older) current processor has a "Smart" cache that allows some of the cores to take over most of the Cache if workload warrants...

>I wonder if that'd be worth changing? The LowMemory triangle is anyway a bit of a remnant of the past but more importantly, I would not expect much difference in how memory is used

Maybe. It's for memory-constrained situations. In theory, it might make the difference between fitting a rendering session in physical ram, and forcing the OS to page to disk. (Virtual Memory) In such a situation, the Low-memory version is probably *faster*, despite constantly re-calculating certain data.

>The totalMemory-freeMemory after building the RTscene is in average only about 3% lower but the MOE90s are varying between 3,5 and 15 %, so nothing can be said with any certainty.

Do you know how many renderingTriangles are involved at this point? Also, are there any *big* textures? (large image-maps is mainly what I'm looking for) We really need to be comparing memory use for the triangles only.

peteihis:
There are textures but not any really heavy ones, but the model is "triangle-heavy" itself. I contains scanned model, as I recall some 40 000 triangles in it alone. The rest I'd have to check. The earlier cases would be lighter as models as they are using live duplicates.

peteihis:
Okay, so.... 573 508  RenderingTriangles in total, 448 880 of them from Lucy alone. Not the best possible model that way.

lcy

peteihis:
And a quick check: I made a copy in all white and removed all the extra textures from the file. Faster to render but not much difference in size. 

BUT 

>  between 3,5 and 15 %, 

I looked at the wrong line. That's what happens with `freeMemory`. The total-free is more like 0.3-1.5%. So there might be a difference after all.

peteihis:
> memory use for the RTTtriangles only

Closest to that I could get by the using the difference from used memory before `buildScene()` and right after it. For that number there were a copule of outliers, that I removed from the analysis. In those cases garbage collection had not happened (fully) before `bulidScene()`.

What comes out is that by using the LowMemory setting we are exchaning about 6-8% of less memory occupied by the object into 10% more of processing time. I honestly doubt if anyone can find a case, when that difference really matters.

When it comes to having the object tolerance present or not, actually seem to matter to `RTTriangleLowMemory` but still very marginally. The T-test _a-kind-of_ says that the results might be true. IMO they are is still on the edge. The memory space that the RT-objects use seems to be increaring by some 2,5% but the strange part is that in the case that doesn _not_ have the variable precalculated appears very slighty worse than the one with the precalculated field. Though that difference can not be considered real. I only did this comparison with a `float` type field. 

---

A bit outside the topig, the T-test is a bit of a strange animal: 

- It is said to test if two populations are equal, but all it really can tell about is the _sample sets_ that it is calculated from.
- It still does not say "yes" or "no", it gives a number.
- I don't think I have ever seen any clear instruction on how to select the threshold value to it. All I was ever told was to look at the entirety of the case... Well.... 

I this case, when I put together the results of two runs of the regular RTTs the Ts got lower, which definitely suggest, that the sample sets are closer to being equal than not. Instead if I'd have 20000 repetitios T's would start to get very high no matter how close the averages would be. Hence defining the threshold-T by comparing two supposedly similar sets is the only way I can think of to make any sense of it.

---

On the regular triangles case I would have expected that `buildScene()` would take a bit more time and that the expected increase in time consumption would come mostly from that, allowing faster processing. Nether of the assumtions turned out to be true. All I can see is process variation that is there anyway.

---

**Conclusions:**

**To me this test case does not indicate any reason to change anything.** Neither the doubles to floats nor have the RTTLM recalculate the value every time. _(If you look at the performance difference values they actually suggest that it'd be better this way than with either of the suggested changes, but statistically the differnce is entirely meaningless.)_

Now reeling back a bit, I did not get any clear difference into processing times even with the earlier PR where I had unique tolerances calculated for each `OctreeNode` either. Honestly the persistent claim, "It hurts performance", did not sound quite right to me after that. 

It would have been nice to at least have a rough expectation, _how much hurt_. Anyway, now it has been tested and obviously the bottlenecks are somewhere else. If you can come up with case where you can point out a **clear** difference, I'd be happy to know. By clear I mean some 15% or more repeatedly (±10% variation happens all the time).

Of course, in this test set-up there was basically only one variable to test. There are plenty of things that were left outside like cache size and/or amount of threads. Also it just might be that with a scene where rays travel more between objects before hitting (or not) a light the difference between HEAD and the PR would be higher. OTOH the cases I ran before this one did'n not indicate anything dramatic there either.

Somehow it comes to my mind that in retrieving data from RAM seems like an operation that takes a certain time and it does not really matter if there are a few bytes more or less to retrieve -- but that is just something I'm picturing in my head.

And again: If we want so get rid of this problem with rendering larger objects (I know I do!) I'd expect that to come with a cost though so far any noticeable cost has not emerged.

---

Attached a zip-package with the test results and the used scene and the parts I modifier to print out results. The modified RT-objects are not included, I think those are quite obvious if you have this PR available.

[Performance Case.zip](https://github.com/ArtOfIllusion/ArtOfIllusion/files/5759673/Performance.Case.zip)

I also realized that the way that I had `RaytraceRenderer` write the values is not optimal for handling the results. It should just print everything on one line/image, so it'd be easier to handle results longer image sequences on a spread sheet.

Sailsman63:
Wow! Lot of data... Please be patient if it takes me a bit to review... A question here has started me down a path that I want to check on before going further, which may take a few test runs to follow up.

Sailsman63:
And here we are, nearly a year later. I'm not sure how this got away from me. I don't even recall what that last question would have been.

Any last-moment concerns that have cropped up since the last time this was discussed?

peteihis:
I can't rememberany more what exactly this contained but nothing new has happened since then.

Sailsman63:
I've rebased the commits here to put the whitespace changes in a single commit. I'll push that before merging. One last test I want to do: I'm going to run this on a RaspberryPI.

Sailsman63:
I never did get this running properly on my Pi-ZeroW. I'd still like some feedback on running this on memory-constrained devices with the LowMemory RTT, but that is an uncommon platform. We'll just have to see how it works in the real world.

PR #249, opened 2020-12-12
Update syntax hilighting style and animation

makiam:


peteihis:
Thanks. Much nicer, esp. with the block marker highlight.

In general the colors are a bit on the strong side to me, but probably best left as they are for now. The thing with using colors & style & highlights to separate different types of information is that it can make the code much clearer to browse _(and I have been missing that feature)_ but you easily end up "fighting fire with fire". Finding a good balance is definitely an art form of its own.

Sailsman63:
Looks good. I'm not inclined to bikeshed the color scheme at this point either.

I'd never actually noticed the animation on the brackets - it passed so fast, I could only see it if I was specifically watching for it...

PR #253, opened 2021-01-11
Avoid AIOB exception in log from Tool Palette

makiam:
Avoid AIOB exception in log from Tool Palette when click on palette empty cell space


Sailsman63:
Thanks!

PR #254, opened 2021-01-24
Remove Track's unused track method Scene parameter

makiam:


Sailsman63:
This PR would be a breaking change of the API. Are we absolutely certain that a plugin would not be calling these methods directly?

Also, @peastman Was the scene parameter intended for something, or is this a legacy from a former design layout?

peteihis:
True that the scene parameter is not used in any of the cases but at least theoretically it is perfectly possible to create an object that interacts with the scene. This possibility has been there "since the beginning of time", so let's not remove it now.

PR #258, opened 2021-10-25
Horizontal wheel scroll

peteihis:
Java 11 adaptation to handle horizontal scroll and prevent unwanted zooming while rotating the view.

So far tested with a Gigabyte mouse, that had "oversensitive" left/right clicks in the scroll wheel. This does not entirely tame down the unwanted twitches, but seems reasonably good. I tested one "smarter" approach to detect accidental multiple inputs but that one was not working as well as I had hoped.

Edit: Tested with MX Master 2S, with a thumb wheel for horzontal scroll. Works quite nicely. :)

The view frustum drawing seems to have a minor bug. The bug clears itself in use and does not appear again but I'll have a look if I can figure it out. Looks like a  parameter missing a value in the beginning.

Also by a bit of stress testing (pressing multiple mouse buttons in random sequences) it may be possible to drive the parameters to a state where the wheel appears partly or fully disabled. However I could not systematically repeat that.

The logic with the directions is that by default all wheel actions move you in the model space. The horizontal move can be reversed separate of zoom or it can be disabled entirely.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ScrollViewTool.java
+++ b/ArtOfIllusion/src/artofillusion/ScrollViewTool.java
@@ -1,4 +1,4 @@
-/* Copyright (C) 2017-2020 by Petri Ihalainen
+/* Copyright (C) 2017-2022 by Petri Ihalainen
    Changes copyright (C) 2020 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
@@ -31,7 +31,7 @@ public class ScrollViewTool
   private Camera camera;
   private double distToPlane, scale;
   private double scrollRadius, scrollBlend, scrollBlendX, scrollBlendY;
-  private int navigationMode, scrollSteps, startOrientation;
+  private int navigationMode, startOrientation;
   private Vec3 startZ, startUp;
   private Rectangle bounds;
   private Point mousePoint;
@@ -46,10 +46,19 @@ public class ScrollViewTool
 
   protected void mouseScrolled(MouseScrolledEvent e, ViewerCanvas v)
   {
-    scrollSteps = v.scrollBuffer;
-    v.scrollBuffer = 0;
-    v.mouseMoving = false;
+    // Igore wheel while processing other mouse events
+
+    if (v.mouseDown || v.rotating || v.moving || v.tilting)
+      return;
+
+    // If the wheel event is a horizontal click, ignore it
+    // The mask, 64 was not mentioned in Java documentation.
+
+    if ((e.getModifiersEx() & 64) > 0)
+      return;
+
     view = v;
+    view.mouseMoving = false;
     view.scrolling = true;
     scale = view.getScale();
     distToPlane = view.getDistToPlane();
@@ -69,7 +78,7 @@ public class ScrollViewTool
     startUp = new Vec3(coords.getUpDirection());
     view.setRotationCenter(coords.getOrigin().plus(startZ.times(distToPlane)));
     mousePoint = view.mousePoint = e.getPoint();
-    scrollTimer.restart(); // The timer takes case of the graphics and updating the children of a camera object
+    scrollTimer.restart(); // The timer takes care of the graphics and updating the children of a camera object
 
     // This could be cleaser if the scrollMoveModel and scrollMoveTravel weres 
     // divided to methods, dedicated to one action only
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -1,5 +1,5 @@
 /* Copyright (C) 1999-2011 by Peter Eastman
-   Changes Copyrignt (C) 2016-2020 Petri Ihalainen
+   Changes Copyrignt (C) 2016-2022 Petri Ihalainen
    Changes copyright (C) 2016-2022 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
@@ -39,8 +39,8 @@ public abstract class ViewerCanvas extends CustomWidget
   protected EditingTool currentTool, activeTool, metaTool, altTool;
   protected ScrollViewTool scrollTool;
   protected PopupMenuManager popupManager;
-  protected int renderMode, gridSubdivisions, orientation, navigation, scrollBuffer;
-  protected double gridSpacing, scale, distToPlane, scrollRadius, scrollX, scrollY, scrollBlend, scrollBlendX, scrollBlendY;
+  protected int renderMode, gridSubdivisions, orientation, navigation;
+  protected double gridSpacing, scale, distToPlane;
   protected boolean perspective, perspectiveSwitch, hideBackfaces, showGrid, snapToGrid, drawFocus, showTemplate, showAxes;
   protected boolean lastModelPerspective;
   protected ActionProcessor mouseProcessor;
@@ -266,12 +266,8 @@ public abstract class ViewerCanvas extends CustomWidget
 
     if (mouseProcessor != null)
       mouseProcessor.stopProcessing();
-
     mouseProcessor = new ActionProcessor();
-    if (e.isAltDown())
-        scrollBuffer += e.getWheelRotation();
-    else
-        scrollBuffer += e.getWheelRotation()*10;
+
     final ViewerCanvas viewToProcess = this;
     final MouseScrolledEvent scrollEvent = e;
     mouseProcessor.addEvent(new Runnable() {


Sailsman63:
>As a possible future project user configurable button order migh be something to think about. What I have seen done with those are turning the mouse left handed or moving the center button function to a thumb button.

These sorts of adjustments are generally mapped at the OS/mouse driver level, though. Not something that an application needs to worry about. Definitely out of scope here, though.

------

We do need to get better at documenting features when they are added. Would you be up to adding appropriate documentation to the manual through this PR?

ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
```diff
@@ -26,7 +26,8 @@ Changes copyright (C) 2017 by Maksim Khramov
   private Properties properties;
   private int defaultDisplayMode, undoLevels;
   private double interactiveTol, maxAnimationDuration, animationFrameRate;
-  private boolean keepBackupFiles, useOpenGL, useCompoundMeshTool, reverseZooming, useViewAnimations;
+  private boolean keepBackupFiles, useOpenGL, useCompoundMeshTool, useViewAnimations;
+  private boolean reverseZooming, disableHScroll, reverseHScroll;
   private boolean drawActiveFrustum, drawCameraFrustum, showTravelCuesOnIdle, showTravelCuesScrolling, showTiltDial;
```
Sailsman63, line 31: I feel like we should maybe consider exactly how many customization options we make available. I realize that this is not entirely, or even mostly, due to this PR, but this set of options seems to be getting out of hand, and moving them to their own tab would only be papering over the issue. AOI wants to keep a fairly simple interface. This is not simple.

In particular, I question the `reverseHScroll` option here. I get that there is an ongoing dispute over what "scroll up" and "scroll down" mean, but do you have *any* precedent for people wanting reversed horizontal controls?

Sailsman63 (changes requested):

Sailsman63:
I would love to be able to merge this. I would like at least one other person to test it first, though, for both feedback purposes.

peteihis:
Sorry, I haven't been able to focus much on AoI in the last couple of weeks and next weekend may be still like that. I would like to rethink the implementation a bit. It works as it is now, but it is a bit 'patchy'. -- But I can do the thinking later in a different PR.

ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
peteihis: If there is no separate `reverseHScroll' then the reverse should affect both directions. Basically the wheel with both directions would have the similar logic to a joystick or a 3D-controller. Thumb wheel usage is probably a bit of a rarity and I suppose you'd want both wheels to work the same way anyway. While I was making it adding the check box just seemed the natural thing to do.

peteihis: > moving them to their own tab would only be papering over the issue

The mouse controls are not the reason why I wanted to split the preferences tab. It just adds on the pile. I have been thinking for a long time that it looks confusing as it is. It'd be much clearer to have things that belong together on their own tabs like general/input devices/display.

peteihis:
> These sorts of adjustments are generally mapped at the OS/mouse driver level, though. Not something that an application needs to worry about.

No, that is not what I am talking about. There are software that have this option and when I was using one it was actually quite common that some users did some of their own modifications. One of the usual things was to move the center button press to one of the thumb buttons. That SW relied entirely on the mouse to do view handling and it had mouse button combinations to do what AoI does with modification keys. Constantly pressing the the wheel started to hurt finger joints after a few hours. At those times I got my first 3D-controller so, luckily, it was not that much of a problem to me (but some 3D-modelers never learn to use the 3D-tool).

ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
Sailsman63: >If there is no separate `reverseHScroll' then the reverse should affect both directions. Basically the wheel with both directions would have the similar logic to a joystick

I've never seen a joystick-reverse-setting swap the H-controls. They're always about the V.

>The mouse controls are not the reason why I wanted to split the preferences tab. It just adds on the pile.

I agree. I'm not even opposed to re-ordering options and adding tabs. I just think that any effort in that direction should include taking a hard look at each of our options, and evaluating whether they are actually adding value to the application.

Sailsman63:
RE: Configurable mouse/key hookups. I've done a bit of digging, and you're right. Of the major OSs, only windows has built-in per-application key/mouse mapping. (Windows had to be the best at something, right?) Linux seems not to have an offering that can do this, and macOS depends on (paid) third-party tools.

So I'm coming around to the idea of having this be at least somewhat configurable within AOI. I'd like to see an idea of how this would look in practice. I wonder if it could be built as a plugin.

ArtOfIllusion/src/artofillusion/ScrollViewTool.java
```diff
@@ -46,10 +47,23 @@ public ScrollViewTool(EditingWindow ew)
...
+
+    if ((e.getModifiersEx() & 64) > 0) 
+    {
+        if (!prefs.getDisableHScroll()) // Can not be in the same if, would be handled as zoom.
+            horizontalScroll(e, v);
+        return;
+    }
+
     view = v;
+    scrollSteps = view.scrollBuffer; // Currently not used. Planned to avoid wheel 'slipping' with large scenes.
```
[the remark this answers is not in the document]
Sailsman63: If the `scrollsteps` variable is not used in the class, we should remove it - it's private, so cannot affect other code.

ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
Sailsman63: The more I think about it, the harder it is to get my head around *ever* reversing the HScroll Direction. Have you actually seen this done? If not, I think we should leave it out.

If there is actual demand, we can always implement it later.

Sailsman63:
Re: *this* PR. 

I think that they only way we'll get real feedback is to get this released on at least a nightly. To Do before merging:

- [ ] Remove or make use of scrollSteps
- [ ] Remove commented out experimental code
- [ ] Respond/help me resolve the HScrollReverse comments
- [ ] Document in the Manual. This would probably be an entry in the Mouse controls table on the ViewManagement page, and maybe additions to some of the Mode details as applicable.

peteihis:
Fresh start.

The parts that removed in [e9b7795](https://github.com/ArtOfIllusion/ArtOfIllusion/pull/258/commits/e9b7795c2f6ca9644565d89977bb9676b862acea) were left behind, when I moved the scroll wheel handling from VieverCanvas to ScrollViewTool. I forgot to remove them _(as I recall there were quite a few other things going on at the time)_. They are not used anywhere in the code and neither the SWT nor any remotely possible plugin could use them as they were declared protected.

Basically this PR could be merged as it is now as first aid to the weird behavior that horizontal wheel clicks do.

I'd like to figure out a concept to recognize cases, when pressing the wheel causes an unwanted h-click. Usually (with a mouse that has the h-scroll in the same wheel) the horizontal click happens first and the actual press-click second. This is simply because of the mechanical construction of the wheel motion sensors and in many applications that is not a problem. In our case it is.

I'm thinking of some kind of a stopwatch that would record the time of the h-click and if within a given amount of time center button press occurs the effect of the h-click, if already processed, would be reverted. I don't want to use a conventional timer there, because it would then slow down the response to the horizontal scroll events making the horizontal movement feel sticky.... It's probably note that difficult but would seem to require some preparation in both the VieverCanvas and the ScrollViewTool.

peteihis:
And the previous because It seems that I can not commit to keeping reasonable schedules lately.

Sailsman63:
Why the restart? I wasn't objecting to making use of the Horizontal scroll...

peteihis:
> Why the restart?

I was not happy at all with the way I had written it. The hScroll part was pretty much a glued on (prototype) piece instead of properly being integrated to the tool. The horizontal scroll events that were handled as zoom events should be filtered out though.

But it seems that I have had a blueprint of how it _should_ be done cooking in my head for a while as I just got it working, without much of struggle .... Shouldn't take long now to PR again.

peteihis:
I seem to have discovered a new problem that comes along with Java 11 and this center button + horizontal scroll click....

As long as the h-clicks went unnoticed, children of  bound cameras followed the camera quite nicely but since Java became 11, the scroll tool is activated at the same time with the center button down _(with the kind of a mouse that has this happen)_. Now when both of the tools finish their work at different times the positions of camera children are sometimes updated in the wrong order or at least from the wrong starting coordinates.... and camera children start to drift away from the camera object. 

Taking the hScroll clicks correctly to account helps a bit. Then I managed to work out some kind of a mechanism that should revert the unintentional hScroll click. That helps another bit but still the grip on the children starts to loosen pretty soon.... I'm not sure, what I'm missing. One other entity stirring the soup is the mouseProcessor that may be stopped in the middle of an action but ehh.. sometimes the damage seems to be done already at least in some extent.

There probably is a logical order in which to cancel and revert actions correctly but there are quite a few moving parts in the equation.

Sailsman63:
> children of bound cameras

You mean children of spotlights, etc?

peteihis:
> You mean children of spotlights, etc?

SceneCameras, SpotLights, DirectionalLights. Any object that can be controlled by a view.

peteihis:
One way of doing it would be to use undo records at every click. It should be an undo list dedicated to this purpose only and cleared after every successful operation.

Sailsman63:
> the scroll tool is activated at the same time with the center button down (with the kind of a mouse that has this happen). Now when both of the tools finish their work at different times the positions of camera children are sometimes updated in the wrong order or at least from the wrong starting coordinates.... and camera children start to drift away from the camera object.

This is most probably a weird artifact of running the mouse-view stuff through multiple interrelated tools the way we do. Why are the tools editing the position of the children directly anyway? That should be a side-effect of moving the parent object, and be handled in the scenegraph.

peteihis:
> Why are the tools editing the position of the children directly anyway? That should be a side-effect of moving the parent object,

Actually no.

When the children are not ancored to the parent by an animation keyframe, their position to the parent is not defined in any way, only their place in the model hierarchy. So when the parent moves the only place where the information of the movement exists is the tool that did it.  --> The tool will have to do finish what it started unless we want to add a tricky follow up system that knows what tool was used an what it did an if something that the tool did need more work.... ... work that the tool could have done in the first place.

Now Java 11 seems to have changed quite a few things and in many details AoI is not exatly up to it. One of them being the accidental activation of two tool simultaneously. It has been possible to mess thing up before by using the wheel scoll and move/rotate view fuctions simultaneously but now it has a kind of escalated. And what ever will be done in future, this issue with horizintal scroll in the middle button will need to be handled as a special case anyway....

But let's see what I hcan come up with.

Sailsman63:
>When the children are not anchored to the parent by an animation keyframe, their relative position to the parent is not defined in any way, only their place in the model hierarchy.

Now that you say that, I do remember running across that before. That *feels* like an oversight in the design that would be worth refactoring. I'd wonder what the ripple effects might be, and what would break?

PR #259, opened 2021-11-15
Update Groovy library

makiam:


Sailsman63:
though I agree that we should upgrade our groovy version, I'm considering holding off on this one for two reasons: 

1. At the (fairly slow) pace of AOI development, I suspect that Groovy 4.0 will be out of beta before we start our next release cycle.
2. I'm not fond of how the runtime seems to be expanding. If the above is true, I think that we'll be able to use a limited subset of groovy modules to limit our library size.

Sailsman63:
Groovy 4 is definitely released now. When we go for updates, we'll use that.

PR #260, opened 2021-12-07
Dispose of unmodified windows after delay courtesy of Runic

Sailsman63:
Workaround for a (weird) timing issue when disposing of GLJPanels on (some) macs. This should hold until the underlying issue is fixed in the JOGL libraries.

Needs testing on windows.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
+++ b/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
@@ -268,7 +268,6 @@ public class ArtOfIllusion
     
     numNewWindows++;
     SwingUtilities.invokeLater(new Runnable() {
-      @Override
       public void run()
       {
         LayoutWindow fr = new LayoutWindow(scene);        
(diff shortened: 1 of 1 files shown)


wizardlore:
Note this is for a specific case to avoid a hang when using File->New menu action with JOGL 2.4.0-rc1 + AOI plugins that maximize the new window with GLPanel(s) and then dispose of older unmodified window with GLJPanel(s). See https://sourceforge.net/p/aoi/bugs/510/ for more info.

Sailsman63:
@peteihis when you get a chance, can you check that this doesn't break anything on Windows? Test by opening or importing a file immediately after launching the application.

I'll add moving to the latest JOGL-rc in a little bit.

peteihis:
> can you check that this doesn't break anything on Windows?

Something seems to have gone wrong here. 
- I have an empty scene open, go **File --> New** it closes the original scene an leaves the new one open. This is as it seems to be with the old implementation.
- I add an object into the original scene and go **File --> New**, the new scene first opens and **then immediately closes**, so I can not for example create a new scene to copy parts of the old one into it.

The earlier version will not allow several empty scenes either but at least it allows one.

Personally I feel that this feature of AoI is a bit too aggressive. It's OK to close _the original_ empty scene, when I open a recent scene from the list or import a file into a new scene, but if I choose to open a couple of new empty scenes myself for what ever reason, I'd expect that to be possible. I might for example like to sort useful bits an old junkyard scene into a two or three new scenes at the same time or maybe just test different approaches side by side...  Even currently that _is_ possible but I'd have to do and undo something on each empty new scene to keep them alive.

I'm not saying what would be the best way of handling it but possibly the order of creation should matter?

Sailsman63:
Well, that's not intended behavior! I'll do some checking. (This is why I wanted another tester.)

Sailsman63:
This should do it - I'd gotten a check wrong (And I remember thinking it felt off when I wrote it!)

peteihis:
👍🏻

PR #99, opened 2018-12-29
Layout window split

makiam:
Refactor class to split huge and complex actionPerformed() into smaller parts. UI tests for Layout window.

PR #261, opened 2021-12-18
Split layout window commands.

makiam:
Refactor class to split huge and complex actionPerformed() into smaller parts.
This is replacement PR for old #99

### DIFF
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -458,11 +458,11 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
     importMenu = Translate.menu("import");
     exportMenu = Translate.menu("export");
     
-    fileMenu.add(Translate.menuItem("new", this, "actionNew"));
-    fileMenu.add(Translate.menuItem("open", this, "actionOpen"));
+    fileMenu.add(Translate.menuItem("new", this, "newSceneAction"));
+    fileMenu.add(Translate.menuItem("open", this, "openSceneAction"));
     fileMenu.add(RecentFiles.createRecentMenu());
     
-    fileMenu.add(Translate.menuItem("close", this, "actionClose"));
+    fileMenu.add(Translate.menuItem("close", this, "closeSceneAction"));
     fileMenu.addSeparator();
     Collections.sort(translators, Comparator.comparing(Translator::getName));
     
@@ -493,7 +493,7 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
     fileMenu.add(fileMenuItem);
     fileMenu.add(Translate.menuItem("saveas", this, "saveAsCommand"));
     fileMenu.addSeparator();
-    fileMenu.add(Translate.menuItem("quit", this, "actionQuit"));
+    fileMenu.add(Translate.menuItem("quit", this, "applicationQuitAction"));
   }
 
   private void createEditMenu()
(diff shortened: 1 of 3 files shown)


ArtOfIllusion/src/artofillusion/RecentFiles.java
```diff
@@ -26,15 +26,21 @@ Changes copyright (C) 2017 by Maksim Khramov
...
   {
     menu.removeAll();
     Preferences pref = Preferences.userNodeForPackage(RecentFiles.class);
-    String recent[] = pref.get("recentFiles", "").split(File.pathSeparator);
-    for (int i = 0; i < recent.length; i++)
+    String recentFiles[] = pref.get("recentFiles", "").split(File.pathSeparator);
+    for (String recent: recentFiles)
     {
-      final File file = new File(recent[i]);
+      final File file = new File(recent);
```
Sailsman63, line 43: This set of changes seems to have nothing to do with the main theme of this PR.

In addition, you are adding a public member to the API. What is the purpose here? Just trying to save a couple of lines of code in one very basic instance?

ArtOfIllusion/src/artofillusion/LayoutWindow.java
```diff
@@ -45,30 +45,41 @@ Changes copyright (C) 2017-2020 by Petri Ihalainen
...
+  private final BMenu editMenu = Translate.menu("edit");
+  private final BMenu sceneMenu = Translate.menu("scene");
+  private final BMenu objectMenu = Translate.menu("object");
+  private final BMenu animationMenu = Translate.menu("animation");
+  private final BMenu toolsMenu = Translate.menu("tools");
+  private final BMenu viewMenu = Translate.menu("view");
+  
+  
+  private BMenu recentFilesMenu, scriptMenu;
+
```
Sailsman63, line 68: We also need to be careful of changing access here. I know of at least one plugin that directly accesses the menus and rearranges things.

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/RecentFiles.java
makiam: Fixed this

ArtOfIllusion/src/artofillusion/LayoutWindow.java
makiam: Which plugin may affect this?

Sailsman63: ViewLayout for certain. The point is, this is a breaking change. What is the migration plan?

makiam: Take a brief look at ViewLayout code...
The plugin does not touches this Layout Window fields at all
at first it gets MenuBar via BFrame getMenuBar() than accesses Menu and Menu item via indexes.

Sailsman63: Thinking about it more, your response just makes me even more skeptical that this change is good or beneficial. Since any plugin *could* get around it by calling the `getMenuBar()`, why pretend that these fields are *not* accessible to any code that can get an instance of the layout window?

makiam: All LayoutWindow root menu items are accessible via getters.

Sailsman63: How is that an improvement over a public field?

makiam: Accessing fields via getters is better solution from incapsulation POV. Anoter reason is that fields now is not public but has only package private access. So hide them is a good decisoin for me

PR #264, opened 2022-01-13
Plugin Registry test

makiam:


Sailsman63:
Couple of basic feedback points:

- I'm having difficulty understanding what you are trying to test based on the test-case names. What is the failure mode that you are guarding against, in each case? *Note:* not every testcase name needs to start with "test"
- Do you expect any of those mock objects to be used outside of this particular set of tests? If not, they should be nested classes.

PR #267, opened 2022-01-29
Avoid NPE from Canvas

makiam:
Clicking on ObjectPreview canvas inside Tools dialogs trows NPE into log because of no animator set for this views

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -1436,7 +1436,7 @@ public abstract class ViewerCanvas extends CustomWidget
     Vec3 cp = pointInSpace.plus(cz.times(-distToPlane));
     coords.setOrigin(cp);
 
-    if(null != animation) animation.start(coords, pointInSpace, scale, orientation, navigation);
+    if(animation != null) animation.start(coords, pointInSpace, scale, orientation, navigation);
   }
 
   /** Show feedback to the user in response to a mouse drag, by drawing a Shape over the
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -1,6 +1,6 @@
 /* Copyright (C) 1999-2011 by Peter Eastman
    Changes Copyrignt (C) 2016-2020 Petri Ihalainen
-   Changes copyright (C) 2016-2018 by Maksim Khramov
+   Changes copyright (C) 2016-2022 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


Sailsman63:
What exact conditions does this affect? I've never seen this, and don't know how to duplicate the issue.

makiam:
See this exception from Tools previews, where Canvas placed inside dialogs but Amimator is not attached to it.
So to see this just open ex CSG tool dialog and click on preview. Exception is thrown into console.

peteihis:
What do you mean by "CSG tool dialog"? 

- The "boolean" tool in the Tools menu uses "CSGDialog", which uses "ObjectPreviewCanvas" which extends "ViewerCanvas" and nowhere it sets an animation engine an nowhere calls one either.
- The "CSGEditorWindow" extends "ObjectEditorWindow" which sets a new animation engine to each of the four windows it creates.

Anyway there is nothing wrong with ViewerCanvas. If there is a call to a part of it that has not been set in the first place then it _should_ show up as an error or exception. If some extension of ViewerCanvas has a call to an animation engine that is not present, then the correct fix would be to add the missing engine or to drop the failing call.

peteihis:
And I could not reproduce the problem at least with the Boolean dialog.

peteihis:
Ookay.... Sorry!

Forget what I just wrote, this thing is happening! What it takes is to center click on the viewer of an ObjectPreviewCanvas. 

OPC does not set an animation engine as there is nothing to animate, but it doe use RotateViewTool, which registers mouse up with no drag vector as the centering command. 

But I dont't think the correction should be done in ViewerCanvas to ignore the call. Rather the mouseRelesed of OPC should be rewritten to ignore the "empty center click". -- Unless we want to add an animation engine to the object previews?

peteihis:
> But I dont't think the correction should be done in ViewerCanvas to ignore the call.

And thinking again this fix **makiam** is suggesting might be the best or at least simplest way to handle it after all. Every time there is a `mousePressed` there needs to be a `mouseReleased`. Otherwise the ViewerCanvas will stay in dragging mode, thinking that the mouse is still being pressed.

In  `ObjectPreviewCanvas` the fix should look like this (tested):

~~~~Java
  private Point mousePressedPoint;
  ....

  @Override
  protected void mousePressed(WidgetMouseEvent e)
  {
    mousePressedPoint = e.getPoint();
    ....
 
  @Override
  protected void mouseReleased(WidgetMouseEvent e)
  {
    if (activeTool != metaTool || ! e.getPoint().equals(mousePressedPoint))
      activeTool.mouseReleased(e, this);
    else
    {
      mouseDown = false;
      rotating  = false;
    }
  }
~~~~

Then again, the idea was thrown in the air already before that the preview canvases could take a face lift...

Sailsman63:
Okay, got this duplicated. I think that we should probably do this for now as a quick-fix.

ArtOfIllusion/src/artofillusion/ViewerCanvas.java
```diff
@@ -1436,7 +1436,7 @@ public void centerToPoint(Point pointOnView)
     Vec3 cp = pointInSpace.plus(cz.times(-distToPlane));
     coords.setOrigin(cp);
 
-    animation.start(coords, pointInSpace, scale, orientation, navigation);
+    if(null != animation) animation.start(coords, pointInSpace, scale, orientation, navigation);
```
Sailsman63, line 1439: Condition order is backwards here. For readability, it should be `if(animation != null)`

Sailsman63 (changes requested):

Sailsman63 (approved):

Sailsman63:
Thanks!

Issue #266, opened 2022-01-29
Script editor windoew closing loses already saved changes


peteihis:
The script editor can be closed by the 'corner X' which is a good thing but the action of the WindowClosing event should always mean that the content of the editing table is NOT sent to the object but just discarded.

It has happened to me several times over now that:
- I have edited a scripted object, done the changes I want and have thing working. ... 
- Then I notice that there are a few instances of open script editor under the main window. 
- I close the unnecessary windows as there is nothing there that I need any more
- **And I lose all the recent changes...** Never knowing or noticing how far back the edit was reverted.

This is logical fail and it is extremely frustrating to lose your work because of something as simple as this!

Sailsman63:
Ouch. We should get this sorted out.

Question: How do you end up with multiple script editor windows in the first place? Should this even be possible?

peteihis:
> How do you end up with multiple script editor windows in the first place? Should this even be possible?

It happens the same way that it would happen with any AoI editor. You can end up having several editors open with the same object in them not realizing that yo do. The latest one you save the object with, will overwrite everything that was done or saved before. 

At the very least WindowClosinEvent shoud be routed to cancel and never ever in any imaginable case directly to "OK" or "save". Cancel may or may not ask if the changes made should be saved or not but the the common expectation is that WindowClosing just closes the window and makes no changes into anything at all.

The thing that by double clicking any object you can open it in as many editors as you like without the editors (or objects in them) being in any way aware of each others is an AoI wide problem. The usual practice in the rest of the world is that if you try to open an object that is already open in an editor, the existing editor is brought to front an activated and you simply cannot have several editors working on a (different copy of) the same object.

Sailsman63:
>It happens the same way that it would happen with any AoI editor. You can end up having several editors open with the same object in them not realizing that yo do.

This actually seems to be the core of the problem. I thought that this was supposed to be impossible, but it seems not.

I'd really like to have some consistency here. Either:

- All `ObjectEditorWindow` variants map `WindowClosingEvent` to `dispose`, (And I think that this has been discussed in the past, to much push-back from the user forum, as it makes it easy to accidentally drop your changes)
- Cannot open a second editor window for an object. Fixes you problem and still keeps the fail-safe. This would probably look like the "edit object" code searching for an already open window, and bringing it to the front if it exists.

peteihis:
> map WindowClosingEvent to dispose,

[...]

PR #271, opened 2022-02-13
Prototype for preventing multiple editors on the same object.

peteihis:
As discussed on #266.

- The prototype works on ScriptedObjects and the editor list is in the LayoutWindow. Now the editor list is an inner class, but as any EditingWindow can be a parent to sub-editors, it should be made a Java file of it's own.
- I also made the ScriptedObjectEditor a BDialog instead of BFrame. The difference is that now, when I have several objects open simultaneously they all appear on the front of the parent window, when one is activated. To me this actually seemed like a huge improvement into the situation where want compare two objects.
  - Now there still is that annoyance that new editors are always placed exactly on the same pixels right in the middle of the screen. AoI does not remember where the window (for that object) was the last time (though the user probably moved it for a reason) and also it does not bother to check where the last one of the similar kind was opened. A basic functionality from the beginning of graphic UI, that when several child windows are created they appear "cascaded".

### DIFF
--- a/ArtOfIllusion/src/artofillusion/LayoutWindow.java
+++ b/ArtOfIllusion/src/artofillusion/LayoutWindow.java
@@ -71,6 +71,8 @@ public class LayoutWindow extends BFrame implements EditingWindow, PopupMenuMana
   private List<ModellingTool> modellingTools;
   protected Preferences preferences;
 
+  public OpenEditorsList editor;
+
   /** Create a new LayoutWindow for editing a Scene.  Usually, you will not use this constructor directly.
       Instead, call ModellingApp.newWindow(Scene s). */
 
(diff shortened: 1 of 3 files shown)


peteihis:
I tried to make it more centralized and have the editor list take care of bringing the already open editor to front and clearing disposed editors more automatically but those ideas did not work as expected. There is "fail safe" system to remove editors that were disposed without removing from the list (can editors crash and close without freezing the entire AoI?) but it may have its limits.

Sailsman63:
I don't think that we should be creating another list of open windows here. The `artofillusion.ArtOfIllusion` core class already has one. Unfortunately, getting it to work as expected will mean some fairly extensive changes.

makiam:
Probably simpler implementation...
Inside ScriptedObject class declared field:
`private final Map<EditingWindow, ScriptedObjectEditorWindow> mapOwnerToEditor = new HashMap<>();`
and now method edit(....)

```
  public void edit(EditingWindow parent, ObjectInfo info, Runnable cb)
  {
    mapOwnerToEditor.computeIfAbsent(parent, (EditingWindow t) -> {return new ScriptedObjectEditorWindow(t, info, cb); }).setVisible(true);
  }
```

PR #276, opened 2022-03-23
Alternate build implementation via Gradle build tool

makiam:
To run just type 
`gradle run`

To create Intellij Idea project files type 
`gradle idea`

Known issue. Build script takes SPManager plugins from GH releases and this build fails on new java's due missed sun.misc.BASE64Decoder class

Good thing to do: relocate java and resource files to standard gradle project structure


Sailsman63:
This does not run properly.

- Splash screen states that it cannot find build data
- I get a message box with "Cannot find plugins directory"
- None of the core tools, renderers, etc. are available

makiam:
Thank You for Your feedback.

Let me explain this.
Once I execute `gradle run` it just compiles and starts main class artofillusion.ArtOfIllusion from corresponding sub-project
In this case there are no packaged application jar with manifest

Otherwise if run `gradle build` this produces artefacts from all subprojects, puts this artefacts to Plugins folder. Also this produces distribution archive under /ArtOfIllusion/ArtOfillusion/build/distributions
This archive can be copied/unpacked somewhere and run without gradle. And it shows proper build data in splash screen.

Sailsman63 (changes requested):
Why are you changing the app build structure at all? This seems to be orthagonal to updating the build system.

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
```diff
@@ -235,6 +219,13 @@ public static ApplicationPreferences getPreferences()
     return preferences;
   }
 
+  /*
+    Method is created for testing purposes to setup preferences at test time.
+  */
+  public static void setPreferences(ApplicationPreferences prefs)
+  {
+    preferences = prefs;
+  }
```
Sailsman63, line 228: This really seems off. Any time that you add a method for "Testing Purposes", figure that you are probably testing the wrong thing.

Sailsman63 (changes requested):
Why are you changing the built app structure anyway? It seems that such changes would be orthagonal to changing the build system.

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
```diff
@@ -136,7 +120,7 @@ public static void main(String args[])
     {
       UIManager.setLookAndFeel(UIManager.getSystemLookAndFeelClassName());
     }
-    catch (Exception ex)
+    catch (ClassNotFoundException | IllegalAccessException | InstantiationException | UnsupportedLookAndFeelException ex)
```
Sailsman63, line 123: Why limit to certain classes of exceptions? Do you expect others to bubble up and be more appropriately caught elsewhere?

Sailsman63 (changes requested):

Sailsman63:
RE: run command. So the actual correct way to launch is `gradle build run`.

I have some deep reservations about using a live build that is an implied stand-in for the actual project build structure.

I get a `NullPointerException` when I try to open the Textures and Materials dialog, because there is no `Textures And Materials` directory.

makiam:
gradle build structure a bit different from current one. All application jars put into lib folder and run shell script and bat file goes to bin folder.

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
makiam: Imagine You want to unit test some new code ex in ViewerCanvas class....
At first You try create instance of this class and only result is NPE because of uninitialized preferences.

makiam:
> Why are you changing the built app structure anyway
This isn't my idea but this is how gradle build works. Current build implementation puts ArtOfIllusion jar at one level and all other jars one level down in lib folder. And this libraries referenced from ArtOfIllusion.jar via Manifest Class-Path entry.

Gradle build system puts all app libraries together in lib folder. This gives difference in where to search Plugins folder.

> I'm also noticing that gradle run is still giving me "Build Data Missing".
Right. But if You run application from packaged distribution arhive (as end-user will do) You will see proper data

Sailsman63:
>     This isn't my idea but this is how gradle build works. Current build implementation puts ArtOfIllusion jar at one level and all other jars one level down in lib folder. And this libraries referenced from ArtOfIllusion.jar via Manifest Class-Path entry.

> Gradle build system puts all app libraries together in lib folder. This gives difference in where to search Plugins folder.

I'm sure that this can be overridden/configured around. I think part of the issue is that you are conflating "Art Of Illusion the modeling suite" with "ArtOfIllusion the core package and entrypoint". The distribution/full application should *not* be a child/under the `ArtOfIllusion/` build. Think of it more as a virtual package that requires all of the other jars, libs, etc. It should have a safe space to build the live/test version of the application without spilling `Plugins/` etc. out into the main working directory.

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
Sailsman63: *If* we need to add such a unit test, and *if* we don't find a better way to pre-configure the test, that would be the appropriate time to add a method. YAGNI.

makiam:
Unfortunately no. The current build system looks not so highly configurable.... At least I don' know how to do this. I think it is possible to write a lot of buld scripts to adopt default gradle build model to custom AOI structure I found this too expensive.

It is not clear for me why gradle model is not applicable for AOI application.
Looks like I failed as gradle advokate.

makiam:
I just added exclusion filter for distribution configuration.
So this is what we have after run 

> gradle clean build installDist

under ArtOfIllusion/ArtOfIllusion/build/install/ArtOfIllusion we have unpacked distribution which contains
lib folder with all application jars except jogl natives
Plugins folder with all plugins
both Scripts and Textures And Materials folders
and bin folder which contains scripts to run application under Linux (and probably mac os) and Windows bat script.

So this is all we need to further steps IMO

Sailsman63:
Hmm... That unpacked distribution doesn't find the Plugins folder when I launch it.

makiam:
SPManager and PostInstall is not an build artefacts but kind of stub projects to download plugins from remote locations

makiam:
Do the Plugins folder exist and not empty in unpacked distribution location? And You run on Linux or Win? I have no Linux host to check how all this works....

Sailsman63:
Yes it does exist. Includes all the Core plugins. Running on Linux. 

> SPManager and PostInstall is not an build artefacts but kind of stub projects to download plugins from remote locations

I thought gradle had a way to do custom dependencies?  You shouldn't need even one stub project for the SPManager/Postinstall pairing, let alone two.

makiam:
Running now from clean on Linux...
изображение

Sailsman63:
Oh... It only works if you are *in* the `bin/` directory.  Try calling that script from another location. That is an odd bit of fragile...

makiam:
Restore back APP_DIRECTORY evaluation code. Now startup script runs from any  point and app properl founds Plugins

makiam:
run 

> gradle site

 to create manual under ArtOfIllusion/ArtOfIllusion/build/site

running installDist now also create windows executable

makiam:
Last commit shows how to use some github project as gradle dependency. No need to store Buoy*.jar's locally

makiam:
Last commit update QT Writer internal library to maven reference. No more libs need to be stored in project

makiam:
Update project build script to produce windows/linux and mac builds
Now produces universal bundle for windows/linux x64 platforms and mac app bundle
Also add task createMacVolume to produce DMG image

Sailsman63:
I just re-tried running this, (`gradle run' on a completely clean directory) and can't get a compile.

 - I'm using java 11 or newer, I get an `IllegalArgumentException: Invalid source release: 18`  somewhere in the CompileJava task, probably at buildSrc/src/main/groovy/aoi.java-conventions.gradle#7  Why are we bumping to java 18 with no prior discussion?
 - When I try to compile with java 19, I get `exception in phase 'semantic analysis' in source unit 'precompiled_AoiJavaConventions' Unsupported class file major version 63`, which means something is trying to analyze a bit of bytecode that was compiled with Java 19, but the analysis tool can't cope with the newer bytecode.

makiam:
I update source compatibility level back to java 9 ( Forgot to change this earlier)
The second issue looks like You have old 6+ gradle but JDK 19 supports only from version 7.6

Sailsman63:
>second issue looks like You have old 6+ gradle but JDK 19 supports only from version 7.6

No. Classfile version numbers don't convert directly to java releases. Classfile version Sixty-four was released to support language features introduced in java 19.

Sailsman63:
Also, when I do get a completed build (using java 11 with your last update) I run into the following:

 - `java.lang.UnsatisfiedLinkError: Can't load library: /mnt/develop/Shared/Source/scratch/ArtOfIllusion/ArtOfIllusion/natives/linux-amd64//libgluegen_rt.so` I'm pretty sure that this one is due to https://github.com/jzy3d/jogl/issues/25, which I bumped my head on a while back.

makiam:
Is this linux issue reproduced with mill build too?
Is there a workaround for it?

Sailsman63:
^ It's not a linux issue. It's a "jzy3d mis-named the built files" issue. I did run into it with mill, and the workaround is to re-name the files. I do this when packing them for deployment. The line I'm using is `.foreach(p => os.copy(p, T.dest / "lib" / p.last.replace("-v2.4.0-rc4", "")))`, which essentially translates to "Strip the trailing version tag off the filename when copying to the deployment location."

Sailsman63:
That didn't fix it for me. I'll have to dig in a bit later and see what the actual file tree looks like.

What's up with the netlib stuff? (`netlib-java-0.9.3.jar` and `arpack_combined_all-0.1.jar`, packed in the `ArtOfIllusion.zip`) Where are they coming from, and what is their purpose?

makiam:
netlib and arpack are transitive dependencies of jama library (which is used by PolyMesh plugin)

Sailsman63:
> jama library (which is used by PolyMesh plugin)

Why are they in the main AOI build, then? Surely PolyMesh provides them directly...

peteihis:
> jama library

AoI uses **Jama**. More precisely it uses the solver, which is wrapped in `math.SVD` which is used by `Actor`, `IKSolver`, etc...

Polymesh uses **mtj**, which is included in PME library.

Sailsman63:
^Okay Looking through the code tree, I can see that. Though that's opposite of what you said above. The question remains: Why are you pulling in a dependency that is specific to Polymesh? AOI core *doesn't* use `mtj`.

peteihis:
> what you said above

Who said? ;)

Sailsman63:
^I'm getting tired. I wasn't watching the avatars or usernames. Fixed.

On the subject of the `UnsatisfiedLinkError`, I found that `gradle run` is running without the natives in the classpath, so not being able to find them is expected. I'd hoped that we would be working with a fully configured project here, just hadn't realized that.

makiam:
About jama/mtj...
I was tired too when I wrote this. Indeed the Polymesh uses mtj not jama. I set this dependency as part of internal experiments

makiam:
Now gradle run see jogl natives

Sailsman63:
Latest updates have broken `gradle run`. I now get the error `A problem occurred evaluating project ':ArtOfIllusion'.
> Resolving dependency configuration 'default' is not allowed as it is defined as 'canBeResolved=false'.` with multiple JDKs, with a reference to build.gradle, line 40.

makiam:
I unable to reproduce this error. Which versions of gradle and java You use for build?

Sailsman63:
Gradle 8.0 (Arch Linux likes to use up-to-the-minute)

OpenJDK 8
Temurin 11
Temurin 19

makiam:
This is it. I just install Gradle 8 and see this error.

PR #277, opened 2022-03-28
Mark deprecated code with annotation

makiam:


Sailsman63:
Thanks. I've been eying these warnings for a bit.

Issue #274, opened 2022-03-06
Migrate to build system with managed dependency support


Sailsman63:
It'd be nice to be able to automatically upgrade our dependency tree, instead of manually downloading and checking in new artifacts.

PR #281, opened 2022-04-09
Toolscript editor usability updates

firstmiddlelast:
Various changes intended to make things safer when editing tool scripts : 
- disabled manual language switching unless we are starting a brand new script, and added an icon to the language selector to make its state more obvious
- display full script path in the window title
- shift successive editors to the bottom right so they don't overlap
- move load/save buttons top
- add a 'save' button that's not prompting for a file name
- remove the close button, and display a save/discard/cancel dialog if the script has unsaved changes
- display a warning and go read-only if trying to edit a script that's already open for editing
- added a menu for recently opened scripts (persistent)
- added "new" and "recent" sub-menus under the current "edit tool script" menu
- see also the discussion at [https://github.com/ArtOfIllusion/ArtOfIllusion/issues/266], with screenshots
- Sorry for adding so many things in one PR, it's my first! The next ones will be much more atomic.


Sailsman63:
Thanks for your first pull request! I like where this is going. I'm going to go through and comment on certain parts of the code - this is a normal part of the process, so please do not be discouraged if I ask you to explain or change some things! If you do not understand or agree with the suggestions, please reply with your viewpoint. This will help us to improve the code before it is merged.

A couple of notes about the PR overall:

- In future PRs, please do *not* merge the AOI main branch into your feature branch as you work. Those merge commits make reviewing and cleanup a bit harder. (Don't worry in this case. We can make it work.)
- When you find that you are doing some refactor work, as well as whatever feature/bug you are working on, please make them in two separate commits. f8bf01e72c1311f78ec750b386061158d350996d would be easier to review if the cleanup was separated out into a first commit. (Again, we can work with it. Just a heads up for the future.)

Unfortunately, I've also run across two bugs:

- If I cancel out from the "Save As" dialog, I consistently get the `errorWritingScript` message with `</path/to/Tools/><path/to/Tools/OriginalName> (no such file or directory)` I have not dug into why this happens.
- The recent scripts list seems to add a script whenever I click the "save" or "save as" buttons - even if I cancel out of the "Save As"! It also does not de-duplicate, so there can be multiple entries pointing to the same script.

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
```diff
@@ -410,23 +410,24 @@ private static void runStartupScripts()
...
+            catch (IOException ex)
+            {
+              ex.printStackTrace();
+            }
+          }
+        else 
         {
-          ex.printStackTrace();
+            // TODO translate this
+            // TODO use some debugging tools instead of System.err
```
Sailsman63, line 428: We have #50 open as a tracking issue re:proper logging. It's on my list for after #274 is resolved.

Overall, I think we'll get more mileage out of having an issue open for, say, "Translate remaining hardcoded strings" than we will from TODO comments - unless you intend to resolve the comment as we fine-tune the PR.

ArtOfIllusion/src/artofillusion/script/ScriptRunner.java
```diff
@@ -56,8 +59,12 @@ else if (language.equals("BeanShell"))
           }
         catch (Exception e)
           {
+            // TODO FIXME NOOOOOO
             e.printStackTrace();
           }
+        // TODO Use a ScriptOutputPanel or something like this
```
Sailsman63, line 65: Also a good spot for a tracking issue so that we can discuss what that should look like overall.

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/script/ScriptRunner.java
```diff
@@ -21,9 +21,12 @@
 
 public class ScriptRunner
 {
-  public static final String LANGUAGES[] = {"Groovy", "BeanShell"};
+  public static final String [] LANGUAGES = {"BeanShell", "Groovy"};
+  public static final String [] EXTENSIONS = {"bsh", "groovy"};
```
Sailsman63, line 25: This is... not my favorite way to go about it. The issue is that you now have to remember/track back what index is which, such as lines 48 & 50.

If we've outgrown hard-coded strings here - and I'm not sure that we have - we would want an enum with fields for the name and extension.

ArtOfIllusion/src/artofillusion/script/ScriptRunner.java
```diff
@@ -153,14 +160,14 @@ public static int displayError(String language, Exception ex)
...
-
+  /** Given the name of a file, determine what language it contains based on the extension. 
+   @return null if the language is not recognized*/
   public static String getLanguageForFilename(String filename)
   {
     for (String language : LANGUAGES)
       if (filename.endsWith("."+getScriptEngine(language).getFilenameExtension()))
         return language;
-    throw new IllegalArgumentException("Filename \""+filename+"\" does not match any recognized scripting language.");
+    return null;
```
Sailsman63, line 170: Returning null is generally not what we want to do, especially when it should be fairly rare.

(A lot of AOI code does return null/uses null to signal things. This is old code, and would usually be written differently today. We just haven't gotten around to fixing it yet.)

Not sure if there's a better way than throwing an exception - maybe `Optional()`? I'd have to look closer at where it's actually used.

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
```diff
@@ -15,7 +15,21 @@
...
+import java.awt.event.KeyListener;
 import java.io.*;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Date;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.prefs.Preferences;
```
Sailsman63, line 28: Generally, if you're importing this many classes from a single package you might as well just wildcard `import java.util.*;`

Sailsman63 (changes requested):
A few more things to keep in mind.

ArtOfIllusion/src/artofillusion/LayoutWindow.java
```diff
@@ -2184,18 +2225,46 @@ public void transformObjectCommand()
...
+    public static Mat4 determineRotationMatrix(double x, double y, double z) {
+        Mat4 m = Mat4.identity();
+        if (!Double.isNaN(x))
+            m = m.times(Mat4.xrotation(x*Math.PI/180.0));
+        if (!Double.isNaN(y))
+            m = m.times(Mat4.yrotation(y*Math.PI/180.0));
+        if (!Double.isNaN(z))
+            m = m.times(Mat4.zrotation(z*Math.PI/180.0));
+        return m;
+    }
```
Sailsman63, line 2250: Does this double up on some of the options available in the `artofillusion.math` package? See `Mat4` and `CoordinateSystem`

If not, should this option be over there instead?

ArtOfIllusion/src/artofillusion/LayoutWindow.java
```diff
@@ -2184,18 +2225,46 @@ public void transformObjectCommand()
...
+        for (int i = 0; i < sel.length; i++)
+        {
+            ObjectInfo info = scene.getObject(sel[i]);
+            if (bounds == null)
+                bounds = info.getBounds().transformAndOutset(info.getCoords().fromLocal());
+            else
+                bounds = bounds.merge(info.getBounds().transformAndOutset(info.getCoords().fromLocal()));
+        }   
+        return bounds.getCenter();
+    }
```
Sailsman63, line 2267: Do you see any reason that these might need to be called from outside the class? If not, they should not be public.

ArtOfIllusion/src/artofillusion/RecentFiles.java
```diff
@@ -23,7 +23,7 @@ Changes copyright (C) 2017 by Maksim Khramov
 
 public class RecentFiles
 {
-  private static final int MAX_RECENT = 10;
+  public static final int MAX_RECENT = 10;
```
Sailsman63, line 26: This is making me just a bit twitchy...

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
```diff
@@ -30,22 +44,104 @@ public class ExecuteScriptWindow extends BFrame
...
+    }
+    
+    public static String [] getRecentScripts()
+    {
+        final Preferences pref = Preferences.userNodeForPackage(ExecuteScriptWindow.class);
+        // The files are ordered by ascending Date, so we need to revert this to get the most recent in to
+        final java.util.List<String> recentScripts = Arrays.asList (pref.get("recentFiles", "").split(File.pathSeparator));
+        Collections.reverse(recentScripts);
+        return recentScripts.toArray(new String [0]);
+    }
```
Sailsman63, line 101: ... And when combined with this, I think that we can do a bit better. I'd favor a refactor of `RecentFiles` that allows the caller to pass a marker class. I've been working on another project that would also benefit. That way we don't have essentially the same implementation floating around in triplicate.

Sailsman63 (changes requested):
And some structural stuff.

ArtOfIllusion/src/artofillusion/LayoutWindow.java
```diff
@@ -2112,70 +2162,61 @@ public void transformObjectCommand()
...
         Vec3 neworig = orig.minus(center);
-        if (!Double.isNaN(values[6]))
-          neworig.x *= values[6];
-        if (!Double.isNaN(values[7]))
-          neworig.y *= values[7];
-        if (!Double.isNaN(values[8]))
-          neworig.z *= values[8];
+        neworig.multiply(resize);
         coords.setOrigin(neworig);
+        // We actually don't need to do this if no rotation has been specified
```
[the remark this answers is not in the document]
firstmiddlelast: I have a habit of using the source files I'm working on as development personal notepads... :) (hence the list of TODO comments) I have to lose this habit, or be consistent with the removal of those notes before committing.

ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
firstmiddlelast: That's what my TODO comments are for - resolving them as the PR evolves. And I did translate this in my working copy, was waiting for feedback before adding other commits to the PR.

ArtOfIllusion/src/artofillusion/script/ScriptRunner.java
firstmiddlelast: Well, my reasoning was each tool script should have its own output and error (as in stdout /  stderr) panel ; but I wanted to have feedback on the PR before submitting such a change. Should this be another PR?

ArtOfIllusion/src/artofillusion/script/ScriptRunner.java
firstmiddlelast: I'm so happy about this comment! This is really not the way I would have written this myself, but for a lot of things I'm trying to blend into the existing source and coding habits that I see, and I haven't seen a lot of enums anywhere, so I thought I wouldn't introduce new coding practices.. but yes, I 100% agree : enums and/or Maps are obviously the way to go here ; and those static strings should be defined in the *ScriptEngine classes, their values retrieved through methods like getLanguageName() and getFileExtension(), so that new scripting languages implementations (I'm thinking Javascript / Python) would be automatically supported. But I have seen so little interfaces and OO design patterns in the code that I ruled that out.

ArtOfIllusion/src/artofillusion/script/ScriptRunner.java
firstmiddlelast: I generally agree about nulls not being what we want to return, but on this particular case, I disagree : the fact that a file extension is not recognized only means no ScriptEngine class matching this language extension is implemented, not that the filename is an "illegal argument". If this method should really throw an exception, it should be a NotImplementedException. But generally speaking, exceptions are not meant to be used as a programmatic logic. In this function, the code was already checking for all possible values and should have taken the "language not found" as a possibility from the start (think switch/case/default). 
Instead of null, a String (such as, say, "UNKNOWN_LANGUAGE" or even "") could be a valid return value, without needing Optionals.

ArtOfIllusion/src/artofillusion/script/ScriptRunner.java
Sailsman63: Yeah, let's consider this idea out-of-scope for this PR. Smaller bites are generally more efficient to review and integrate when possible.

ArtOfIllusion/src/artofillusion/RecentFiles.java
firstmiddlelast: I understand! When I coded "recent scripts" persistency, I looked into the way AOI handled the "recent scenes" list, and I saw this constant. It seemes pointless to create another identical constant for the maximum length of "recent scripts" list, because actually these constants would achieve the same goal : describe the maximum number of items in a sub-menu listing files. So I made it public for reuse.
Ideally, the number of files in both the "recently used" scenes and scripts list should be a user preference, but I didn't want to extend the scope of this PR any further. What do you think about this reasoning? Should I actually code this "user pref" feature and add it to the PR?

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
firstmiddlelast: Aaaaaboslutely! I just didn't want to also refactor the Recent Files class (even though I did feel the urge) and keep code changes smaller for a first PR.

ArtOfIllusion/src/artofillusion/LayoutWindow.java
firstmiddlelast: First, a quick _mea culpa_ : those changes are actually not related to this PR, but I forgot about them when submitting the PR. They should have been in another branch, but when I made them I hadn't created one, and I forgot about these changes when submitting the PR so here they are, out of place. 

Those changes were actually made in correlation with my re-creation (through scripting) of the seven AOI splash screens ("title images"). And those changes are really not additions to the code, the GIT diff got a little lost here : all I did was refactor and isolate those two methods, but all of this code was already present in the `transformObjectCommand()` method. 
As a side note, I totally agree they belong in a math package and another class, and not in this class (which suffers from a huge God-Class syndrome - in other words, it's way too big and does way too many things). 

And now, back to the reasoning about why I made those changes : I feel Scripts and plugins should be able to easily invoke standard AOI Object menu tools (in this case, I needed to invoke the "Object/transform Object" menu item, but the reasoning applies to all of the "Object" menu  items and most of the Tools). I mean, doing things with code instead of by hand_is_ basically the purpose of scripting, isn't it? 

So, if the basic AOI object manipulation menu items are to be invoked by scripting / plugins, the actual "working" and utility code should be smore eparate from the GUI code, and be made public - ideally, in separate classes (`BasicTools` or something equivalent), without any reference to GUI code. 


In the end, I'm sorry this code change appears in this branch and PR, but remember all of it was already present in the class source code, the change is really just some lines moved around. 

What do you think of my reasoning about 

- separating GUI code from "work" code, 
- LayoutWindow doing too many things, and 
- what scripts should be able to do / what should be public?

firstmiddlelast:
> * cancel out from the "Save As" dialog

I haven't tested this but I think I have an idea what's happening here. 

> * It also does not de-duplicate,

Quite surprised about this one, I spent a bit of time making sure it was not possible to get duplicates. I'll try to reproduce. 

> they could look pretty daunting. Don't be discouraged!

Thanks for the nice words, but don't worry, I understand code review != code acceptance. 
I didn't expect to nearly spend more time reading and writing plain english than source code, though :).

ArtOfIllusion/src/artofillusion/script/ScriptRunner.java
Sailsman63: Certainly exception-driven logic is not ideal, either. The problem with returning `null` is that it can mean anything, and therefore means nothing. It's also very easy to forget to handle the `null` case. Much better to return a named constant of some sort, or other explicit flag, just to force calling code to handle the case. This is another place where I could imagine an enum or similar being handy.

ArtOfIllusion/src/artofillusion/LayoutWindow.java
Sailsman63: > I forgot about them when submitting the PR. They should have been in another branch

Ahh... That happens. We should figure out where, exactly, they belong. (& what signature) I really don't want to expose them as `public` when we are very likely to be moving them elsewhere.

I'd agree that `LayoutWindow` is very large, and does a lot of things. If some of them are more naturally a separate concern, we should figure out where/how they should be used. But that's another project.

Separation between GUI building and business logic would be welcome. I'm not sure that we need to break them out into separate *classes*, as sometimes the UI itself needs to look at the data closely. (See for example the code for all of the raytracer options. Some of them depend on other options. Would *you* want to figure out how to communicate which parts of the UI are supposed to be disabled through an interface?) Moving logic into a set of methods with a public API - and no implicit dependencies on having a desktop environment available - would definitely be something I would support. Way out of scope for this PR, though.

Yes, I think that scripts should be able to leverage most AOI code, and PRs in support of that are welcome. Do expect some discussion on getting the exposed API right in those cases.

ArtOfIllusion/src/artofillusion/RecentFiles.java
Sailsman63: I'm a bit skeptical about making this configurable - it's really tied in with UI design. How do you display a long list? I really don't want to expose the constant as public, though.

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
Sailsman63: I actually wonder if a refactored `RecentFiles` might be a smaller change overall? I'm thinking an API something like this:

```
//All these methods take a class argument. Usually `this.getClass` for the calling location
public static void updateRecent(Class interfacePoint, File newestFile)

public static List loadRecent(Class interfacePoint)

public static BMenu createMenu(Class interfacePoint)

@Deprecated(forRemoval=true) // Just for back-compatibility - 
public static void addRecentFile(File file)

@Deprecated(forRemoval=true) // Just for back-compatibility
public static void createMenu(BMenu menu)

**Comment by Sailsman63** [comment]
> I didn't expect to nearly spend more time reading and writing plain english than source code, though :)

Well, we also are discussing program architecture... :)

**Comment by firstmiddlelast** [comment]
> * If I cancel out from the "Save As" dialog, I consistently get the `errorWritingScript` message

This is due to a bug is already present in the existing code : pressing the "Cancel" button is the same as pressing "Save" if you have selected a file, because the only thing that's tested to detect is whether or not a file is selected (instead of using the return value of the `showDialog() `method. Same thing for loading files : if you select a file then press cancel, the file is currently loaded anyway. This bug is also present on the Scripted Object Editor file dialog, too because the code is pretty much a copy / paste, so beware!
My next PR will be something like "the tool script editor and scripted object editor classes should both inherit from a common script editor class", so I'm not fixing this in the scripted object editor. 

> * It also does not de-duplicate

This was linked to the previous bug, so this should disappear as soon as I commit the bugfix changes. I don't think there is a way to clear the recent files list, so you'll have to load 9 different files in order to get 10 unique files, and then test. This seems clumsy - should there be a way to clear the recent files list? 

I'm also adding a filename extension filter for known script types.

File: ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java (hunk 9 above)
Comment (firstmiddlelast) [reply ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java:101]: I like the refactoring idea - and recent files management is a feature that could also be reused for anything in the future - I'm thinkin loading / saving of meshes, procedural textures, renderer settings, whatever comes next. 
One critic, though - I don't like the current code logic of rebuilding the "recent" menu everytime a file is loaded / saved ; I think a "rebuild on-demand" (when the menu is actually opened) would be smarter. 
Maybe leaving it as is for this PR and then grouping a refactored "recent files" class, plus this "rebuild menu on-demand" feature should be a separate PR? Or is it a small enough change that it could be grouped with this PR? What do you think?

File: ArtOfIllusion/src/artofillusion/script/ScriptRunner.java (hunk 3 above)
Comment (firstmiddlelast) [reply ArtOfIllusion/src/artofillusion/script/ScriptRunner.java]: I committed changes so we now use 
> an enum with fields for the name and extension.

..and also a field for the corresponding ScriptEngine class. In most of the code, the language is still implemented as a String (so I have added an UNKNOWN_LANGUAGE constant and used its value wherever I have found a "== null" test), 
and although I would have liked to change things a little more by using the enum value instead of the String wherever possible, I decided to respect the "if it's not broken, don't fix it" coding wisdom and stop here for now. 
As a side note, this change should make implementing a new scripting language easy (I'm thinking javascript with the built-in Nashorn engine).

Comment (Sailsman63) [reply ArtOfIllusion/src/artofillusion/script/ScriptRunner.java]: I'm feeling very mixed about the changes in this area:

- On the one hand, I do think that enums are the right solution if we grow to have multiple *built-in* scripting solutions.
- On the other hand, I think that you would need to make a *very* strong case to add other scripting engines to the core application
- If scripting engines are to be supported by plugin, that's a whole different set of API, and enums will *not* work in that case.

How do you see other scripting environments being supported? What are your reasons to want to add them?

File: ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java (hunk 9 above)
Comment (Sailsman63) [reply ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java:97]: Not sure which would be better at this point. I do have a project of my own that would benefit from a refactored/extensible RecentFiles implementation. Let's leave it as is, and do the refactor later. (Probably still before next release, but after this PR)

**Comment by Sailsman63** [comment]
> I don't think there is a way to clear the recent files list, so you'll have to load 9 different files in order to get 10 unique files, and then test. This seems clumsy - 

The original RecentFiles implementation avoids this entirely by running what is essentially a somewhat naive LRU cache that always kicks out the older copies of a filepath. Your implementation is doing something with timestamps that seems unnecessary, and might be a bit fragile?

File: ArtOfIllusion/src/artofillusion/keystroke/KeystrokePreferencesPanel.java [hunk 11]
```diff
@@ -133,7 +133,8 @@ private void editRecord()
 
   private void addRecord()
   {
-    KeystrokeRecord record = new KeystrokeRecord(0, 0, "", "", ScriptRunner.LANGUAGES[0]);
+    // Beanshell is the only supported language here 
+    KeystrokeRecord record = new KeystrokeRecord(0, 0, "", "", ScriptRunner.Language.BEANSHELL.name);
```
Sailsman63, line 137: This is not actually true. Both Beanshell and Groovy are fully functional options here. It's just that creating a new record *defaults* to one of them - and, in fact, the main branch defaults to *groovy*. The language can be edited while editing the actual keystroke script.

firstmiddlelast: Hey, I'm having trouble understanding what's happening here on GitHub. 
After your comment about the main branch defaulting to Groovy, just to be sure I didn't by mistake change the default to Beanshell, I am trying to check the original version of this file on GitHub, in the master branch of AOI/AOI [here](https://github.com/ArtOfIllusion/ArtOfIllusion/blob/master/ArtOfIllusion/src/artofillusion/keystroke/KeystrokePreferencesPanel.java) and what I'm seeing is changes that I have only introduced in the first commit of my branch (see line 136,     `KeystrokeRecord record = new KeystrokeRecord(0, 0, "", "", ScriptRunner.LANGUAGES[0]);`) for this PR. 
Is this how it's supposed to be? Am I looking at the right place?

ArtOfIllusion/src/artofillusion/script/ScriptRunner.java
firstmiddlelast: I agree with all your points, but honestly, I don't really see the issue with adding javascript support (or any other language for that matter) except maybe for code maintenance -even though the code and build process changes involved with javascript look pretty small to me, since it's natively included-. 
Anyway, if you ask me, I would support JS natively with enums, and other languages via plugins, but I'm really not locked in onto any decision right now. 

My reasons to want Javascript (and Python) are long-term projects related to generative art and machine learning.. that, at the current speed things are going (I mean, for me - the time I can spend weekly on AOI has diminished quite a bit since I first started posting on the SF and GH forums :/), are not going to be visible for a long time. So, I'm not going to try hard and push in any direction, and will just go with the flow. 

For now, I would just suggest sticking to enums because it's clean and explicit, and I'll talk again about adding JS (or any other language) again later, when I can give more specific arguments.

ArtOfIllusion/src/artofillusion/keystroke/KeystrokePreferencesPanel.java
Sailsman63: That line is in the original file. `ScriptRunner` was already using an array for "languages", you had added one for the file extensions.

ArtOfIllusion/src/artofillusion/LayoutWindow.java
Sailsman63: It looks like you backed out the changes to the code that *used* these methods, but have left the methods themselves? These should wait until we figure out where they should live.

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
```diff
@@ -30,22 +38,97 @@ public class ExecuteScriptWindow extends BFrame
...
+
+      @Override
+      public void keyPressed(KeyEvent e) {
+      }
+
+      @Override
+      public void keyReleased(KeyEvent e) {
+      }
+        
+    }
```
Sailsman63, line 99: Hmm... I wonder if we might be better off switching to a `TextEditorPane` and listening for a `PropertyChangedEvent`. Perhaps something to consider when refactoring to a unified script editor.

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
```diff
@@ -30,22 +38,97 @@ public class ExecuteScriptWindow extends BFrame
...
     setContent(content);
     window = win;
-    if (scriptDir == null)
-      scriptDir = new File(ArtOfIllusion.TOOL_SCRIPT_DIRECTORY);
-    scriptText = new RSyntaxTextArea(lastScript, 25, 100);
+    String editorTextContent = "";
+    if (scriptLanguage != ScriptRunner.UNKNOWN_LANGUAGE && scriptAbsolutePath.contains(".")) {
+        editorTextContent = ArtOfIllusion.loadFile(new File (scriptAbsolutePath));
+    }
+    scriptText = new RSyntaxTextArea(editorTextContent, 25, 100);
```
Sailsman63, line 130: I think that we may have to change the layout used by the `BorderContainer`

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
```diff
@@ -101,56 +274,70 @@ private void loadScript()
...
   }
 
-  /** Prompt the user to save a script. */
+ /** Prompt the user to save a script. */
 
-  private void saveScript()
+  private void saveScriptAs()
   {
     BFileChooser fc = new BFileChooser(BFileChooser.SAVE_FILE, Translate.text("saveScriptToFile"));
+    // QUESTION Are all those directory changes really needed?
```
Sailsman63, line 332: Yes. See #263 and the linked discussion for why we added this.

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
```diff
@@ -168,66 +355,121 @@ private void saveScript()
...
-      for (int i = 0; i < allWindows.length; i++)
-        if (allWindows[i] instanceof LayoutWindow)
-          ((LayoutWindow) allWindows[i]).rebuildScriptsMenu();
+        for (EditingWindow edWin : ArtOfIllusion.getWindows())
+        {
+            if (edWin instanceof LayoutWindow)
+            {
+                ((LayoutWindow) edWin).rebuildScriptsMenu();
+            }
+        }
```
Sailsman63, line 371: Hope this doesn't sound too nit-picky, but a bit of a style comment:

We generally don't use braces for one-line `if (condition) doThis;` statements.

I also see that at least some of your additions are indented 3 spaces, rather than the 2 that we try for. (I am aware that not all of our existing code matches this.) Is this something that we can make easier to set in an IDE's auto-formatter?

firstmiddlelast: I don't use braces for one-ifs either ; this is a mistake, thanks for pointing it out and for the reminder about the 2 spaces. I'll check my Netbeans formatting settings.

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
firstmiddlelast: Didn't know this class - thanks for the link. I agree, will try this out.

ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
firstmiddlelast: This looks like a bug to me, because this strange behaviour appears depending on whether or not scrollbars are visible in the text area : try alternating Windows-Right, Windows-Up, Windows-Down with different files and line lengths, you'll see it (I hope this behaviour is the same on all platforms..). 
And also, the scrollbars sometimes do not appear when they should. I have absolutely no idea where to look for for the root cause of the bug : RSyntaxTextArea, Buoy, AOI? Layout, as you mentioned, events, scrollbars? 
I'd be willing to let this bug live for now and let someone who is more familiar with the code than I am hunt it, as it's not really critical nor blocking - apparenty no one had realized its presence until now..?

Sailsman63: And I was wrong. Somehow, the maximize (and restore) events are not triggering as a `WindowResizedEvent` repaint on the `RSyntaxTextArea` - or more probably the `RTextScrollPane` or `AWTWidget` that it is wrapped in.

This has probably not been noticed because most of us don't actually use the script windows maximized, and manually re-sizing the window *does* work correctly. Clicking in, or attempting to scroll the script pane also triggers a repaint.

Sailsman63: Now has it's own issue: #283

Sailsman63:
I think that we could be just about ready to merge this, unless the change to a TextEditorPane would be simple to implement.

Do you have any other points that you think need addressing before we turn this loose on the alpha-testers? I'd like to get this out to be tested by a broader set of our users well before we think about doing an actual stable release.

I do appreciate the format fix that you checked in, but the other changes that you've made do need a going-over - a good bit of the changed code has slightly inconsistent indenting, and some K&R "Egyptian" braces, where AOI tries to stick to "Braces on Left"

firstmiddlelast:
I was not thinking of adding any more commits, and was waiting for your feedback. Thanks!
But I don't know exactly what the next steps should be - I'm sorry but I'm still a newbie about the PR ending process.. 
Will you merge this into master, and are we done with this branch after that? Should I commit code style changes to fit the AOI indenting and bracing style before, or after the merge if the branch stays alive?

Sailsman63:
> Should I commit code style changes to fit the AOI indenting and bracing style before, or after the merge if the branch stays alive?

Definitely before, much like the "Code formatting and cleanup" commit. You don't need to go through the whole of the files you worked on - in fact, please don't - just the bits that include your changes. If you're not sure what lines to check, click the "files changed" link near the top of the PR page here.

> I don't know exactly what the next steps should be - I'm sorry but I'm still a newbie about the PR ending process.

When the above is done, I will merge into master. At that point, this branch can be deleted. Any new changes should then be built on top of the new master.

Sailsman63:
Thank you! I'm pretty sure that this is ready to go, but I just noticed a... limitation. If you "Run to Cursor", there is no way to *continue* from that point. You can only start a separate run. This means that certain kinds of scripts, in certain situations, leave behind orphaned stuff: Dialogs that cannot be closed without shutting down the program, for example.

Thoughts on this?

firstmiddlelast:
Here they are : 
- My implementation of "run to cursor" is really simple and actually means "run the script that has been truncated after the end of the line where the cursor is". I added this option because I wanted to test the beginning scripts and didn't want to repeatedly add and remove "return" statements for the sole purpose of stopping the script from running to its end, and I thought that it was a common enough development process that other sript authors would benefit this feature. 
- As such, this is nothing like a debugging tool such as a breakpoint or a pause/continue feature. Such a feature would come in another PR, with other debugging options (I'm thinking the PR where scripted objects and tool scripts inherit the same editor class), as we discussed earlier. 
- The "run selected" works the same : it creates a new script with the selected text and runs it, so the same limitation applies. 
- If you think some clarification is required, maybe a differently named button would make things clearer? Or some informative help bubble or text? 
- The current `ScriptEngine `interface gives no hint about whether a script that's to be executed runs in the same environment as the previous one. And actually, both implementations (Beanshell and Groovy) script run scripts in fresh environments everytime, restarting from scratch. My opinion is there is work to do on both the interface and the implementations in order to implement such a "pause / continue" and / or "keep the environment" feature, and also in the GUI in order to make it clear whether the environment is preserved or discarded before each run. This, again, looks like material for a "script debugging" PR. 
- If you think it is not possible to make things clear for the user and there is a chance of catastrophic (=work lost) failures, maybe the buttons could be disabled, hidden or removed until an appropriate PR is on track to re-enable them. 
- More generally, you raise another interesting and important point : how do we interrupt tool scripts (or even scripted objects) that get stuck into long operations and/or infinite loops? Maybe we would need to launch them into separate threads, and add a visual interface to manage them (and most importantly, stop them). I think this feature should be implemented so it can manage any possibly long task (like rendering -the current "cancel" button when rendering could be moved to this new GUI interface and the window made non-modal), long modelling operations, file format translations, movie creation, etc.) that may have to be paused or stopped, into a single visual interface (ideally, with visual clues about the task progression). The most obvious example I have in sight right now is the Netbeans task bar. 
As a side note, I had to kill the AOI java process yesterday (and lose scene changes) because it went into an infinite rendering loop no later than yesterday, so this "task manager" feature is something I'd appreciate :). Another PR awaits!

Sailsman63:
Okay, good points. Sounds like you've given this thought, and have a planned outline. I think that we can merge this to get it to our alpha testers. If it turns out to have issues in practice, we can disable them before making an actual release. I have not yet figured out a way to generate a "lost work" scenario with the "run to"/"run selected" buttons.

Being able to interrupt scripts, or at least have some feedback on what they are doing, sounds handy.

> I had to kill the AOI java process and lose scene changes because it went into an infinite rendering loop no later than yesterday

If you know how to reproduce that, please post it elsewhere. Either as a bug (if you've got a good idea what's going on) or in the help forum, if it's going to need troubleshooting.

firstmiddlelast:
Unfortunately, it occurred after a long series of (small texture tweak/render/small modelling tweak/render/small lighting tweak/render/etc.), so I can't reproduce it.. and I have no idea why the renderer got suddenly stuck, so I don't have anything to post.

Sailsman63 (approved):

Sailsman63:
Sorry for the delay in merging this. I got distracted by other stuff.

#284 now exists for the "Orphaned script UI" issue.

Also, we will need to update the manual to document these changes. Is this something that you would be up to?

firstmiddlelast:
Didn't think of this, but sure - should I fork, branch, open a PR on the "Manual" repo?

Sailsman63:
No, we're no longer using a separate repo for the manual. The manual source is under docs/manual in the main AOI repo,  so just a PR against the main repo. We are currently using www.sphinx-doc.org to build the manual. If you have questions, let me know.

PR #285, opened 2022-05-08
Declarative Tools initialization via Annotations

makiam:
Replacement for #25


Sailsman63:
Please do not close an existing PR when rebasing your changes. Instead, use a `git push -f` over the existing branch, so that we can keep all the discussion in one place.

PR #286, opened 2022-05-08
Editing Tools test

makiam:
Tests for #285 PR


Sailsman63:
We've discussed testing the output of our localization files before. That is not something that should be put in unit tests.

`Tests/src/artofillusion/tool/help/TestToolsHelp.java` already exists in #285 / #25, and should be discussed there.

PR #287, opened 2022-05-16
Extact UI from registry

makiam:
Plugin registry now not depends of UI at startup. All plugin errors collected and passed to ArtOfIllusion and show error list only one time


ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -41,19 +44,19 @@ Changes copyright (C) 2018-2022 by Maksim Khramov
    * contained in them.
    */
 
-  public static void scanPlugins()
```
Sailsman63, line 44: You're changing a public method signature here without warning. What is the expected deprecation/upgrade pathway like?

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -65,50 +68,26 @@ public static void scanPlugins()
...
-        // Not a zip file.
-      }
-      catch (Exception ex)
-      {
-        System.err.println("*** Exception loading plugin classloader");
-        ex.printStackTrace(System.err);
-      }
-    }
-    processPlugins(jars);
-  }
```
Sailsman63, line 100: Another public method. Where did this functionality get moved to?

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -65,50 +68,26 @@ public static void scanPlugins()
...
-
-  private static void processPlugins(HashSet<JarInfo> jars)
```
Sailsman63, line 102: This, again more like the first.

ArtOfIllusion/src/artofillusion/PluginRegistry.java
```diff
@@ -149,7 +131,7 @@ private static void processPlugins(HashSet<JarInfo> jars)
    * @param nameMap   maps plugin names to JarInfo objects
    */
 
-  private static void processJar(JarInfo jar, Map<String, JarInfo> nameMap)
```
Sailsman63, line 152: And again a changed signature.

Sailsman63 (changes requested):

ArtOfIllusion/src/artofillusion/PluginRegistry.java
makiam: Only use for this method is to start plugins scan at application start. I did not find any other usages for this so I think it is safe enough to change. Moreover if someone want to call this method it is safe to ignore return value

ArtOfIllusion/src/artofillusion/PluginRegistry.java
makiam: Just removed. Did not find any usage for it

PR #288, opened 2022-06-04
Fix AIOB in ExecuteScript Window

makiam:
Fix IndexOutOfBounds exception when actual newRecentFiles list size less than hardcoded 10 value. Also removed unnesessary convert and cast to String array


Sailsman63:
Good Catch! I probably didn't notice because I have 10 items in that list already.

PR #289, opened 2022-08-18
Migrate class mappings to resource

makiam:


Sailsman63:
Okay, I see how this can work. The question is: Why? Do you forsee the need to be able to update the class translations without recompiling the application? Do you see a need to load different translation maps depending on configuration?

makiam:
From my POV this just looks clearer not to store translations in code. On the other hand I not shure this translations need at all. As I understand this created years and years ago for compatiility purpose. I think no such old files still exist.

peastman:
I think it's fine either way.  I can see the argument that moving it to a data file is cleaner.

Sailsman63:
I'm kind of coming around to the idea that this is cleaner. Not sure that I see any reasons *not* to separate it.

On the subject of removing the translations entirely: That is a breaking change, even if we think it's unlikely to be in use. Something to consider on the leadup to 4.0.

Sailsman63:
Okay, I'm merging this one. Thanks for the contribution!

Issue #283, opened 2022-04-20
Text Area on script editing windows does not resize on maximize/restore.


Sailsman63:
Maximizing or restoring a Script editing window leaves the RSyntaxTextArea set to the previous size. Clicking in or attempting to scroll the textarea triggers a re-layout, so it's still useable.

Manually resizing the window by dragging correctly triggers a re-layout.

PR #291, opened 2022-12-31
unify styling of comment types

Sailsman63:
Make sure that documentation-style comments are styled like other comments. I think these just got missed originally.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/keystroke/KeystrokeEditor.java
+++ b/ArtOfIllusion/src/artofillusion/keystroke/KeystrokeEditor.java
@@ -73,22 +73,13 @@ public class KeystrokeEditor extends BDialog
     languageChoice = new BComboBox(ScriptRunner.getLanguageNames());
     languageChoice.setSelectedValue(record.getLanguage());
 
-    syntaxTextArea = new RSyntaxTextArea(record.getScript(), 25, 100);
-    try
-    {
-      Theme theme = Theme.load(getClass().getResourceAsStream("/scriptEditorTheme.xml"));
-      theme.apply(syntaxTextArea);
-    } catch (Exception e) //shouldn't happen unless we are pointing at a non-existant file
-    {
-      e.printStackTrace();
-    }
+    syntaxTextArea = ScriptEditingWidget.getScriptWidget(record.getScript());
+
 
     syntaxTextArea.setSyntaxEditingStyle(record.getLanguage()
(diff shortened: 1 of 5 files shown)


peteihis:
That should do it, though I probably would have called the 'newStyle' something bit more unique. :)

Sailsman63:
Got the paren-matching to highlight both ends of a pair, and toned down the outline. Hopefully a little easier on the eyes.

> probably would have called the 'newStyle' something bit more unique. :)

What I *should* do is remove all of the code altogether and use an XML theme, as supported by the library. Puts all of the details in one place, instead of duplications.

peteihis:
> Hopefully a little easier on the eyes.

Definitely! -- Quite nice actually. 

Now, being as subtle as it is, it makes me notice, what I have been thinking all along that the colors in general could take serious taming down but that's a different project _(--> candidade for the xml-theme, I guess)_.

Sailsman63:
This was still itching at me a bit, so I went ahead and transitioned to the XML theme. Tweaking the colors should be simple from here, if you have thoughts.

I do want to de-duplicate the loading code, but we need a better way to communicate where configuration resources can be found.

makiam:
Missed now tabSize setup for all editors and also missed codeFolding enablement for KeyStroke script editor

Sailsman63:
Yes, I did. Fixed. Thanks!

Sailsman63:
And... I couldn't just leave it alone. I ended up extracting the common functionality out into a separate widget. Probably more to be done here, built on top of it.

I wasn't able to do quite what I anticipated. This widget is a little odd by buoy standards - the RSTA components are in a tight little stack that doesn't compose the way I'd normally approach things. I particular, I don't think that it's possible to remove the awt/swing style `KeyListener` from `artofillusion.script.ExecuteScriptWindow` without re-implementing it somewhere else.

Sailsman63:
And one last tweak. I got the RSTA properly wrapped in Buoy components, which allows to move to the buoy event mechanism. As a bonus, this fixes the layout bug #283 - I suspect that the old way just wasn't triggering a re-layout correctly, while the more specialized buoy `BTextArea` has this already properly daisy-chained in.

Sailsman63:
Final call for feedback on the code changes!

- Suggestions for alternative color schemes should probably be talked out separately.
- Discussions about other UI/UX concerns wrt the script editors and menus should also have their own discussions

PR #293, opened 2023-01-22
Update language selection renderer

makiam:
Update renderer to properly render list items when hovered with mouse

### DIFF
--- a/ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
+++ b/ArtOfIllusion/src/artofillusion/script/ExecuteScriptWindow.java
@@ -1,5 +1,6 @@
 /* Copyright (C) 2002-2013 by Peter Eastman
    Changes Copyright (C) 2023 by Lucas Stanek
+   Changes Copyright (C) 2023 by Maksim Khramov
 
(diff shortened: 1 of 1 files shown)


Sailsman63:
What is the intent here? Just a cleaner implementation, or something more?

makiam:
Before this fix
изображение
And after
изображение

Sailsman63:
I can't duplicate that issue so far. What java? I'm assuming you're running windows...

peteihis:
>I can't duplicate that issue so

I pointed that blanking text out already before. Windows & Java 11.

makiam:
See this with Windows 10 and Java 11 and Java 18
Not reproduced on Linux Mint

Sailsman63:
@peteihis I must have missed that while sorting through the comments in that post. Can you confirm that this change fixes it on your machine? I still can't duplicate the issue on linux, but this change does still run normally (As well as looking neater!)

makiam:
This change fixes rendering issue on my Windows desktop as I shows on screenshots above. And yes this rendering glitch is windows specific and not reproduced on Linux

peteihis:
> Can you confirm that this change fixes it on your machine?

Confirmed.

PR #296, opened 2023-01-23
Fix invalid shortcut letter

makiam:


Sailsman63:
Is this "Invalid" in the sense that it doesn't function? Russian speakers tend to use letters available across Latin/english character sets for their shortcuts? Other?

makiam:
изображение

makiam:
>> Is there a difference between a Cyrillic A and a latin A when typed?
What kind of difference do You mean?

peteihis:
> Is there a difference between a Cyrillic A and a Latin A when typed?

They are on a different key.... But really it can't be right that you have Latin short cut letters on the UI when you use a Cyrillic/Russian keyboard. I have no idea what the fix should be though.

OOT but usually the fonts also look different. Some times I have seen instruction/warning tags written in character sets where they did not bother to change all the letters but only the ones that were too different read ... And the outcome looks exactly like they didn't bother ... (but in those cases, though the translations were understandable, they were done ... should I say, in a somewhat 'creative' manner... I guess, sometimes you just have to use, what you got.)

makiam:
Well. On my keyboard russan A is on same key with latin F
And as I see ex for Mozilla Firefox and/or Chrome even they has localized UI the keyboard shortcuts is still latinic. And it works ok.

Sailsman63:
>On my keyboard russan A is on same key with latin F

Does that mean that you need to use a modifier code to generate an "А"?

peteihis:
The standard way to switch between character sets is to press Alt-Shift. That is the case with Scandinavian keyboard as well where things like `;:'[` give way to öäå and `<>\/[];'` will be relocated to give way to the first set... If I'm not mistaken Ctrl+key responds to what key is pressed, not what local character is routed to it. Sometimes it a bit confusing if you accidentally toggle to the other character set, as it only shows as a three letter text on the corner of the taskbar, where you never look.

I don't know what is the most common practice, but I suspect that keyboards for non-Latin character sets would most often come with hybrid prints.... (Think about writing code otherwise.)

RusHyb

makiam:
>> Does that mean that you need to use a modifier code to generate an "А"?
Nope. Once I switch input language to Russian I print in my language but when I press ex Ctrl-F this shortcut works ok.

makiam:
BTW. Localized Maxon Cinema4d UI. The menu items are in russian but shortcuts are latinic
изображение

peteihis:
Just playing with the thought but on a _fully_ Russian keyboard "Contrlol + A" would have to be "Контроль + Ф" ... 
agh ... image

Sailsman63:
Just making sure that we're defaulting to the common case: Is Latin/Cyrillic swapping of keyboards the norm for Russian speakers, or are more people most likely using Cyrillic exclusively?

makiam:
I think that most of us uses swapping because we still need english to enter URL's, emails, logins etc...

makiam:
>> Just playing with the thought but on a fully Russian keyboard
You yet not see the pink keyboard for blond....
https://www.sven.fi/ru/catalog/keyboard/blonde.htm

Sailsman63:
>Once I switch input language to Russian I print in my language but when I press ex Ctrl-F this shortcut works ok.

Is this with the shortcut defined with a Cyrillic A as in the current head, or with a Latin A, as you've suggested with this PR?

peteihis:
> You yet not see the pink keyboard for blond....

Hehe.. That pink one came up but I did not look at it that closely. "The biggest button" for Enter, it seems ... :D

Sailsman63:
Okay, you've got me convinced that this is probably the best quick way forward. Thanks for bringing it up.

PR #297, opened 2023-01-25
Compiler warnings cleanup

makiam:
Clean a lot of compiler warnings against raw-type usages and, unnessesary cast calls and use diamond operator

### DIFF
--- a/ArtOfIllusion/src/artofillusion/CameraFilterDialog.java
+++ b/ArtOfIllusion/src/artofillusion/CameraFilterDialog.java
@@ -1,5 +1,5 @@
 /* Copyright (C) 2003-2009 by Peter Eastman
-   Changes copyright (C) 2022 by Maksim Khramov
+   Changes copyright (C) 2022-2023 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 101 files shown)


Sailsman63:
These should be grouped by warning type.

makiam:
This compiler warnings cleanup are tied thogether... I change raw type usage, then clean typecasts which becomes useless for same usage and updates constructor calls with diamond operator.
So I see no reason to split this on multiple requests

Same comment for Polymesh and SPManager

ArtOfIllusion/src/artofillusion/CameraFilterDialog.java
```diff
@@ -59,14 +59,7 @@ public CameraFilterDialog(WindowWidget parent, Scene scene, SceneCamera camera,
...
-    filtersPanel = new FiltersPanel(theCamera, new Runnable()
-    {
-      @Override
-      public void run()
-      {
-        applyFilters();
-      }
-    });
+    filtersPanel = new FiltersPanel(theCamera, this::applyFilters);
     BorderContainer previewPanel = new BorderContainer();
```
Sailsman63, line 63: This change isn't related to compiler warning cleanups. While I do prefer making this cleanup, it belongs in a separate commit.

Sailsman63:
The thing is, I can't even tell what compiler warnings you are actually working with here. What version of the JDK generates these warnings? What `Xlint` should I pass to see the whole list?

I also see that these changes break compatibility with JDK 8.

makiam:
I used -Xlint:unchecked -Xlint:rawtypes -Xlint:cast compiler args and also IDEA inspection "Explicit type can be replaced with <>"

makiam:
Daimond inference in anonimous inner classes started to support from Java 9. This is my fault. But this may be reason to switch to Java 9

Sailsman63:
> So I see no reason to split this on multiple requests

I'm not asking for separate requests, just separate commits.

>  I change raw type usage, then clean typecasts which becomes useless for same usage and updates constructor calls with diamond operator.

Each of these steps should have been a separate commit, even if they end up covering the exact same lines.

> -Xlint:unchecked -Xlint:rawtypes -Xlint:cast compiler args and also IDEA inspection "Explicit type can be replaced with <>"

And this is hard to review because I can't tell what changes were due to "Hey, this unchecked operation could break things!" and which are "The IDE thought it would be cleaner."

Sailsman63:
While I do appreciate contributions to clean up code in the app, in its current form this PR would take too much effort to review in relation to the benefit.

PR #298, opened 2023-02-08
Extract scene explorer

makiam:
Extract all scene-tree only related code to separate class


Sailsman63:
Do you see any use-case for a scene explorer to be used outside of `LayoutWindow`? If not, it should probably remain a member of `LayoutWindow`.

makiam:
Layout uses also Score and PropertiesPane objects which used in layout only. And they separated to classes. But scene tree code is still mixed with layout

PR #299, opened 2023-02-10
Drop JDK8 code 

makiam:
Drop JDK8 code from OSX plugin

### DIFF
--- a/OSSpecific/src/artofillusion/osspecific/MacOSPlugin.java
+++ b/OSSpecific/src/artofillusion/osspecific/MacOSPlugin.java
@@ -1,5 +1,5 @@
 /* Copyright (C) 2002-2020 by Peter Eastman
-   Changes copyright (C) 2017 by Maksim Khramov
+   Changes copyright (C) 2017-2023 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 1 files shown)


Sailsman63:
We'd need community consensus before we can drop support for JDK 8. Historically, AOI tends to be conservative in what versions of java we require because part of the userbase tends to run systems that cannot run updated versions of java.

peastman:
What are the benefits of dropping support?

makiam:
it seemed to me that we switched build version to JDK9. 
So we may left this PR for future

Sailsman63:
>it seemed to me that we switched build version to JDK9.

What gave that impression? (I'm truly curious. Did we accidentally break back-compatibility in some way?)

makiam:
it's weird for myself ))

Sailsman63:
In general, JDK 8 is still getting security updates. I think that we need some stronger reasons to update our minimum supported version, such as new features that would really positively impact the application. If you are aware of anything along those lines, please open up an issue to make a case.

Sailsman63:
I did figure out where the idea that we've moved to jdk9 came from:

The mac .dmg package *is* java 9+. this is due to needing to have different launcher configurations for jdk8 and jdk9. However, mac users can still run on java 8 if they use the generic/non-install package, so we want to keep this code for now.

PR #300, opened 2023-03-02
Move preview UI code to Procedure Editor

makiam:
Move preview UI code to Procedure Editor where it only used. Preview window now follow procedure editor

### DIFF
--- a/ArtOfIllusion/src/artofillusion/procedural/ProcedureOwner.java
+++ b/ArtOfIllusion/src/artofillusion/procedural/ProcedureOwner.java
@@ -1,5 +1,5 @@
 /* Copyright (C) 2002 by Peter Eastman
-   Changes copyright (C) 2020 by Maksim Khramov
+   Changes copyright (C) 2020-2023 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -10,6 +10,8 @@
    PARTICULAR PURPOSE.  See the GNU General Public License for more details. */
 
 package artofillusion.procedural;
+
+import artofillusion.MaterialPreviewer;
 
 /** This interface defines an object which "owns" a procedure, and helps
     define the user interface for editing it. */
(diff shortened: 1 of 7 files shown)


ArtOfIllusion/src/artofillusion/procedural/ProcedureOwner.java
```diff
@@ -22,20 +24,14 @@ define the user interface for editing it. */
   
   /** Create an object which displays a preview of the procedure. */
   
-  default Object getPreview(ProcedureEditor editor)
+  default MaterialPreviewer getPreview()
```
Sailsman63, line 27: Changing the signature here technically breaks compatibility at both a source and binary level. While updating the core plugins is simple, have you tested that this does not break TaPD, which builds of the procedural system?

makiam:
TAPD plugin use its own modules system not tied to aoi

вт, 7 мар. 2023 г., 10:50 Sailsman63 ***@***.***>:

peteihis:
To me this and #301 seem like very wrong things to do. 

The thinking very much through the entire AoI has been that each object should do itself everything that is related to it and I think there is a good reason behind it. That of course leads to some level of redundancy in the code but what is going to happen when somebody add something new that needs an entirely different logic?

Let's not start breakin the basic principles that the software has been built on.

makiam:
>>  the entire AoI has been that each object should do itself everything that is related to it
Totally agree with this. And this way the ProceduralXXX objects should NOT create UI parts for ProceduralEditor and should NOT control disposal of UI (via disposePreview).

Sailsman63:
I don't think that this one has the same sort of issues as #301. It does, however, rest on two assumptions:

1. All procedure editors will want their preview windows to be laid out in the same way.
2. All types of procedures will be able to be effectively previewed through a `MaterialPreviewer` Widget.

The first I am unsure about. The second I have more grave suspicions: better, I think, to accept a generic Widget here.

makiam:
Once ProcedureEditor changes some in it's editing object it triggers ProcedureOwner updatePreview(previewer)  anв inside it usually calls MaterialPreviewer render(). So if pass previewer as Widget we need extra cast and we have no confidence that proper object passed to update

PR #301, opened 2023-03-02
Move procedural module background render code to Procedure Editor

makiam:


makiam:
>> What prompted the change to the second in this case?
There are not clear second case nor first case but combined approach. The module know how to draw itself and it is remains unchanged. But ProcedureEditor controls modules selected/unselected state and know how to draw this state and draw common module background (which is always the same for all modules)

Also I want to try to rework this editor and made rendering colors customizable. This PR is for it.

Sailsman63:
This introduces inconsistency on how the UI is modeled, though. AOI tends toward "Object knows how to represent *itself* visually." (And, indeed, you're still allowing the `IOPort`s to draw themselves, which means you're hopping between architectural models in very closely related code.)

> Also I want to try to rework this editor and made rendering colors customizable. This PR is for it.

Why would that goal require this change?

Sailsman63:
I think you misunderstood me. Please take a step back and explain why this change is needed to meet your overall goal.

Sailsman63:
Without understanding why this change is necessary, I can't properly evaluate or review. Closing for now.

Issue #303, opened 2023-05-20
AOI hangs on shutdown if SPManager has been opened.


Sailsman63:
This is most noticeable if AOI is launched from the command line. The editing windows will close, but the command prompt never clears properly. This dates to https://github.com/ArtOfIllusion/ArtOfIllusion/pull/123/commits/1b62191092a2a387cf4843cccb0bc058e5616f10, which was merged as part of #123.

It's unclear if something is wrong in SPManager that this has brought to the surface, or if the main error is in the new shutdown hooks.

Tested on Linux.

PR #304, opened 2023-05-20
Issue #303 Fix

makiam:
SPManager prevents Swing autoshutdown because it's window remains hidden on close not disposed.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
+++ b/ArtOfIllusion/src/artofillusion/ArtOfIllusion.java
@@ -12,24 +12,22 @@
 
 package artofillusion;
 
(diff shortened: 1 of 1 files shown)


Sailsman63:
Thanks for digging in so fast!

Has this been tested with OSX? It looks okay, but I'd like to be sure that this doesn't break the app tray integration.

makiam:
I cannot check this on OSX now

Sailsman63:
@peastman: I think you might be the only one with a mac at the moment. Can you check if a build of this commit allows the mac menubar integration to continue working? In particular, does closing the last `LayoutWindow` leave things running in background the way it is supposed to?

PR #307, opened 2023-08-28
Grid timing

peteihis:
Fix to the disappearing grids issue mentioned [here](https://sourceforge.net/p/aoi/discussion/47783/thread/db8076ee17/). 

One way to could have been been to draw the grid always, when it is on. In many cases it would not hurt, but if there is a large move or a strong scaling happening, the flashing can be rather disturbing.

There were a couple of minor comment typo fixes in the last commit included, I hope those are not too disturbing. I could not just scroll past them when they hit my eye. :)

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -469,8 +469,8 @@ public abstract class ViewerCanvas extends CustomWidget
 
   /**
     ViewAnimation calls this when the animation is finished, so the orientation menu
-    will be up to date and the perspective is set to it's value without launched
-    launcing a new animation sequence.
+    will be up to date and the perspective is set to it's value without
+    launching a new animation sequence.
   */
 
   public void finishAnimation(int which, boolean persp, int navi)
@@ -623,7 +623,7 @@ public abstract class ViewerCanvas extends CustomWidget
     Set the distance from drawing plane to view camera. <p>
 
     Depending on the view action this may be used to recalculate
-    the positon of the camera or the rotationCenter or it may be
+    the position of the camera or the rotationCenter. Conversely, it may be
     recalculated from the rotationCenter and camera position.
    */
   public void setDistToPlane(double dist)
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -1211,7 +1211,7 @@ public abstract class ViewerCanvas extends CustomWidget
 
     // Draw the grid, if necessary.
 
-    if (showGrid)
+    if (showGrid && !animation.animatingMove() && !animation.changingPerspective())
     {
       float scale1 = 0.75f/255.0f;
       float scale2 = 0.25f/255.0f;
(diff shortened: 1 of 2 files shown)


Sailsman63:
This is a fairly simple PR! Shouldn't take long to review.

> if there is a large move or a strong scaling happening, the flashing can be rather disturbing.

Can you be a little clearer here? What flashing? I've been playing around with the camera moves, and have not had any visual issues with the grid being visible during view animations. Indeed, I found the disappearance of the grid disconcerting until I lengthened the max animation duration.

> There were a couple of minor comment typo fixes in the last commit included, I hope those are not too disturbing. I could not just scroll past them when they hit my eye. :)

I understand the impulse. This is were `git add -p` is your friend. You can stage just the typo fixes and then `git commit --amend` to add them to the previous commit. (Or just `git commit` them as their own thing.) Small enough in this case that I probably won't worry about it.

peteihis:
> What flashing?

Excatly why the grid is _not_ drawn during animation. :D

If you have zoomed close, the grid appears big. Then you decide to autosoom to another object that is on the other side of the scene or very different in size. -- Now if the grid was drawn you would not get the impression of continuous zooming/moving of the grid but rather random grid lines drawn on every frame of the animation. That flashing _(which is not happening now, because the gris is drawn the next time only after the animation stops)_.

Sailsman63:
> if the grid was drawn you would not get the impression of continuous zooming/moving of the grid but rather random grid lines drawn on every frame of the animation. 

How big/far are you looking? I've tested w/objects at 0,0,0 and 2500,1500,1000 and actually prefer the gridlines moving by.

peteihis:
Oookay... I should not have mentioned it and you would not have noticed....  It's been like that as long as the animation engine has been there.

> How big/far are you looking?  ... actually prefer the gridlines moving by.

Not necessarily that big or far and most of the time, so would probably I. But at it's worst it actually becomes something like the "movie effect" on car wheels that look like rotating backwards ... or then just random crosshairs flashing on the screen. And then the change from orthographic to perspective view is one thing where the grid "just snaps" into the different mode. I rather have the view do it's preparations first and draw the grid, when "the word is ready for it".

Hard to do a screen capture of it for a variety of reasons _(capture rate vs. refresh rate, the exact dimensions an positions of the objects, etc. ...)_ but here's one. On the last move the grid is not nice at all.

Eventually it might be possible to determine some rules on, when the grid is worth drawing during animation and when not but that would take quite a bit of experimenting. Automatic grid size would help on that too... _(On my part I tested that idea many years ago but that never went anywhere. Basically a size series ... 0.2, 0.5, 1.0, 2.0, 5.0, 10 , 20...  The smaller size fading in as you zoom in and vice versa.)

https://github.com/ArtOfIllusion/ArtOfIllusion/assets/22843957/ba6d13a7-a6ae-4980-909b-36a5bc512cf7

Sailsman63:
Okay, I see what you're saying and understand why the animations are set the way they are.

 I agree that auto-scaling grids would actually be a better mitigation, and maybe we can revisit this if they ever become an actual thing. Planning to rebase (The typo fixes in the bugfix commit are bugging me more than I anticipated) and merge shortly.

ArtOfIllusion/src/artofillusion/ViewerCanvas.java
```diff
@@ -470,7 +470,7 @@ public void preparePerspectiveAnimation()
   /**
     ViewAnimation calls this when the animation is finished, so the orientation menu
     will be up to date and the perspective is set to it's value without launched
```
makiam, line 472: launched at this line and launching again on next

Sailsman63: I'll catch it in the rebase.

Sailsman63:
Thanks for the contribution!

PR #309, opened 2023-10-01
Update JOGL libraries

makiam:


makiam:
No more x86 libraries but added linux-arm

PR #322, opened 2024-02-18
Primitive providers

makiam:


peteihis:
Does this make any sense?

Creating a primitive is just one line in the code. Why would you build a factory of their own for for every single one of them?

I admit though,  that `createObjectCommand()` may sound a bit of a mouthful...

peteihis:
`BMenu fileMenu, recentFilesMenu, editMenu, objectMenu, toolsMenu, viewMenu, scriptMenu;`

Logically the order should be:

`BMenu fileMenu, recentFilesMenu, editMenu, objectMenu, toolsMenu, scriptMenu, viewMenu;`

So, first fileoperatios then object and scene and last visualization, not shuffled with view between the scene operation menus. And from that forward, things like "About", "Help"....

makiam:
IMO this is better from customization point of view. Ex someone can write and register script to add more items to layout without accessing LayoutWindow internals

Next step: add similar style plugin to adding Tracks to objects

Sailsman63:
I share @peteihis concerns here. What sort of types would you like to be able to add to Primitives?

peteihis:
> What sort of types would you like to be able to add to Primitives?

A bit off topic but actually lights, cameras and reference images are not primitives at all. They may be basic objects of AoI but not geometrical primitives. I'd give them sub menus of their own.

Then, on the other side, torus, pyramid etc... are, _(well, it depends on who you ask, what still is a primitive, but the common thing is that they are defined by relatively simple mathematical equations)_ but they don't exist in AoI as original mathematical primitives, just Triangle Meshes produced by a dedicated tool.

makiam:
Probably Plane and Torus may be added as first-class primitives

peteihis:
> Probably Plane and Torus may be added as first-class primitives

See my edit above.... And if they were added, then the ray tracer would need the corresponding RTObjects. 

The useful thing would be to create an implicit object editor, where you could build objects out of mathematics by more or less normal looking modeling tools. Basically by parametrized primitives and boolean operations. With the current implicit tools you just wire equations together with very limited visual feedback.

The use that I can see in a tool like this would be optically perfect surfaces _(AoI is all about visualization!)_. One of the current long term issues is that while smoothed meshes respond to camera rays as if the surface was curved, to light rays the surfaces still consist of flat triangles. That shows up in reflection patterns.

PR #328, opened 2024-06-06
Ray aim

### LINKED ISSUE #239: Raytracer anti-aliasig aim seems a bit off
Hi! 

I have been wondering for a long time why the a raytraced image travels so clearly to left and up in the anti-aliasing phase. I thought that this would be because the first phase image was already created with anti-aliasing rays that are not aimed to the center of each pixel but the shift is too large for that: It is not half a pixel, horizontal and vertical, as you'd expect but more like a pixel and a half.

aa jump

Because of the shift on an image that is rendered with low ray count (4/4 below) edges that lean right look a kind of shredded.

aa quality

Then I noticed that magnification of an anti-aliased image is different from an non anti-aliased one. Below red is from the initial phase of an anti-aliased image and cyan from a non anti-aliased image.

aa size

The rays on the aa-image are aimed to a slightly narrower FOV than on the non-aa one. I believe the non-aa to be correct as there each ray is simply aimed to the center of the corresponding pixel. As the mismatch seems to be related to pixel count this would also cause a small distortion in the aspect ratio of the image, when the image is not a perfect square.

I have not studied the code very deep but for anti-aliased images the working image size is `w×2+2, h×2+2` as if the purpose was to add a temporary line of pixels to each side of the image. I assume this to be for the maximum level of anti-aliasing and I suspect that the issue is related to that. 

I also had images to compare the different anti-aliasing selections, but I could not really perceive a magnitude difference between medium and maximum anti-aliasing.

Any hints? Or should I just dig in? -- The logic in the rr-aa is a bit mazy to follow, but then it just takes the time.

peteihis:
Resolves #239

There were two problems:
- The aim of the 1st set of rays in anti-aliasing
- The perceivable difference in FoV, when comparing a single-ray render to an anti-aliased one. 

It turned out that the pixel shift issue was actually a combination of two different miscalculations, which threw me off course in 2020: 

1. The firts set of rays was shot too much right and up
2. For the latter rays the positions _(= affected sub-pixels)_ of the first set were calculated wrong. 

The picture below presents what happened in the rendering space of a 2×2 pixel image. -- So there are 6×6 sub-pixels and 4/4 anti-aliasing, randomization ignored.

Aim

This is how the fix affects rendering _(FoV fix included)_ : 

Before-After

The FoV issue was just the mistake of using the size of the rendering space as the reference size of the image. The rendering space _(that never exists as such but as a calculation method)_ is larger than the actual image.

These before and after images compare a single-ray render to an anti-aliased one. The object looks like changing size in the Before-case. Really it is the FoV that is changing.

FoV before 
FoV after

### DIFF
--- a/Renderers/src/artofillusion/raytracer/RaytracerRenderer.java
+++ b/Renderers/src/artofillusion/raytracer/RaytracerRenderer.java
@@ -1244,6 +1244,7 @@ public class RaytracerRenderer implements Renderer, Runnable
     Vec3 orig = ray.getOrigin(), dir = ray.getDirection();
     double h = i-rtWidth*0.5+0.5, v = j-rtHeight*0.5+0.5;
     Random random = workspace.context.random;
+    int imgHeight = height;
 
     if (antialiasLevel > 0)
     {
@@ -1254,6 +1255,13 @@ public class RaytracerRenderer implements Renderer, Runnable
       int col = num-row*cols;
       h += (col+random.nextDouble())/cols-0.5;
       v += (row+random.nextDouble())/rows-0.5;
+
+      // To aim the rays correctly we need to use the height of 
+      // the actual image area, not the rendering area with the 
+      // extra antialiasing pixels around the edges. The value is 
+      // the rendering time height in pixels.
+
+      imgHeight = height*2;
     }
     double dof1 = 0.0, dof2 = 0.0;
     if (depth)
@@ -1261,7 +1269,7 @@ public class RaytracerRenderer implements Renderer, Runnable
       dof1 = 0.25*(random.nextDouble()+distrib1[number&15]);
       dof2 = 0.25*(random.nextDouble()+distrib2[number&15]);
     }
-    sceneCamera.getRayFromCamera(h/rtHeight, v/rtHeight, dof1, dof2, orig, dir);
+    sceneCamera.getRayFromCamera(h/imgHeight, v/imgHeight, dof1, dof2, orig, dir);
     theCamera.getCameraCoordinates().fromLocal().transform(orig);
     theCamera.getCameraCoordinates().fromLocal().transformDirection(dir);
     ray.newID();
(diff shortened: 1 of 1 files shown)


Sailsman63:
Please do not merge `master` *into* working branches like this. If there are conflicts, rebaseing is the correct way to resolve them. If there are no conflicts, we can just merge the working branch into master. I'll try to take a look at this one later today.

peteihis:
**I did not!**

_A couple of edits... :_

The verified tag says "This commit was created on GitHub.com and signed with GitHub’s verified signature. "I had nothing to do with that. At the time on the stamp I was getting ready to go to work.

BUT this PR was missing for a while and there was a note from GitHib that they were recovering some branches that had gone missing. That obviously was a larger scale incident. The note bar was on the www-page for several days and  still is on some views.

... Even the branch was hidden from the normal view, but it was under "All" and it showed that there was a PR, but inaccessible. I probably pressed "Sync Fork" at some point but that should be a pretty normal thing to do and it should not appear as merge the master into the branch.

peteihis:
OK. I rebased the local branch and force pushed. At least the illegal operation is gone now.

Sailsman63:
No worries, I wouldn't have expected "Sync Fork" to default to a merge either. However, when I read github's [documentation of the feature](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/syncing-a-fork?platform=windows) I see that their instructions on "how to do this on the cli" *do* have you merge the upstream into the branch, rather than doing a rebase (which is how I would have implemented it.)

The "missing" PRs have been a bit annoying - apparently, something went wrong with their search index, and this has been happening to every project, though you should be able to get to an unlisted PR if you know its number. I'm not impressed that they are using a general search index for something like this.

peteihis:
Anyway.... I discovered something more:

Lesnd-much off
<br>

That is in mid anti-aliasing an image with heavy DoF blur. The first set of rays seems to come from about a "lens diameter much" different spot than the final rays. The finals look correct to me. I'll have to come back with a different PR to fix that. Probably not a big mistake but those are the hardest to detect.

Sailsman63:
This was a subtle one (or two). Thanks for the fixes.

PR #329, opened 2024-06-11
Comma as alternate decimal separator

### LINKED ISSUE #129: Alternate decimal separator to input fields
Hi.

Generally in the computer world _point_ (period, dot, full-stop) is the decimal separator and basically text turns into a number correctly by default settings, when there is a point.  However in the parts of the world I come from, the standard decimal separator is comma and naturally the number pad of the keyboard also types comma...

With AoI this is a bit of an annoyance as the user will have to get the decimal point from the typing keyboard, when otherwise typing numbers on the number pad.

The typical localization of course is, that the decimal separator is then comma and nothing else. For example Windows calculator does not "see" a point and if I copy and paste 124.76 to the number display, it becomes 12 476. _(If I download lists like star maps, they usually come as text, with decimal points, and sometimes the software that I bring it into does localization right and sometimes not....)_ 

I don't know, what would be available for keyboard localization, but at least Dassault Sytems and AutoDesk use a system that recognizes both . and , as a decimal separators and you can actually use both even inside the same calculation.

Now the profound difference is, that value fields in professional CADs are not number fields but equation fields, where there is an interpreter constantly reading, what the user is typing -- the text turns red if there is an error etc... The equations are stored as equations, with some standardizing happening when the content is reproduced from the data, but you can happily mix imperial and SI units if you like and use dimensions from what you have done before.

AoI does not have any dimension controlled features, so the equation engine would not be very useful but, what about the point vs. comma issue? Are there any "hidden" possibilities to have the input fields understand, that if I type comma from the number pad, I mean decimal and if I copy & paste numbers with a point to the fields, I still mean decimal? -- And all the other variations. IMO one clear solution would be to read and display a comma as a point instantly, but I don't know if something useful might be available?

peteihis:
A fix to a 20 year annoyance ... : ) Resolves #129

`ValueField` and `ValueSlider` can now handle comma as decimal separator. I'd expect this to be a very welcome tweak to most European users at least.

The function is that, while the field is being edited, it treats a single comma as a decimal separator just like a point. When the field loses focus, the comma is replaced by a point. The traditional computing- and American style of using point for the purpose is not affected.

If there are any other numbers fields I missed, let me know. _(Occasionally I have used software where some fields can handle comma and some not, which is again one of those quality things I don't like to leave behind.)_

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ui/ValueField.java
+++ b/ArtOfIllusion/src/artofillusion/ui/ValueField.java
@@ -75,16 +75,16 @@ public class ValueField extends BTextField
     double val = value;
 
     try
-      {
-        if ((constraints & INTEGER) != 0)
-          val = (double) Integer.parseInt(getText());
-        else
-          val = new Double(getText());
-      }
+    {
+      if ((constraints & INTEGER) != 0)
+        val = (double) Integer.parseInt(getText());
+      else
+        val = new Double(getText());
+    }
     catch (NumberFormatException ex)
-      {
-        return false;
-      }
+    {
+      return false;
+    }
     return isValid(val);
   }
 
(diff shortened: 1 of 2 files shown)


makiam:
May this validation to be implemented via https://docs.oracle.com/javase/1.5.0/docs/api/javax/swing/InputVerifier.html
At least for ValueField?

peteihis:
> May this validation to be implemented via ... 

Sure. 

This has been pretty much a standard approach in engineering CAD tools for many years so basically there's nothing new in it. Though, I suspect that a an enable/disable toggle for accepting comma as decimal separator might be requested in some cases.

There would be some alternative ways to implement it, like substituting the comma with a point immediately, when you type it _(which I found somewhat blunt)_ or that the comma on the field text is left as comma and the next time you open the field, the field text is generated from the value _(which felt a kind of unfinished)_ -- The most friendly way seemed to be to read the field through a filter while editing and rewrite the text with a point when done.

Thankfully the programming language takes case of things like minus and exponent characters.

Sailsman63:
Thanks for the contribution! Sorry for the long wait on something that I know has been a pet difficultly of yours for a while.

I wonder if this could have been coded a little more compactly, but it's all private - we can revisit when there is nothing more urgent to tackle.

PR #332, opened 2024-06-24
Suggesting `getSceneCamera` method

peteihis:
Ever wondered how many times the code repeats,
~~~~Java
if (view.getBoundCamera() != null && view.getBoundCamera().getObject() instanceof SceneCamera ) 
   ((SceneCamera)view.getBoundCamera().getObject()).whatEverMethod();
~~~~
?

If Java were a spoken language, you'd probably say "It's a bit of a mouthful".

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
+++ b/ArtOfIllusion/src/artofillusion/ViewerCanvas.java
@@ -443,7 +443,7 @@ public abstract class ViewerCanvas extends CustomWidget
     }
 
     // SceneCamera has logic of it's own, so just jump there!
-    if (boundCamera != null && boundCamera.getObject() instanceof SceneCamera){
+    if (getSceneCamera() != null){
       perspective = perspectiveSwitch = nextPerspective;
       return;
     }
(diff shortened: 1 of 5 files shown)


ArtOfIllusion/src/artofillusion/ViewerCanvas.java
```diff
@@ -1968,7 +1986,7 @@ public void draw(ViewerCanvas viewToDrawOn)
       else
         tipPoint = rotPoint;
 
-      if (boundCamera != null && !(boundCamera.getObject() instanceof SceneCamera))
+      if (getSceneCamera() != null)
```
Sailsman63, line 1989: I think that you might not want to change this line. It's actually checking to make sure that the bound camera is *not* a `SceneCamera`. (Orthographic/flat view.)

Sailsman63 (commented):
I understand the urge to make the code more expressive. However, I'd rather not add a bunch of bespoke methods for things like this.

I'm also not sure that we should continue to perpetuate passing `null` around as a marker. In many cases, this would be cleaner with an `Optional`, though that would also mean shifting some logic to use the stream-ish `filter()` and `map()` methods and company.
     
                                     .if

PR #333, opened 2024-06-25
Concurrent rendering bug fix (preview setting overriding render setup)

### LINKED ISSUE #331: Camera view RENDERED mode get copied to RenderScene
If the view with `SceneCamera` bound to it, is set RENDERED, Raytracer rendeing of a final image ignores user settings on the dialog and renders the image in preview settings. The next time you open the dialog, it is in preview settings, Maximun anti aliasing, 4/4 rays and all check boxes off.

peteihis:
Resolves #331 

The idea is that if a view is using the same renderer that renders the image the rendermode of the view is temporarily changed for the image rendering.

I tried to have the rendermode of the rendered view change only just before the rendering starts, but so far it did not work. Mostly you notice the difference in animation rendering. The configuration information bounces between places a few time over, and I must have missed a turn there. Anyway the issue itself is gone.

This required editing a couple of constructors, but the old ones are still there, so nothing should be broken.

Force-push = a format mistake fix.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/RenderSetupDialog.java
+++ b/ArtOfIllusion/src/artofillusion/RenderSetupDialog.java
@@ -1,6 +1,6 @@
 /* Copyright (C) 1999-2015 by Peter Eastman
    Changes copyright (C) 2016 by Maksim Khramov
-   Changes copyright (C) 2020 by Petri Ihalainen
+   Changes copyright (C) 2020-2024 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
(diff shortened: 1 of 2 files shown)


Sailsman63:
I'm kind of frustrated that I missed this one - this has got to be annoying in use.

I'm trying to understand how we end up with the interactive renderer being listed as `currentRender` in the first place. I don't think that the settings ever go through the configuration system.

peteihis:
The way I handle it here is really more like a patch over a probem, not really fix to the root cause. 

Instead of switching the rendering mode of the view, the real fix would be to create a dedicated renderer for the rendered view. The original approach seems to be from the time when memory size was driving decision making, so only one instance of a renderer was created an that was used for both purposes.

> this has got to be annoying in use

Yes it is!

Sailsman63:
Thanks very much for both the report and the effort to find a fix. With #341 merged, I'm closing this one.

Issue #335, opened 2025-08-02
Camera Filters


peteihis:
As came up in this discussion: [https://sourceforge.net/p/aoi/discussion/47783/thread/e2c49a30e2/#e8c9](https://sourceforge.net/p/aoi/discussion/47783/thread/e2c49a30e2/#e8c9)

There is a number of issues with ImageFilters:
- Most of the filters do not have version numbers. The only reasonable way to develop them would be to deprecate the current one and write a new one with a different name.
- Blur filter tries to save itself with Integer value whereas the property is a double.
- Animation tracks of filters are very much a hidden feature as automatically generated subtracks toPose track.
- Opening the CameraFilterDialog results ALL filter animation keyframes to be lost. Does not matter how you exit the dialog.
- Undo does not cover the filter tracks.
- Programwise the filters, all but the abstract ImageFilter, were moved to a Plugins like the Renderes are. However, this transition was only partial as the filter tracks and the editor are still in core AoI.

I managed to sort all the usability issues (disappearing keyframes, ineffective undo, non saving Blur) out but the structure of the prototype is still weird. 

It looks like the the plan had originally been to send the SceneCamera to the CameraFilterDialog and have it do all the changes. So far so good. --- But then somebody noticed that camera parameters that are still only in the UI-widgets of the camera parameters dialog need to be sent to the preview camera of the filter editor, and then things took a bad turn: Instead of making the needed changes to the filter dialog constructor(s) they decided to make a duplicate of the Scene Camera to carry the new parameters to the filter dialog. What happened is that not just the camera was a duplicate but also the filters. What came back from the trip was the duplicate filters, which the filter tracks no longer recognized, and the system replaced the parentless tracks with new ones. Naturally in that process all the work that was done to the animation tracks was lost.

When I look at it, it comes to my mind that the filter tracks must be a later addition. All edit()-methods of 3D-object get as parameter a Runnable that is supposed to take care of the UI updates and undo functions. As it happens that undo does not include the automatically generated tracks. Also in the SceneCamera.edit() method after the CameraFilterDialog closes, it does not leave behind any hint, if it actually made changes or not! The same procedure, with reconnecting tracks with their parents is ran anyway, needed or not.

This is an area where there never seems to have been a very consistent plan on how to handle the entirety. It is as if the decision making had been fragmented into short moments in time and had every time been guided by entirely secondary constraints.

Well... I guess somebody had to dig it out, so why not me? -- I'm sketching a plan.

PR #337, opened 2025-08-03
Camera filters bug fixes

peteihis:
Bug fixes to #335

No new features or interface changes.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/CameraFilterDialog.java
+++ b/ArtOfIllusion/src/artofillusion/CameraFilterDialog.java
@@ -1,5 +1,6 @@
 /* Copyright (C) 2003-2009 by Peter Eastman
    Changes copyright (C) 2020, 2022 by Maksim Khramov
+   Changes copyright (C) 2025 by Petri Ihalainen
 
    This program is free software; you can redistribute it and/or modify it under the
    terms of the GNU General Public License as published by the Free Software
@@ -44,9 +45,12 @@ public class CameraFilterDialog extends BDialog implements RenderListener
   private static final int PREVIEW_WIDTH = 200;
   private static final int PREVIEW_HEIGHT = 150;
 
+  private boolean clickedOk;
+
   public CameraFilterDialog(WindowWidget parent, Scene scene, SceneCamera camera, CoordinateSystem cameraCoords)
   {
     super(parent, Translate.text("Filters"), true);
+    clickedOk = false;
     theCamera = camera;
     theScene = scene;
     this.cameraCoords = cameraCoords;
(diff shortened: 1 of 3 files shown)


ArtOfIllusion/src/artofillusion/CameraFilterDialog.java
```diff
@@ -44,9 +45,12 @@ public class CameraFilterDialog extends BDialog implements RenderListener
   private static final int PREVIEW_WIDTH = 200;
   private static final int PREVIEW_HEIGHT = 150;
 
+  public boolean okPressed;
```
Sailsman63, line 48: What is this boolean doing? I see that you explicitly set it on lines 53 and 142, but it's not *read* anywhere.

peteihis: The way it _should_ have been done it would probably be to have the exit of the dialog return a boolean, but then it would have changed the the interface... And the entire process is so weird with the interactions between the filters, track and the camera parameters that it would take a much more thorough redesign of the process to make it look any clearer. So better to keep it in minimum this time.

If you'll look at the file, not just the changes it should explain thing a bit.

Sailsman63:
For clarity, am I correct in thinking that these changes would prevent the type of mixed-up file that Konrad posted? If so, which specific change fixes that? I never fully understood the proximate cause (Was it the Blur filter outputting the wrong value? I thought he didn't have a blur filter in the first place.)

peteihis:
No. I could not reproduce that particular situation. 

He had a Brightness filter there, but somehow the animation track of it had not been saved. The missing track probably had never been created in the process of copying the SceneCamera from one scene to another. The reason to that seemed to have been somewhere else ... The scene just behaved strangely and stopped painting the score at all. Then after reopen some update corrected it. I could not reproduce that situation either, but the behavior was similar to some division by zero cases that I have seen before.

And then at file open it tried to read the filter track that was never written. PoseTrack.intiFromStream does not read the number of subtracks itself the subtrack array is already there and I suppose it assumes that if there is a filter there must be a track. I did not manage to track down where the subrtack array came from _(or did not feel like spending time on it)_.

What I did here was merely to make the tracks usable and fix the type bug in the Blur filter -- Strangely that bug seems to have been there since the filter was first created, but nobody ever pointed out that files can not be saved with a Blur filter present.

ArtOfIllusion/src/artofillusion/CameraFilterDialog.java
peteihis: I am not particularly happy with the structure either, but I'd rather keep this one just as a functional bug fix. It was a patch-over-patch but now it least now it works.

Reworking it will be a much bigger project and if we go into that it may never be finished --> users will never ge the working version. I also found a bunch of other things too that are wrong with the filters, but they don't prevet using the filter tracks. I'd hold those to the next round.

Sailsman63:
Thanks for the fixes! Finally got around to testing this properly.

PR #339, opened 2025-08-09
Scene camera focus/magnification glitch

peteihis:
"Focus distance" and "focal length" confusion. I separated these anywhere I could find many years ago, but I missed this one and it has bee haunting me ever since.

The issue:
- Focus distance _("Focal distance" in AoI terms)_ is the distance from camera lens to the depth in scene that it is looking at.
- Focal length _("Distance to plane", sorry for the name...)_ is the distance from the camera lens to the film or projection plane that generates the is image show to the user.

These two don't exist in the same space and are fully independent of each other. Sadly many sources that talk about 3D-graphics don't make the distinction between them and present as if the image of the scene was actually drawn in the scene... That was the case with AoI too.

A renaming PR would be in order, but bug fix first!

### DIFF
--- a/ArtOfIllusion/src/artofillusion/object/SceneCamera.java
+++ b/ArtOfIllusion/src/artofillusion/object/SceneCamera.java
@@ -265,14 +265,14 @@ public class SceneCamera extends Object3D
   {
     if (perspective)
     {
-      double scale = 0.5*height/Math.tan(getFieldOfView()*Math.PI/360.0);
+      double scale = 0.5*height/Math.tan(fov*Math.PI/360.0);
       Mat4 screenTransform = Mat4.scale(-scale, -scale, scale).times(Mat4.perspective(0.0));
       screenTransform = Mat4.translation((double) width/2.0, (double) height/2.0, 0.0).times(screenTransform);
       return screenTransform;
     }
     else
     {
-      double scale = 0.5*height/(Math.tan(getFieldOfView()*Math.PI/360.0)*getFocalDistance());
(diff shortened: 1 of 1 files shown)


Sailsman63:
Sorry for letting this one sit so long, I've been away for a bit.

While I don't really doubt that you are correct in these changes, could you tell me how to replicate the issue caused by the buggy version?

----------------------------------------------------------------------------

Not relevant to fixing the code, but "distance to plane" is more how far the *clipping plane* is ahead of the notional camera pinhole/center point, and has little to nothing to do with "focal length" as applied to real-world lenses.

peteihis:
I probably should have explained the effect a bit better. Took me a while to get there again but here we are.

This happens, when you render a picture with **Perspective off**. 

**CORRECTION-->**  With perspective set off, Focus distance was also zooming the camera. To make things worse the zoom effect was happening backwards to what you might expect and it was happening whether you set **Depth of Field** on or not in the rendering dialog. **<-- CORRECTION**

This is a case that you may not run into in decades, but when you do, you feel that the camera is entirely out of control. That's what happened to me. It really took some time to understand what was happening, though after all the cause was only one variable in a wrong place.

Here's a picture of what it looks like: Top row with the bug, bottom row fixed.

Before and after<br>

With **Perspective on** changing focusing distance was working as expected. This bug is from the time AoI did not make any difference between the modeling space and the rendering space inside the camera. This is a spot that I missed in the process, when the entire logic of managing the camera view was changing.

peteihis:
And I rebased it to the HEAD.

Sailsman63:
That's a trippy little bug. Thanks for the fix!

PR #341, opened 2026-05-13
Default renderers are separate instances. Resolves #331

### LINKED ISSUE #331: Camera view RENDERED mode get copied to RenderScene

Sailsman63:
@peteihis Can you test and confirm that this fixes the issues that you found?

Alternative to #333, that should avoid having to teach the renderer setup dialogs about `ViewerCanvas` and its states.

Just makes sure that the default renderers for each type of preview are different instances, and that changing the preferred render type doesn't break that. It's still possible for scripts, etc. to override through the `ApplicationPreferences` API, but if you do that, hopefully you know what you are doing.

Long term, I'd like to avoid the Plugin system carrying around instances in the first place, but that's not for today.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
+++ b/ArtOfIllusion/src/artofillusion/ApplicationPreferences.java
@@ -1,6 +1,7 @@
 /* Copyright (C) 2002-2009 by Peter Eastman
    Changes Copyright (C) 2016-2019 by Petri Ihalainen
    Changes copyright (C) 2017 by Maksim Khramov
+   Changes Copyright (C) 2026 by Lucas Stanek
 
    This program is free software; you can redistribute it and/or modify it under the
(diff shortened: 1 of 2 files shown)


peteihis:
>Can you test and confirm that this fixes the issues that you found?

It does. Much cleaner than circulating the one renderer around.

I have more fixes to the rendered mode coming: So far the rendered mode comes always in perspective and it is fixed to the default vertical angle of SceneCamera. Now it is possible to give SceneCamera the instructions to match the non-rendered mode settings. I guess, I'll dig that branch up after this fix. :)

PR #343, opened 2026-05-14
Rendered mode can do parallel

peteihis:
Perspective and parallel work in rendered mode the same way they do in other modes.

I also tried to remove the limit that only one view can be rendered at a time but a problem emerged: The change was an easy one itself but for some reason, if more than one view was in rendered mode, some of them did not respond to changes in the scene or in the object. Have to look at that some time later.

### DIFF
--- a/ArtOfIllusion/src/artofillusion/ObjectViewer.java
+++ b/ArtOfIllusion/src/artofillusion/ObjectViewer.java
@@ -1,5 +1,5 @@
 /* Copyright (C) 1999-2008 by Peter Eastman
-   Modifications copyright (C) 2017-2019 Petri Ihalainen
+   Modifications copyright (C) 2017-2024 Petri Ihalainen
    Changes copyright (C) 2017-2020 by Maksim Khramov
 
    This program is free software; you can redistribute it and/or modify it under the
@@ -28,6 +28,7 @@ import java.util.*;
 public abstract class ObjectViewer extends ViewerCanvas
 {
   protected MeshEditController controller;
+  protected SceneCamera renderedModeCamera;
   protected boolean showScene, useWorldCoords, freehandSelection, draggingBox, squareBox, sentClick;
   protected Point clickPoint, dragPoint;
   protected Vector<Point> selectBoundsPoints;
@@ -39,6 +40,7 @@ public abstract class ObjectViewer extends ViewerCanvas
   {
     super(ArtOfIllusion.getPreferences().getUseOpenGL() && isOpenGLAvailable());
     this.controller = controller;
+    renderedModeCamera = null;
     buildChoices(p);
   }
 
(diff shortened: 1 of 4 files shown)


ArtOfIllusion/src/artofillusion/ObjectViewer.java
```diff
@@ -147,9 +148,13 @@ public void viewChanged(boolean selectionOnly)
...
       rend.configurePreview();
-      Rectangle bounds = getBounds();
-      SceneCamera sceneCamera = new SceneCamera();
-      sceneCamera.setFieldOfView(Math.atan(0.5*bounds.height/(cam.getDistToScreen()*getScale()))*360.0/Math.PI);
+      int h = getBounds().height;
+      if (renderedModeCamera == null)
+        renderedModeCamera = new SceneCamera();
+      renderedModeCamera.setDistToPlane(distToPlane);
+      renderedModeCamera.setFocalDistance(distToPlane);
+      renderedModeCamera.setFieldOfView(2.0*Math.toDegrees(Math.atan(0.5*h/100.0/cam.getDistToScreen())));
```
Sailsman63, line 166: I'd be interested in knowing exactly what this (line 156)  means in geometrical terms...

Sailsman63:
Small surprise while testing: `SceneCamera`s still don't do parallel mode.

-------------------------------------------------------------------------------------------------------

> but for some reason, if more than one view was in rendered mode, some of them did not respond to changes in the scene or in the object.

Probably that the scene renderer is still a single instance. You'll probably find that if a renderer is called while it is already rendering a scene, the calls just get ignored. I'd expect that the first `SceneViewer` that gets the update signal will get the updated render, and the others are out in the cold.

peteihis:
> Small surprise while testing: SceneCameras still don't do parallel mode.

The perspective switch for SceneCamera is in the properties dialog of the SceneCamera itself. 

It is not a very good idea to have the view control the properties of the camera that you use for rendering. -- Not the perspective switch at least. Focusing distance could be coupled with the working distance of the view, but that decision too needs to be in in SceneCamera controls.

It could be helpful to put some kind of a note on the tool bar of the view that perpective switchin happens through the camera dialog ... Not sure what it should be like? Tool tip text?

peteihis:
... Or maybe the SceneCamera properties could pop up from something on the  tool bar directly?

peteihis:
> I'd be interested in knowing exactly what this (line 156) means in geometrical terms...

Well, Camera and ViewerCanvas do not use the view angle to control perspective, but SceneCamera does... so you take the the projection distance and half of height of the image to produce the tangent of half of the vertical angle. As the default scaling is 100 pixels/unit you divide the vertical pixel count by 100 to get the image height in AoI units. The projection distance is of course `distToSceen`. Then just `toDegrees` 2×arctan of `0.5×height/distToScreen`.

It would be simpler without the default scaling, just expressing the projection distance directly in pixel scale. In the default case that would be would be 2000 pixel units instead of 20 AoI units. This is one of the remaining effects of confusing the view/camera space with the model space _(which still seems to be extremely common, when perspective calculations are explained)_.

I have been meaning to draw a picture but .... ehhh ... just didn't get to it.

